PR #12, opened 2026-03-17
feat: plugin manager

midweste:
## Summary

Adds a PluginManager class that enables SocratiCode to be extended via self-contained plugins without modifying core code. Plugins are auto-discovered from src/plugins/*/index.ts at startup and receive lifecycle hooks. All plugin errors are non-fatal — a failing plugin never affects the indexer.

SocratiCode gives AI agents context about what code does and how it's structured. Plugins extend that context with knowledge that can't be extracted from source files alone — things like why code was written a certain way, which parts of the codebase are most volatile, or what implicit dependencies exist between components. This additional context is stored alongside the existing index and surfaced automatically during search, giving AI agents a deeper understanding of the project without any changes to the core.

## Changes

- **[NEW] src/services/plugin.ts** — PluginManager class with auto-discovery, registration, lifecycle dispatch, and non-fatal error isolation
- **[NEW] src/plugins/README.md** — Plugin convention docs: folder structure, interface, and how to create a plugin
- **[NEW] tests/unit/plugin.test.ts** — 12 tests covering registration, hook dispatch, error isolation, and shutdown

## Type of change

- [x] Documentation update

## Testing

- [x] Unit tests pass (npm run test:unit)
- [x] Integration tests pass (npm run test:integration) — if applicable
- [x] TypeScript compiles cleanly (npx tsc --noEmit)

12 tests covering: plugin registration, hook dispatch order, non-fatal error isolation, onProgress forwarding, shutdown resilience, and test reset. No plugins directory = gracefully skipped.

## Checklist

## Related issues

None

### DIFF
--- a/src/plugins/README.md
+++ b/src/plugins/README.md
@@ -5,10 +5,10 @@ Plugins extend SocratiCode's indexer with custom behavior. They are auto-discove
 ## Creating a Plugin
 
 1. Create a folder: `src/plugins/<your-plugin>/`
-2. Add an `index.ts` entry point that calls `registerPlugin()`:
+2. Add an `index.ts` entry point that registers with the plugin manager:
 
 ```typescript
-import { registerPlugin } from "../../services/plugin.js";
+import { pluginManager } from "../../services/plugin.js";
 import type { SocratiCodePlugin } from "../../services/plugin.js";
 
 const myPlugin: SocratiCodePlugin = {
@@ -19,10 +19,10 @@ const myPlugin: SocratiCodePlugin = {
   async onShutdown() { /* ... */ },
 };
 
-registerPlugin(myPlugin);
+pluginManager.register(myPlugin);
 ```
 
-3. That's it — `loadPlugins()` discovers and loads it automatically.
+3. That's it — `pluginManager.load()` discovers and loads it automatically.
 
 ## Plugin Structure
 
@@ -49,7 +49,3 @@ All hooks are optional. Errors are non-fatal (logged, never crash the indexer).
 | `onProjectUpdated` | After incremental update | `projectPath`, `onProgress?` |
 | `onProjectRemoved` | After index removal | `projectPath` |
 | `onShutdown` | Graceful shutdown | — |
-
-## Existing Plugins
-
-- **[git-memory](./git-memory/)** — Extracts knowledge from git history via LLMs. Enabled with `GIT_MEMORY_ENABLED=true`.
--- a/src/plugins/README.md
+++ b/src/plugins/README.md
@@ -1,0 +1,55 @@
+# SocratiCode Plugins
+
+Plugins extend SocratiCode's indexer with custom behavior. They are auto-discovered at startup — no core code changes needed.
+
+## Creating a Plugin
+
+1. Create a folder: `src/plugins/<your-plugin>/`
+2. Add an `index.ts` entry point that calls `registerPlugin()`:
+
+```typescript
+import { registerPlugin } from "../../services/plugin.js";
+import type { SocratiCodePlugin } from "../../services/plugin.js";
+
+const myPlugin: SocratiCodePlugin = {
+  name: "my-plugin",
+  async onProjectIndexed(projectPath, onProgress) { /* ... */ },
+  async onProjectUpdated(projectPath, onProgress) { /* ... */ },
+  async onProjectRemoved(projectPath) { /* ... */ },
+  async onShutdown() { /* ... */ },
+};
+
+registerPlugin(myPlugin);
+```
+
+3. That's it — `loadPlugins()` discovers and loads it automatically.
+
+## Plugin Structure
+
+```
+src/plugins/<name>/
+├── index.ts              # Entry point (required)
+├── types.ts              # Plugin-specific types
+├── __tests__/            # Co-located tests
+│   ├── unit/             # Pure function tests (no external deps)
+│   ├── integration/      # Tests requiring Qdrant, Docker, etc.
+│   └── e2e/              # Full workflow tests
+├── docs/                 # Documentation
+│   └── README.md
+└── *.ts                  # Implementation files
+```
+
+## Plugin Interface
+
+All hooks are optional. Errors are non-fatal (logged, never crash the indexer).
+
+| Hook | When | Args |
+|------|------|------|
+| `onProjectIndexed` | After full index | `projectPath`, `onProgress?` |
+| `onProjectUpdated` | After incremental update | `projectPath`, `onProgress?` |
+| `onProjectRemoved` | After index removal | `projectPath` |
+| `onShutdown` | Graceful shutdown | — |
+
+## Existing Plugins
+
+- **[git-memory](./git-memory/)** — Extracts knowledge from git history via LLMs. Enabled with `GIT_MEMORY_ENABLED=true`.
--- a/src/services/indexer.ts
+++ b/src/services/indexer.ts
@@ -21,6 +21,7 @@ import {
 import type { FileChunk } from "../types.js";
 import { ensureDynamicLanguages, getAstGrepLang, rebuildGraph, removeGraph } from "./code-graph.js";
 import { ensureArtifactsIndexed, loadConfig, removeAllArtifacts } from "./context-artifacts.js";
+import { pluginManager } from "./plugin.js";
 import { generateEmbeddings, prepareDocumentText } from "./embeddings.js";
 import { createIgnoreFilter, shouldIgnore } from "./ignore.js";
 import { acquireProjectLock, releaseProjectLock } from "./lock.js";
@@ -919,6 +920,10 @@ export async function indexProject(
     onProgress?.(`Context artifact indexing failed (non-fatal): ${artifactMsg}`);
   }
 
+  // Run post-index lifecycle hooks (git memory, future extensions)
+  progress.phase = "running post-index hooks";
+  await pluginManager.onProjectIndexed(resolvedPath, onProgress);
+
   onProgress?.(`Indexing complete: ${filesIndexed} files, ${chunksCreated} chunks`);
   lastCompleted.set(resolvedPath, {
     type: "full-index",
@@ -1217,6 +1222,10 @@ export async function updateProjectIndex(
     logger.warn("Context artifact indexing failed during incremental update (non-fatal)", { projectPath: resolvedPath, error: artifactMsg });
   }
 
+  // Run post-update lifecycle hooks (git memory, future extensions)
+  progress.phase = "running post-update hooks";
+  await pluginManager.onProjectUpdated(resolvedPath, onProgress);
+
   onProgress?.(`Update complete: ${added} added, ${updated} updated, ${removed} removed`);
   lastCompleted.set(resolvedPath, {
     type: "incremental-update",
@@ -1256,6 +1265,8 @@ export async function removeProjectIndex(projectPath: string): Promise<void> {
   await removeGraph(resolvedPath);
   // Also remove context artifacts (if any)
   await removeAllArtifacts(resolvedPath);
+  // Run post-remove lifecycle hooks
+  await pluginManager.onProjectRemoved(resolvedPath);
   projectHashes.delete(projectId);
   projectHashesLoaded.delete(projectId);
 }
--- a/src/services/plugin.ts
+++ b/src/services/plugin.ts
@@ -1,0 +1,169 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+// Copyright (C) 2026 Giancarlo Erra - Altaire Limited
+
+import { readdirSync } from "node:fs";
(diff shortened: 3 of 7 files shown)


giancarloerra:
Closing this one for now, happy to explore it again, also in consideration of all the core features that are being added and will keep being added, because the product is still new and expanding :-)

midweste:
Currently just using it as is (adding context meta to qdrant) and getting a handle on what the data looks like after using it on a few repos and what it gives in value.  Didn't have time to do the conversion to flat json yet but keeping it in mind.

One thing that I noticed and this may be relevant when using a remote qdrant server is that I was hoping to set this us for use as a shared resource for our small development team and the paths are absolute in the qdrant store.  This need to be that way?  Can we write a file to the source control that identifies the project either by name or hash and have the data be written relatively?  Just a thought

midweste:
The one gap that I saw in initial planning to put it into the existing json files schema is that it that the current json schema doesn't seem to be designed to hold extra meta.  This wouldn't allow a plugin to define what data is important in it's plugin scope:

```
    // ── Git-memory-specific fields ──
    contentHash:    "a1b2c3d4e5f67890",   // 16-char hex for dedup
    sourceCommits:  ["abc123...", "def456..."],  // full commit hashes
    filePaths:      ["src/services/auth.ts", "src/config.ts"],  // all related files
    tags:           ["authentication", "jwt", "architecture"],
    importance:     85,                    // 0-100
    confidence:     70,                    // 0-100
    memoryType:     "decision",            // one of GIT_MEMORY_TYPES
    createdAt:      "2025-06-15T10:30:00Z", // ISO date of earliest source commit
```
These are all fields I didn't see a way to make available in the current schema design but are VERY important to the knowledge of git memory operation and value.

Without this, really loses a lot of value

giancarloerra:
> Currently just using it as is (adding context meta to qdrant) and getting a handle on what the data looks like after using it on a few repos and what it gives in value. Didn't have time to do the conversion to flat json yet but keeping it in mind.
> 
> One thing that I noticed and this may be relevant when using a remote qdrant server is that I was hoping to set this us for use as a shared resource for our small development team and the paths are absolute in the qdrant store. This need to be that way? Can we write a file to the source control that identifies the project either by name or hash and have the data be written relatively? Just a thought



> Currently just using it as is (adding context meta to qdrant) and getting a handle on what the data looks like after using it on a few repos and what it gives in value. Didn't have time to do the conversion to flat json yet but keeping it in mind.
> 
> One thing that I noticed and this may be relevant when using a remote qdrant server is that I was hoping to set this us for use as a shared resource for our small development team and the paths are absolute in the qdrant store. This need to be that way? Can we write a file to the source control that identifies the project either by name or hash and have the data be written relatively? Just a thought


This was actually addressed already in the recent weeks (you might need to update): the Qdrant payload stores both [filePath] (absolute, used internally for reading the file) and [relativePath] (relative to project root). Search results returned to the AI use the relative path, so what the agent sees is something like src/services/auth.ts (lines 42-67), not /home/dev1/projects/myapp/src/services/auth.ts.

The collection naming also doesn't expose your path. By default, the collection name is a SHA-256 hash of the absolute path (e.g. codebase_a1b2c3d4e5f6), not the path itself. So, looking at Qdrant directly, you'll see hash-based names, not filesystem paths.

For your shared remote Qdrant use case, the key is [SOCRATICODE_PROJECT_ID]. 

Set this to a stable team-wide name (e.g. my-project) and every team member's instance will read/write the same Qdrant collections regardless of where they cloned the repo locally. Without it, each developer's different absolute path produces a different hash, so they'd each get their own index.

Config example (in your MCP config):

```
{
  "socraticode": {
    "command": "npx",
    "args": ["-y", "socraticode"],
    "env": {
      "SOCRATICODE_PROJECT_ID": "my-project",
      "QDRANT_URL": "https://your-remote-qdrant:6333"
    }
  }
}
```

With this, the whole team shares one index on your remote Qdrant. The SOCRATICODE_PROJECT_ID is also documented in the [Git Worktrees section] of the README and in the [Environment Variables] table.

If you want you can also join on Socraticode Discord: https://discord.gg/5DrMXfNG

giancarloerra:
> The one gap that I saw in initial planning to put it into the existing json files schema is that it that the current json schema doesn't seem to be designed to hold extra meta. This wouldn't allow a plugin to define what data is important in it's plugin scope:
> 
> ```
>     // ── Git-memory-specific fields ──
>     contentHash:    "a1b2c3d4e5f67890",   // 16-char hex for dedup
>     sourceCommits:  ["abc123...", "def456..."],  // full commit hashes
>     filePaths:      ["src/services/auth.ts", "src/config.ts"],  // all related files
>     tags:           ["authentication", "jwt", "architecture"],
>     importance:     85,                    // 0-100
>     confidence:     70,                    // 0-100
>     memoryType:     "decision",            // one of GIT_MEMORY_TYPES
>     createdAt:      "2025-06-15T10:30:00Z", // ISO date of earliest source commit
> ```
> 
> These are all fields I didn't see a way to make available in the current schema design but are VERY important to the knowledge of git memory operation and value.
> 
> Without this, really loses a lot of value

I think there may be a misunderstanding about the constraint here. Qdrant payloads are schemaless, you can store any JSON fields you want. The current code chunks have fields like [filePath], [relativePath], [content], [startLine], etc., and context artifacts add [artifactName], [artifactDescription], [contentHash], but those aren't a rigid schema that limits what can be stored.

For a git-memory plugin, you wouldn't need to modify the existing payload structure at all. The plugin would manage its own data in its own Qdrant collection (or in the existing collection with a different [type] field to distinguish git-memory points from code/artifact points). Your PR #12 already exports getClient() from qdrant.ts, so the plugin can upsert points directly with whatever payload fields it needs: sourceCommits, tags, importance, confidence, memoryType, createdAt, all of it.

The helper functions like [upsertChunks] are convenience wrappers for code indexing specifically. A plugin doing something fundamentally different (like git memory) would use the Qdrant client directly, which gives full control over the payload. The collection creation and embedding generation utilities are also exported and reusable.

So the architecture actually supports what you described already. The plugin system from your PR provides the lifecycle hooks (when to run), and the exported Qdrant client provides the storage layer (where to write). No changes to the existing schema needed.

Happy to dig into specifics, there have been many several updates and more in the planning (one of the reasons I was mentioning a plugin system at the moment must be very light touch because the product will keep changing at speed).

midweste:
Great to hear in regard to paths!  In regard to the project ID, it seems some AI client implementation use global mcps for some unknown reason and putting the "SOCRATICODE_PROJECT_ID": "my-project", in the mcp would fix it to one index yes?  (currently using Antigravity and there is no per project mcp config).  I'll look to see if .socraticode.json in project root allows for project id, that would cover it.

I think I misunderstood how .socraticodecontextartifacts.json worked.  Not at the computer yet but after looking again, it seems I could generate git memories in a folder to my own json schema and then just add them to .socraticodecontextartifacts.json ? Essentially bypassing the need for direct access to Qdrant.  Sorry I'm having to catch up mentally!  Then, the "plugin manager" would scale down to just something that triggered hooks?  This is probably what you were suggesting before but it went over my head!  Apologies

giancarloerra:
> Great to hear in regard to paths! In regard to the project ID, it seems some AI client implementation use global mcps for some unknown reason and putting the "SOCRATICODE_PROJECT_ID": "my-project", in the mcp would fix it to one index yes? (currently using Antigravity and there is no per project mcp config). I'll look to see if .socraticode.json in project root allows for project id, that would cover it.
> 
> I think I misunderstood how .socraticodecontextartifacts.json worked. Not at the computer yet but after looking again, it seems I could generate git memories in a folder to my own json schema and then just add them to .socraticodecontextartifacts.json ? Essentially bypassing the need for direct access to Qdrant. Sorry I'm having to catch up mentally! Then, the "plugin manager" would scale down to just something that triggered hooks? This is probably what you were suggesting before but it went over my head! Apologies

**Project ID with global MCPs**

Setting SOCRATICODE_PROJECT_ID in the MCP config pins it to that specific index regardless of what path the client resolves. So even if Antigravity uses a global MCP config, adding "SOCRATICODE_PROJECT_ID": "my-project" means every call maps to the same Qdrant collections. That's the intended escape hatch for clients that don't support per-project MCP configs.

On .socraticode.json: right now that file only supports linkedProjects (for cross-project search). It doesn't have a projectId field yet. So the env var is the way to go for your setup. Adding projectId to .socraticode.json is a reasonable feature request though if you want it, would cover the case where the env var isn't an option.

**Context artifacts for git memories**

No apologies needed! That's exactly the pattern:

1. Your tool generates git memory files (markdown, JSON, whatever format works for you) on disk
2. You list them in .socraticodecontextartifacts.json with a name and description
3. SocratiCode chunks, embeds, and indexes them into Qdrant automatically
4. They become searchable via codebase_context_search alongside any other artifacts

The description field in the config is important here because it gets stored in every chunk's payload. So if you set it to something like "Git-derived architectural decisions and code evolution patterns", the AI gets that context when results come back, helping it understand what it's looking at.

And yes, the "plugin manager" simplifies down to just triggering hooks. Something like: on codebase_index completion, run the git analysis script, write the output files, and let SocratiCode pick them up on the next codebase_update. No need to touch Qdrant directly at all.

The beauty of this approach is it's completely decoupled. Your git-memory tool doesn't even need to know SocratiCode exists. It just writes files. SocratiCode just indexes files. The .socraticodecontextartifacts.json config is the glue.

midweste:
That approach would simplify some of what I'm doing in the code as well.  Currently:

- Reading commits, splitting them if they are too large to be processed in one context window and reassembling them into memories about a particular commit (think initial commit with a LOT of files).  
- After all commit memories are generated, there is a synthesis step that looks at the whole corpus and evaluates the connections between them, what supersedes what and would then edit the existing memories providing meta data and re-scores the importance, confidence, links etc.

Most recently, I had changed it to write to Qdrant after every generation as if some LLM calls failed I was having to regenerate the whole commit again.  Later revisions wrote to qdrant per generation and later altered the existing memories on synthesis.

Files would work well because its a semi heavy operation. Having files based on memories makes it idempotent where once memories and generated, they would never have to be regenerated (for instance if the qdrant remote connection was down or something).  This would save LLM hits.  Also, some AI will just look at files in the codebase anyway and gather knowledge so it would help if the LLM decides its "too hard" to use the MCP.  Would run into the same memories then.

giancarloerra:
> That approach would simplify some of what I'm doing in the code as well. Currently:
> 
> * Reading commits, splitting them if they are too large to be processed in one context window and reassembling them into memories about a particular commit (think initial commit with a LOT of files).
> * After all commit memories are generated, there is a synthesis step that looks at the whole corpus and evaluates the connections between them, what supersedes what and would then edit the existing memories providing meta data and re-scores the importance, confidence, links etc.
> 
> Most recently, I had changed it to write to Qdrant after every generation as if some LLM calls failed I was having to regenerate the whole commit again. Later revisions wrote to qdrant per generation and later altered the existing memories on synthesis.
> 
> Files would work well because its a semi heavy operation. Having files based on memories makes it idempotent where once memories and generated, they would never have to be regenerated (for instance if the qdrant remote connection was down or something). This would save LLM hits. Also, some AI will just look at files in the codebase anyway and gather knowledge so it would help if the LLM decides its "too hard" to use the MCP. Would run into the same memories then.


---

**File-based approach**

That's a great way. The file-based pattern gives you exactly what you described: idempotency (skip already-generated commits), resilience (Qdrant being down doesn't lose work), and discoverability (agents browsing the repo find the memories directly). It's a strictly better architecture than writing to Qdrant on every LLM call.

**Where this leaves the PR**

Thinking about it, the file-based approach means your git-memory tool is fully external to SocratiCode. It generates files, you list them in `.socraticodecontextartifacts.json`, and indexing just works. No hooks needed in core, no plugin manager, no Qdrant client access.

So PR #12 as was scoped (plugin manager with lifecycle hooks) may not be needed anymore for your use case. That said, if after building it out you find there's a small, focused contribution that would help, like a `projectId` field in `.socraticode.json` (you mentioned the global MCP config issue), that's the kind of thing that would fit well as a standalone PR.

I'd say: build out the git-memory tool with the file-based approach, see how it feels end-to-end, and maybe we can then see how it could be integrated in SocratiCode? That way it maps directly to a real friction point rather than theoretical infrastructure (with one main application only).

Looking forward to seeing what you build with it.

midweste:
Thanks for the mental tennis match, it helped flesh things out more.  

Moving in the direction of a completely stand alone plugin presents new questions though!  It will likely live as a plugin in my fork for a while until i see all the lines clearly about how to separate it (or whether to!).  

First question on my mind now, what would trigger it...git commit hook seems like the obvious answer there, only running when something is committed, but having to add lines to every project git hooks seems like a step backward as the memory system works cross project.  Maybe there is still value in just the lifecycle hooks, though not quite as efficient as git commit hooks, they would be more user friendly as there would be no need to "install" anything.  One ENV variable boolean could turn on git memory for all projects.

I'll need to brainstorm about the new direction...

Issue #21, opened 2026-04-09
[Feature]: Java import & dependency graph support (enterprise Java/Spring Boot codebases)


sb-giithub:
## First, a sincere thank you

I want to start by expressing genuine gratitude for SocratiCode. The hybrid semantic + BM25 search quality is outstanding — it has meaningfully improved how I navigate large codebases day to day. The work you've put into this tool is clearly thoughtful and high quality, and I deeply appreciate that it's available to the community.

## The request

I'd humbly like to request Java support for the dependency graph feature (`codebase_graph_query`, `codebase_graph_stats`, etc.).

I work on a large enterprise Java/Spring Boot monorepo (30+ Maven modules, ~29,000 Java files). The semantic search works beautifully for this codebase, but the dependency graph currently produces **0 edges** for Java files — which makes sense, since the graph parser doesn't yet handle Java `import` statements.

## Why this matters

For Java projects, the ability to trace:
- Which classes import a given interface/class
- Which services call a given method (call hierarchy)
- Module-level dependency graphs (circular dependency detection between Maven modules)

...would make SocratiCode a complete, standalone solution for Java teams. Right now, Java users get the excellent RAG search but miss out entirely on the graph intelligence that JS/TS/CSS users enjoy.

## What other languages have taught us

I noticed the graph already handles JS/TS (ESM + CommonJS), CSS `@import`, Svelte, and Vue — all resolved via file path. Java is different: imports are fully qualified class names (`import cn.sino.sso.service.UserService`), which need to be resolved against the source tree (typically `src/main/java/`).

A minimal but highly valuable implementation could be:

1. **Parse `import` statements** from `.java` files (straightforward regex or tree-sitter `java` grammar, which already exists in the tree-sitter ecosystem)
2. **Resolve FQN → file path** by mapping `cn.sino.sso.service.UserService` → `src/main/java/cn/sino/sso/service/UserService.java` within the project tree
3. **Register edges** the same way JS imports are handled today

This would give Java projects the same graph-based navigation already available to frontend projects.

## Environment

- SocratiCode: v1.3.2
- Project: ~29,000 Java files, 30+ Maven modules (Spring Boot microservices)
- Platform: Windows 11, Docker (Qdrant) + native Ollama

## Happy to help

I'd be glad to test any experimental branch, provide sample Java files for parser testing, or give feedback on intermediate implementations. I understand this may not be a small ask, and I'm truly grateful for any consideration you can give it.

[...]

PR #22, opened 2026-04-09
fix: resolve JVM imports in multi-module Maven/Gradle projects (0 edges bug)

sb-giithub:
## Problem

Any Java/Kotlin/Scala project using a **multi-module Maven or Gradle layout** produces `0` dependency edges in `codebase_graph_stats`, making `codebase_graph_query`, `codebase_graph_circular`, and `codebase_graph_visualize` useless for the majority of real-world JVM codebases.

Tested on a 30-module Spring Boot monorepo (~29,000 Java files):

```
Total dependency edges: 0   ← before this fix
```

### Root cause

`resolveImport` already had correct import parsing and FQN→path conversion for Java. The resolution failed silently because the fallback directory list was limited to three fixed prefixes:

```
src/main/java/
src/main/
src/
```

A multi-module project places sources at:

```
module-sso/module-sso-service/src/main/java/cn/sino/sso/UserService.java
```

None of the three prefixes match when resolved from the project root, so every cross-module import returned `null`.

## Fix

Introduce `buildJvmSuffixMap(fileSet)` — a one-time O(n) scan that registers every JVM file by its *class-path key* (everything after `src/main/<lang>/`):

```
"cn/sino/sso/UserService.java" → "module-sso/module-sso-service/src/main/java/cn/sino/sso/UserService.java"
```

`resolveImport` accepts the map as an optional last argument and falls back to an O(1) map lookup when the existing prefix approach fails.

`buildCodeGraph` builds the map once per graph build, only when the project contains JVM files — zero overhead for non-JVM projects.

## Changes

| File | Change |
|------|--------|
| `src/services/graph-resolution.ts` | Export `buildJvmSuffixMap`; add `jvmSuffixMap?` param to `resolveImport`; use it in Java/Kotlin/Scala case |
| `src/services/code-graph.ts` | Build suffix map once before the import analysis loop; pass to `resolveImport` |
| `tests/unit/graph-resolution.test.ts` | 7 new unit tests covering map construction, test-source exclusion, multi-module Java/Kotlin resolution, unresolvable class, stdlib guard |

## Compatibility

- Fully backwards-compatible: `jvmSuffixMap` is optional; single-module projects continue to resolve via the existing prefix approach without reaching the suffix map.
- Windows/POSIX safe: map keys are built and looked up using `path.sep`.
- No new dependencies.

## Testing

All 7 new unit tests pass. The 42 pre-existing test failures on Windows are unrelated path-separator issues in the test harness (`/` vs `\`) that existed before this PR.

---

*Relates to issue #21 — Java dependency graph support*

🤖 Co-authored with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/src/services/code-graph.ts
+++ b/src/services/code-graph.ts
@@ -10,7 +10,7 @@ import { EXTRA_EXTENSIONS, getLanguageFromExtension, MAX_GRAPH_FILE_BYTES } from
 import type { CodeGraph, CodeGraphEdge, CodeGraphNode } from "../types.js";
 import { loadPathAliases } from "./graph-aliases.js";
 import { extractImports } from "./graph-imports.js";
-import { resolveImport } from "./graph-resolution.js";
+import { buildJvmSuffixMap, resolveImport } from "./graph-resolution.js";
 import { createIgnoreFilter, shouldIgnore } from "./ignore.js";
 import { logger } from "./logger.js";
 import { deleteGraphData, getGraphMetadata, loadGraphData, saveGraphData } from "./qdrant.js";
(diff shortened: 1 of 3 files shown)


giancarloerra:
Thanks merged!

PR #25, opened 2026-04-13
feat: support global config fallback and configurable embedding batch size

sb-giithub:
## Summary

Two small, independent improvements that make SocratiCode easier to share knowledge across projects and to tune for different hardware:

1. **Global fallback for `.socraticodecontextartifacts.json`** — when a project root has no config file, fall back to a global location. Configurable via `SOCRATICODE_GLOBAL_CONFIG_DIR` env var; defaults to `~/.claude/arch/`. Lets users maintain a shared knowledge base (architecture docs, SDK conventions, etc.) once and reuse it across all projects without duplicating config files.

2. **`EMBEDDING_BATCH_SIZE` env var** — the hardcoded `BATCH_SIZE = 32` in `embeddings.ts` is now configurable. Default unchanged (32). Users with high-throughput embedding providers (TEI on GPU, OpenAI, etc.) can raise it (e.g. 128) for better throughput; users on constrained hardware can lower it.

Both changes are fully backwards-compatible — without setting either env var, behavior is identical to current main.

## Changes

- `src/services/context-artifacts.ts`: `loadConfig()` falls back to global config dir when project-level file is absent
- `src/services/embeddings.ts`: `BATCH_SIZE` reads from `EMBEDDING_BATCH_SIZE` env var with validation (positive integer, throws on invalid)

## Test plan

- [x] Project-level config still takes priority when present
- [x] Global config used when project-level absent
- [x] `SOCRATICODE_GLOBAL_CONFIG_DIR` env var override works
- [x] Returns null when neither project nor global config exists (no breaking change)
- [x] `EMBEDDING_BATCH_SIZE=64` loads correctly
- [x] `EMBEDDING_BATCH_SIZE=invalid` throws clear error
- [x] Default 32 preserved when env var unset
- [x] `npm run build` passes

## Summary by CodeRabbit

* **New Features**
  * Embedding batch size is now configurable via environment, enabling performance tuning.
  * Configuration loading supports a global fallback when project-level config is missing.

* **Tests**
  * Added and expanded unit tests to validate global-fallback config behavior and embedding batch-size parsing/validation.

### DIFF
--- a/src/services/context-artifacts.ts
+++ b/src/services/context-artifacts.ts
@@ -59,6 +59,7 @@ export async function loadConfig(projectPath: string): Promise<SocratiCodeConfig
     process.env.SOCRATICODE_GLOBAL_CONFIG_DIR || path.join(os.homedir(), ".claude", "arch");
   const globalConfigPath = path.join(globalConfigDir, CONFIG_FILENAME);
   let actualPath = configPath;
+  let usingGlobalFallback = false;
 
   try {
     await fsp.access(configPath);
@@ -66,6 +67,7 @@ export async function loadConfig(projectPath: string): Promise<SocratiCodeConfig
     try {
       await fsp.access(globalConfigPath);
       actualPath = globalConfigPath;
+      usingGlobalFallback = true;
     } catch {
       return null; // neither project nor global file exists — that's fine
     }
@@ -119,6 +121,17 @@ export async function loadConfig(projectPath: string): Promise<SocratiCodeConfig
       }
       names.add(normalized);
     }
+  }
+
+  // When config was loaded from the global fallback directory, resolve relative
+  // artifact paths against that directory so downstream code (which assumes
+  // project-root resolution) receives correct absolute paths.
+  if (usingGlobalFallback && Array.isArray(config.artifacts)) {
+    const baseDir = path.dirname(actualPath);
+    config.artifacts = (config.artifacts as ContextArtifact[]).map((artifact) => ({
+      ...artifact,
+      path: path.isAbsolute(artifact.path) ? artifact.path : path.resolve(baseDir, artifact.path),
+    }));
   }
 
   return config as SocratiCodeConfig;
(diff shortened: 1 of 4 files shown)


giancarloerra:
@sb-giithub thanks for the PR, both features are useful additions. A couple of things need addressing before this can be merged (flagged by Code Rabbit):

**1. Relative artifact paths break with global config fallback**

When `loadConfig()` loads from the global directory, artifacts with relative paths (e.g. `"path": "docs/schema.sql"`) will resolve against the **project root** in `indexArtifact()`, not the global config directory. This means the main use case for this feature, that is shared artifacts across projects, won't actually work unless all paths are absolute.

You'll need to either resolve relative paths against the global config directory at load time, or propagate the config source directory so downstream code resolves correctly.

**2. `EMBEDDING_BATCH_SIZE` validation isn't strict enough**

`Number.parseInt("64abc", 10)` silently returns `64`, and `"1.5"` returns `1`, both pass validation despite not being valid positive integers. Consider something like the below?

```ts
const num = Number(raw);
if (!Number.isInteger(num) || num <= 0) {
  throw new Error(`Invalid EMBEDDING_BATCH_SIZE: "${raw}". Must be a positive integer.`);
}
return num;

sb-giithub:
Both issues fixed in 49b5b35:

**1. Relative artifact paths with global config fallback**

Added a `usingGlobalFallback` flag. After config validation, if loading from the global directory, all relative artifact paths are resolved against `path.dirname(actualPath)` (the global config dir). Absolute paths are left as-is. Project-level config behavior is unchanged.

**2. Stricter `EMBEDDING_BATCH_SIZE` validation**

Replaced `Number.parseInt(raw, 10)` with `Number(raw)` + `Number.isInteger(num)` as suggested. Now:
- `"64abc"` → `NaN` → rejected
- `"1.5"` → not integer → rejected  
- `""` → `0` → rejected (also changed guard from `!raw` to `raw === undefined` so empty string doesn't silently fall back to default)

Added tests covering both fixes, including the existing "returns null when no config" test which now isolates itself from real global configs via env var override.

giancarloerra:
@sb-giithub thanks, merged!

PR #36, opened 2026-04-28
fix(docker): include api-key header in external Qdrant readiness probe

### LINKED ISSUE #35: External Qdrant readiness probe fails on Qdrant Cloud — /healthz requires api-key
## Summary

When `QDRANT_MODE=external` points to Qdrant Cloud (or any auth-required Qdrant deployment), `codebase_index` fails on startup with:

> Cannot reach external Qdrant at https://....cloud.qdrant.io:6333.
> Verify that QDRANT_URL (or QDRANT_HOST/QDRANT_PORT) is correct and the server is reachable.

even though `codebase_health` reports the same endpoint as Reachable and `codebase_list_projects` returns the actual collection list, both succeeding.

## Root cause

`ensureExternalQdrantReady()` in `src/services/docker.ts` probes `${baseUrl}/healthz` via `waitForService()`, which calls `fetch(url)` without any auth header. Qdrant Cloud requires `api-key` on **every** endpoint, including `/healthz`, so the probe receives `403 forbidden` for all 5 retries and throws.

Meanwhile `codebase_health`, `codebase_list_projects`, and `codebase_search` go through the qdrant-js client, which attaches `QDRANT_API_KEY` automatically — hence the asymmetry.

## Reproduction

1. Configure SocratiCode against Qdrant Cloud:
   ```
   QDRANT_MODE=external
   QDRANT_URL=https://<cluster-id>.aws.cloud.qdrant.io:6333
   QDRANT_API_KEY=<token>
   EMBEDDING_PROVIDER=openai
   OPENAI_API_KEY=<key>
   ```
2. Run `codebase_health` → all green, Qdrant Reachable.
3. Run `codebase_list_projects` → returns the (possibly empty) project list.
4. Run `codebase_index <path>` → fails immediately with `Cannot reach external Qdrant`.

Confirmed against socraticode 1.6.1 and 1.7.1, Qdrant Cloud (us-west-2 cluster).

## Fix

A PR forwards `QDRANT_API_KEY` as an `api-key` header in the readiness probe. Docker-managed and unauthenticated external deployments remain unaffected (no init forwarded → no headers).

tazawa-masayoshi:
Fixes #35.

## What

Forward `QDRANT_API_KEY` as an `api-key` request header from `ensureExternalQdrantReady()` by extending `waitForService()` with an optional `RequestInit` argument.

## Why

Qdrant Cloud requires authentication on **every** endpoint, including `/healthz`. The current readiness probe sends no headers, so it always receives `403 forbidden` and throws "Cannot reach external Qdrant" — even though `codebase_health` (qdrant-js client, sends api-key) reports the same endpoint as Reachable.

## How

- `src/services/docker.ts`
  - Add `QDRANT_API_KEY` to the constants import.
  - In `ensureExternalQdrantReady`, build `init?: RequestInit` carrying `{ headers: { "api-key": QDRANT_API_KEY } }` when the env var is configured.
  - Pass `init` through `waitForService(...)`.
  - Mention `QDRANT_API_KEY` in the failure message so misconfiguration surfaces quickly.
- `waitForService` gains an optional 5th parameter `init?: RequestInit` forwarded to `fetch()`. All existing call sites (Docker-managed Qdrant, Ollama) remain unchanged → backwards compatible.

## Tests

`tests/unit/docker.test.ts` adds two cases (using the existing external-mode `vi.doMock` harness):

- `sends api-key header to /healthz when QDRANT_API_KEY is configured`
- `omits api-key header when QDRANT_API_KEY is not set`

`npm run lint` clean. `npm run test:unit` 704 pass (702 baseline + 2 new). `npm run build` clean.

## Compatibility

| Mode | Behavior |
| --- | --- |
| Docker-managed Qdrant | Unchanged — no `init` forwarded → no headers sent. |
| External Qdrant without auth | Unchanged — `QDRANT_API_KEY` unset → `init` is `undefined`. |
| External Qdrant with auth (Qdrant Cloud, self-hosted with auth) | Works — `api-key` header attached. |

## Manual verification

Reproduced the original failure against a Qdrant Cloud cluster (us-west-2), then ran the patched build through the same MCP `codebase_index` invocation: indexing now proceeds and completes (10625 chunks across 3040 files in ~5min, OpenAI text-embedding-3-small). `codebase_search` returns expected results.

## Summary by CodeRabbit

* **New Features**
  * API-key authenticated readiness checks for external Qdrant endpoints.

* **Bug Fixes**
  * Clearer error when Qdrant is unreachable, mentioning QDRANT_API_KEY as a possible missing config.
  * Prevents using an API key with plain HTTP on non-local hosts to avoid insecure key transmission.

* **Documentation**
  * Notes that QDRANT_API_KEY requires HTTPS except for loopback addresses (localhost/127.0.0.1/[::1]).

* **Tests**
  * Added unit tests covering API-key behavior for readiness checks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1095,7 +1095,7 @@ The rest of this section documents the variables themselves. Pass them using whi
 | `QDRANT_PORT` | `16333` | Qdrant REST API port (managed mode, or external without `QDRANT_URL`) |
 | `QDRANT_GRPC_PORT` | `16334` | Qdrant gRPC port (managed mode only) |
 | `QDRANT_HOST` | `localhost` | Qdrant hostname (alternative to `QDRANT_URL` for non-HTTPS external instances) |
-| `QDRANT_API_KEY` | *(none)* | Qdrant API key (required for Qdrant Cloud and other authenticated deployments) |
+| `QDRANT_API_KEY` | *(none)* | Qdrant API key (required for Qdrant Cloud and other authenticated deployments). When set, the URL must be `https://...` so the key is not transmitted over plain HTTP. Loopback URLs (`localhost`, `127.0.0.1`, `[::1]`) are accepted on `http://` for local development. |
 
 ### Indexing Behaviour
 
--- a/src/services/docker.ts
+++ b/src/services/docker.ts
@@ -201,6 +201,32 @@ async function ensureExternalQdrantReady(onProgress?: InfraProgressCallback): Pr
   const baseUrl = QDRANT_URL ? QDRANT_URL.replace(/\/$/, "") : `http://${QDRANT_HOST}:${QDRANT_PORT}`;
   const healthUrl = `${baseUrl}/healthz`;
 
+  // If an api-key is configured, refuse to send it over a non-TLS connection.
+  // Localhost loopback URLs are accepted because some users run authenticated
+  // Qdrant locally during development. The URL is parsed (rather than checked
+  // with startsWith) so that hostnames like "http://localhost.evil.com" are
+  // not mistaken for loopback. Placed before the try/catch so the specific
+  // error is not masked by the generic "Cannot reach" message below.
+  if (QDRANT_API_KEY) {
+    let parsed: URL | null = null;
+    try {
+      parsed = new URL(baseUrl);
+    } catch {
+      /* fall through: the unparseable URL will fail the reachability check */
+    }
+    const isHttps = parsed?.protocol === "https:";
+    const isLoopback = parsed
+      ? ["localhost", "127.0.0.1", "[::1]"].includes(parsed.hostname)
+      : false;
+    if (parsed && !isHttps && !isLoopback) {
+      throw new Error(
+        `QDRANT_API_KEY is set but ${baseUrl} is not HTTPS. ` +
+        "Refusing to send the API key over a non-TLS connection. " +
+        "Use https://... for remote or cloud Qdrant (a localhost URL is also accepted for local development).",
+      );
+    }
+  }
+
   onProgress?.(`Checking external Qdrant at ${baseUrl}...`);
   logger.info("Checking external Qdrant", { url: baseUrl });
 
--- a/src/services/docker.ts
+++ b/src/services/docker.ts
@@ -7,6 +7,7 @@ import {
   OLLAMA_HOST,
   OLLAMA_IMAGE,
   OLLAMA_PORT,
+  QDRANT_API_KEY,
   QDRANT_CONTAINER_NAME,
   QDRANT_GRPC_PORT,
   QDRANT_HOST,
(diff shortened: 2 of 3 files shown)


giancarloerra:
Thanks for the clear diagnosis and for verifying against a real Qdrant Cloud cluster. The asymmetry between the qdrant-js client and the raw `fetch` in `waitForService` is exactly right, and the approach (optional `RequestInit` threaded through `waitForService`) keeps the change minimal and backwards compatible with Docker mode and unauthenticated external Qdrant.

Before merging I'd like the two CodeRabbit findings addressed, plus a couple of small refinements:

### 1. CodeRabbit critical: gate the api-key on HTTPS

As flagged, the patch currently sends the API key over whatever scheme `QDRANT_URL` carries, including plain `http://`. Configuring authenticated Qdrant over non-TLS is uncommon but possible (auth proxy on HTTP), and we shouldn't leak the secret on the wire.

Two small refinements on top of the suggested fix:

**(a) Allow localhost over HTTP.** Some users run authenticated Qdrant locally during development. Without an exception the guard would break those workflows. Suggested check:

```ts
const isLocalHttp =
  baseUrl.startsWith("http://localhost") ||
  baseUrl.startsWith("http://127.0.0.1") ||
  baseUrl.startsWith("http://[::1]");

if (QDRANT_API_KEY && !baseUrl.startsWith("https://") && !isLocalHttp) {
  throw new Error(
    "QDRANT_API_KEY requires an HTTPS Qdrant endpoint (or a localhost URL).",
  );
}
```
**(b) Move the guard outside the existing try { waitForService(...) } catch { ... } block***. As written, the catch swallows any error from waitForService and replaces it with the generic "Cannot reach external Qdrant" message. If the HTTPS guard throws inside that block its specific error gets masked. Putting the check right after baseUrl is computed (before the try) keeps the message visible.

### 2. CodeRabbit minor: make the no-key test deterministic
loadDockerWithExternalMode spreads the real constants.js export:

```
const original = await importOriginal<...>();
return { ...original, QDRANT_MODE: "external", ..., ...overrides };
```

original.QDRANT_API_KEY is process.env.QDRANT_API_KEY || undefined. If a developer has QDRANT_API_KEY exported in their shell when running tests, the env value leaks through the spread, the production code attaches the header, and the toHaveBeenCalledWith(url, undefined) assertion fails. One-line fix: add QDRANT_API_KEY: undefined to the overrides in the no-key test.

### 3. Optional: a third test for the HTTPS guard
Once the guard is in place, a test asserting that ensureQdrantReady throws when QDRANT_API_KEY is set against an http://... (non-localhost) URL would lock in the security behaviour and prevent regressions.

### 4. Documentation
While you're here, please add a note in the README/DEVELOPER.md (whichever covers external Qdrant configuration) that QDRANT_API_KEY is required for Qdrant Cloud and is rejected over plain HTTP unless the host is localhost. Saves the next user the same debugging round-trip.

Once those are in I'll merge and cut a 1.7.2.

giancarloerra:
Heads up: a second user reported the same issue ( #37 ). To ship the fix sooner I'm going to push two follow-up commits to this branch myself addressing the CodeRabbit findings I called out (HTTPS guard for the api-key + the test determinism fix), plus a short docs note.

giancarloerra:
Pushed `7cdf21a` addressing the two CodeRabbit findings plus a docs note in the README.

Changes:
- HTTPS guard placed before the readiness probe (so its specific error is not masked by the generic "Cannot reach" handler), with `localhost`/`127.0.0.1`/`[::1]` accepted on `http://` for local dev. URL is parsed (not prefix-matched) to avoid hostnames like `http://localhost.evil.com` slipping through.
- `QDRANT_API_KEY: undefined` in the no-key test override so it does not flake on machines that have the env var set in the shell.
- Two new tests for the guard (rejection on plain HTTP for remote host, acceptance for localhost).
- README row for `QDRANT_API_KEY` now documents the HTTPS requirement and localhost exception.

All quality gates green: 706 unit tests, 154 integration tests, CodeRabbit clean, Snyk 0 issues. Once CI and the bot signal green I'll merge and cut 1.7.2.

PR #40, opened 2026-05-04
fix: extract JVM symbol names from declarations

a5345534:
## Summary
- fix JVM symbol extraction so Java/Kotlin/Scala declarations use the declaration name instead of the first subtree identifier
- avoid treating Java annotations like @Test / @AfterEach as method symbols
- add regression tests for Spring Boot entrypoints and annotated Java test methods

## Problem
Java symbol extraction was walking the whole declaration subtree and taking the first identifier or type_identifier. In real Spring Boot code this caused class symbols like WorkflowFlowableApplication and CamundaApplication to be indexed as String, and annotated test methods to be indexed as Test / AfterEach.

## Verification
- npm run test:unit -- tests/unit/graph-symbols.test.ts
- npm run build

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved symbol extraction for JVM languages (Java, Kotlin, Scala).
  * Strip annotations from identifiers so annotations aren’t emitted as method names; methods still extracted when annotation and method share a line.
  * Prefer declared class name for Spring Boot entrypoint detection over parameter type names.

* **Tests**
  * Added JVM unit tests for annotation handling, entrypoint selection, and expanded Scala symbol coverage.

### DIFF
--- a/src/services/graph-symbols.ts
+++ b/src/services/graph-symbols.ts
@@ -470,10 +470,8 @@ function extractFromJvm(
     : ["class_declaration", "interface_declaration", "enum_declaration", "object_declaration"];
   for (const k of classKinds) {
     for (const cls of safeFindAll(root, k)) {
-      const nameNode = cls.find({ rule: { kind: "type_identifier" } })
-        ?? cls.find({ rule: { kind: "identifier" } });
-      if (!nameNode) continue;
-      const name = nameNode.text();
+      const name = extractJvmTypeName(cls.text(), langKey);
+      if (!name) continue;
       const r = cls.range();
       const startLine = r.start.line + 1;
       const endLine = r.end.line + 1;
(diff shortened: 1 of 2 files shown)


giancarloerra:
Thanks for the detailed fix. Both root causes (parameter-type masking the class name on Spring Boot entrypoints, annotations masking method names on Java tests) are diagnosed precisely, the regex-based extraction is appropriately scoped, and the regression tests cover the realistic patterns. CodeRabbit's two findings on the new helpers are addressed across `6a76ad4` and `1dbc1eb`.

CI is green across Lint & Type Check, Unit Tests on Node 18 / 20 / 22, and CodeRabbit. Merging now.

Non-blocking thought for any future iteration: regex parsing of declaration headers will likely need a tree-sitter field-based fallback at some point (records, multi-line headers with comments between annotations and the `class` keyword, primary constructors with default values, generics with bounds). Filing as a separate "harden JVM symbol extraction" issue if you hit anything else makes sense. Not for this PR.

The companion issue #41 is next.

PR #53, opened 2026-05-06
Feat/project id in socraticode json

airmonitor:
## Summary

Adds a new optional `projectId` field to `.socraticode.json` so teams can commit a stable project identifier to the repo. Without this field, the project ID is derived from the SHA-256 of the absolute checkout path — which means the same project resolves to a different Qdrant collection on every machine (different OS user, different filesystem layout, different worktree, etc.). With it, every checkout addresses the same `codebase_*` / `codegraph_* `/ `context_*` collections regardless of where the working tree lives on disk.

This is the path-independent, multi-project complement to the existing `SOCRATICODE_PROJECT_ID` env var: env is process-scoped and global to all projects in a host (so it doesn't scale to a developer who works on several projects on the same laptop), whereas the file is per-project and shared across teammates via git.

## Changes

- **`src/config.ts`**
  - Add `projectId` to the `SocratiCodeConfig` shape (`.socraticode.json`).
  - Extend `projectIdFromPath()` with new precedence: `SOCRATICODE_PROJECT_ID` env > file `projectId` > path hash. Validation `[a-zA-Z0-9_-]+` and whitespace trimming applied to both override paths; invalid characters fail loud, malformed JSON falls through silently (matching `loadLinkedProjects`'s resilience). Branch-aware mode is suppressed for either override — explicit identifiers are stable by intent.
  - Extract a private `loadSocratiCodeConfig()` helper used by both `projectIdFromPath()` and `loadLinkedProjects()` (DRY).
  - Extract `effectiveBaseProjectId()` (file `projectId` ∨ path hash, no env/branch suffix) and use it in `resolveLinkedCollections()` for *linked* projects so each linked project's own committed `projectId` is honored. Without this, cross-project search would silently miss a linked project's data when that project pinned its id via the file.
  - In `resolveLinkedCollections()`, seed the dedup set with the *current project's full id* (`projectIdFromPath`, env-aware) so the dedup key trivially matches the collection-name suffix being added. A linked project is skipped only when it would resolve to the exact same collection the current project already occupies. (See the dedicated `fix:` commit on this branch — kept separate for reviewability.)
- **`tests/unit/config.test.ts`** — 17 new TDD-driven unit tests:
  - `projectIdFromPath` (13): file resolution, ignores path differences, whitespace trimming, throws on invalid chars, falls back to hash on empty/whitespace/wrong-type/null/missing-file/malformed-JSON, env-var precedence, branch-suffix suppression, coexistence with `linkedProjects`.
  - `resolveLinkedCollections` (4): linked project's committed `projectId` honored, dedup on shared committed `projectId`, env var does not leak into linked-project collection names, no wrong-dedup when env override diverges from file `projectId` (regression for the seed-vs-collection-name misalignment fixed in this PR).
- **`README.md`** — new "Team-Shared Index (committed `projectId`)" section between *Git Worktrees* and *Cross-Project Search*; env-var table notes the new precedence relationship.
- **`DEVELOPER.md`** — *Project ID & Collection Naming* section now documents the three-level precedence and explains why both override paths suppress branch-aware suffix.

## Type of change

- [x] Documentation update

## Testing

- [x] Unit tests pass (`npm run test:unit`) — **730 passed / 7 failed**. The 7 failures are pre-existing on `main` and unrelated to this change (verified by re-running with the change stashed). They are environmental: 3 in `config.test.ts` (`detectGitBranch` / `SOCRATICODE_BRANCH_AWARE` tests fail when the developer's *global* git config has `commit.gpgsign=true` because the test helper `initTempRepo` does not disable signing) and 4 in `constants.test.ts` / `docker.test.ts` (tied to Qdrant/Docker env on the host). Happy to file a follow-up PR for the `initTempRepo` GPG hardening if helpful.
- [x] Integration tests pass (`npm run test:integration`) — **154 passed / 0 failed / 10 files / 156s**, run against the Docker-managed Ollama (`nomic-embed-text`) and Qdrant containers. My local dev env normally points at LM Studio + an external Qdrant, so I had to unset `SOCRATICODE_QDRANT_URL`, `EMBEDDING_DIMENSIONS`, `OLLAMA_URL`, `LMSTUDIO_*` and set `EMBEDDING_PROVIDER=ollama`, `EMBEDDING_MODEL=nomic-embed-text` for the run to match the suite's expectations.
- [x] New tests added for new/changed functionality (17 tests, all RED → GREEN through TDD)

Lint also clean: `npm run lint` (`biome check src/ tests/`).

## Checklist

- [x] I have updated documentation (`README.md` / `DEVELOPER.md`) if needed
- [ ] I have addressed all [CodeRabbit](https://coderabbit.ai) review comments (or marked as resolved with explanation) — *will address after the bot runs on this PR*
- [ ] I have read the [Contributing Guide](../CONTRIBUTING.md)
- [ ] I agree to the [Contributor License Agreement](../CLA.md)

## Related issues

---

## Notes for reviewers

- **Backwards compatibility**: zero. Existing repos without the new field continue to use the path hash exactly as before. The `SocratiCodeConfig` interface gains an optional field; existing `linkedProjects` parsing is untouched semantically (just routed through the new shared loader).
- **Why fail loud on invalid `projectId` chars but silent on malformed JSON?** Same rationale as the existing env-var path: silent fallback would hide misconfigurations that map a project to the *wrong* collection (a different — possibly empty — index). Malformed JSON, in contrast, indicates "config not usable" rather than "config says X but X is invalid"; falling through preserves the resilience of `loadLinkedProjects()` against hand-edited files.
- **Why doesn't the env var apply to linked projects?** It's process-scoped — applying it as a single value to *every* linked path would collapse all of them onto the env-var collection name, silently losing per-project isolation. Each linked project's own `.socraticode.json` is the only source that carries unambiguous identity for that project.
- **Migration for existing indexed projects**: a team can either keep the existing path-hash collection by setting `projectId` to the current 12-hex value, or pick a semantic name and re-index once. The README notes this as a recommendation but does not enforce a migration.

## Summary by CodeRabbit

* **New Features**
  * Configurable project ID via committed project config; env var still takes precedence
  * Branch-aware mode to append sanitized branch names to project IDs; explicit overrides exempt
  * New environment variables for linked projects and logging

* **Documentation**
  * Clarified project ID resolution order, trimming/validation rules, and collection naming across worktrees/linked projects
  * Guidance for team-shared index with committed project IDs

* **Tests**
  * Expanded tests covering config-file precedence, validation, branch-aware behavior, and linked-project deduplication

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -269,17 +269,24 @@ Defined in `src/services/docker.ts`: after starting the container, the server po
 
 ### Project ID & Collection Naming
 
-Defined in `src/config.ts`:
-- **Project ID**: First 12 characters of SHA-256 hash of the absolute project path.
+Defined in `src/config.ts`. `projectIdFromPath()` resolves the project ID with the following precedence (highest first):
+
+1. **`SOCRATICODE_PROJECT_ID` env var** — per-machine override; bypasses both file lookup and path hashing.
+2. **`projectId` field in `.socraticode.json`** — committed, team-wide stable identifier; survives different filesystem layouts and OS users.
+3. **First 12 characters of SHA-256 of the absolute project path** — default fallback.
+
+In both override paths the value must match `[a-zA-Z0-9_-]+`; whitespace is trimmed; empty/whitespace-only values fall through to the next level. Invalid characters in an explicit override fail loud (throw) — silent fallback would hide misconfigurations that map a project to the wrong (or new) collection.
+
+Collection names derived from the project ID:
 - **Code collection**: `codebase_{projectId}`
 - **Graph collection**: `codegraph_{projectId}`
 - **Context artifacts collection**: `context_{projectId}`
 
-This means the same folder path always maps to the same collection, even across restarts.
+With the default (path-hash) ID, the same folder path always maps to the same collection across restarts. With either override, the mapping is stable across machines and checkouts.
 
 #### Branch-aware mode
 
-When `SOCRATICODE_BRANCH_AWARE=true`, the current git branch is detected via `git rev-parse --abbrev-ref HEAD` and appended to the project ID (e.g. `abc123def456__feat_my-feature`). Branch names are sanitized: non-alphanumeric characters (except `-`) become `_`, consecutive underscores collapse, leading/trailing underscores are stripped. Detached HEAD states fall back to the branchless ID. Ignored when `SOCRATICODE_PROJECT_ID` is set explicitly.
+When `SOCRATICODE_BRANCH_AWARE=true`, the current git branch is detected via `git rev-parse --abbrev-ref HEAD` and appended to the project ID (e.g. `abc123def456__feat_my-feature`). Branch names are sanitized: non-alphanumeric characters (except `-`) become `_`, consecutive underscores collapse, leading/trailing underscores are stripped. Detached HEAD states fall back to the branchless ID. Ignored when `SOCRATICODE_PROJECT_ID` is set explicitly or when `projectId` is set in `.socraticode.json` — explicit identifiers are treated as stable and not augmented per branch.
 
 #### Linked projects
 
--- a/README.md
+++ b/README.md
@@ -766,6 +766,20 @@ With this config, agents running in `/repo/main`, `/repo/worktree-feat-a`, and `
 - Your AI agent reads actual file contents from its own worktree; the shared index is only used for discovery and navigation
 - When changes merge back to main, the file watcher re-indexes the changed files and the index converges
 
+### Team-Shared Index (committed `projectId`)
+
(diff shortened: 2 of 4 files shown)


airmonitor:
Both findings refer to the same bug — already fixed in commit `392c77e` ("fix(config): align resolveLinkedCollections dedup seed with collection name"), which is included in this PR.

The fix is exactly what was proposed:

```diff
-const seen = new Set<string>([effectiveBaseProjectId(resolvedRoot)]);
+const seen = new Set<string>([currentId]);
```

A regression test was added at `tests/unit/config.test.ts:360` (`does not wrongly dedup a linked project when env override diverges from file projectId`) covering the exact env-override + matching-file-projectId scenario described.

The auto-walkthrough comment at 12:00:21Z (covering range `e0350629...392c77e`, i.e. *after* the fix) confirms: "No actionable comments were generated in the recent review." This inline review appears to have been prepared against the pre-fix state and posted after the fix landed.

Marking as resolved.

PR #57, opened 2026-05-08
Litellm

airmonitor:
## Summary

Adds **LiteLLM Proxy Server** as a first-class embedding provider. LiteLLM exposes an OpenAI-compatible `/v1/embeddings` endpoint and fans out to 100+ underlying providers (OpenAI, Anthropic, Cohere, Voyage, HuggingFace, Bedrock, Vertex AI, Ollama, ...), so a single SocratiCode install can target any of them without per-provider code paths. The implementation mirrors the LM Studio strategy but is a dedicated provider rather than a flag on `provider-openai`, because three behaviours diverge meaningfully from both OpenAI and LM Studio: mandatory authentication (master key / virtual key gates `/v1/models`), no sensible defaults for `EMBEDDING_MODEL` / `EMBEDDING_DIMENSIONS` (model aliases come from the operator's `config.yaml`), and provider-dependent acceptance of the `dimensions` parameter (Matryoshka-aware backends accept it, BGE / nomic / Cohere v3 reject — opt-in via `LITELLM_SEND_DIMENSIONS=true`).

## Changes

- **New** `src/services/provider-litellm.ts` (+303 LOC): `LiteLLMEmbeddingProvider` built on the OpenAI SDK with a custom `baseURL` (default `http://localhost:4000/v1`). Batch size 256 (between OpenAI's 512 and LM Studio's 64). `ensureReady` distinguishes *proxy-unreachable* / *auth-rejected* / *alias-not-registered* by duck-typing `err.status` (401/403) and lists up to 10 currently-registered aliases in the missing-alias error so the operator can sanity-check `config.yaml` from the log alone.
- `src/services/embedding-config.ts`: extends `EmbeddingProvider` union with `"litellm"`, adds `litellmUrl`, fail-fast validation for `LITELLM_API_KEY` + `EMBEDDING_MODEL` + `EMBEDDING_DIMENSIONS` (key checked first so a virtual-key user fixes the easy problem before touching proxy config), updates the *Invalid EMBEDDING_PROVIDER* error and the `hasApiKey` log expression.
- `src/services/embedding-provider.ts`: factory case for `litellm` with dynamic import — keeps the OpenAI SDK out of startup for non-LiteLLM users.
- `encoding_format=float` fix from `bb141a0` ported verbatim — without it the OpenAI SDK 6.x base64-decode path corrupts backends that return plain JSON float arrays (many LiteLLM aliases do, including Ollama-routed and TEI-wrapped ones).
- `README.md`: dedicated LiteLLM section, MCP host config example, env-var table entries clarifying which values are mandatory under `litellm`, new *LiteLLM Configuration* table.
- `tests/unit/embedding-config.test.ts`: +9 cases (model/dim/key required, error-ordering, URL default + override, dimensions parsing, `EMBEDDING_CONTEXT_LENGTH` override for unknown aliases, auto-detection when alias matches a known model name) + updated *full external config* fixture and invalid-provider error message.
- `tests/unit/embedding-provider.test.ts`: factory test for `litellm`, plus 4 cases against a deliberately-closed port (config rejects construction without `API_KEY`; `ensureReady` unreachable error format; `healthCheck` short-circuits on missing key without a network call; `healthCheck` reaches *Not reachable* without throwing).

**Backward compatible.** Provider is opt-in via `EMBEDDING_PROVIDER=litellm`. Existing `ollama` / `openai` / `google` / `lmstudio` paths are untouched.

Diffstat: 6 files changed, +638 / −14.

## Type of change

- [x] Documentation update

## Testing

- [x] Unit tests pass (`npm run test:unit`) — 64/64 on the touched suites (`embedding-config`, `embedding-provider`); also `biome` lint clean per the commit author.
- [x] Integration tests pass (`npm run test:integration`) — **153/154 with `QDRANT_MODE=external` against a remote Qdrant**; **150/153 with the default Docker-managed Qdrant**. Both runs were performed against this commit (`1708510`). Detail on the residual failures:
  - The 1 failure on the external-Qdrant run is `tests/integration/docker-ollama.test.ts > Qdrant container management > reports Qdrant as running after ensure`. The test asserts that a *Docker-managed* Qdrant container is up; with `QDRANT_MODE=external` no such container is started, so the assertion is structurally incompatible with the config — **not a regression introduced by this PR** (this PR doesn't touch the Docker-Qdrant code path).
  - The 3 failures on the managed-Qdrant run (`tools.test.ts > impact-analysis tool handlers` setup + `graph tool handlers > codebase_graph_query` + `graph tool handlers > codebase_graph_remove`) all surface as `ECONNREFUSED 127.0.0.1:16333` from the `@qdrant/js-client-rest` client, while the parallel `qdrant.test.ts` / `code-graph.test.ts` / `indexer.test.ts` suites use Qdrant successfully in the same run. The same three tests pass cleanly against the external Qdrant. Looks like a pre-existing race / suite-ordering issue in the local managed-Qdrant test infrastructure (container teardown vs. cross-suite client reuse), not embeddings-related — also **not introduced by this PR**.
- [x] TypeScript compiles cleanly (`npx tsc --noEmit`) — per commit author's verification.
- [x] New tests added for new/changed functionality — 9 new cases in `embedding-config.test.ts` covering provider validation; 5 in `embedding-provider.test.ts` covering the factory and the closed-port error paths.

## Checklist

- [x] My code follows the existing code style and conventions (mirrors the existing `provider-lmstudio` / `provider-openai` shape; OpenAI-SDK + custom `baseURL` pattern; ESM `.js` import extensions; SPDX header).
- [x] I have added/updated JSDoc comments where appropriate (public methods on `LiteLLMEmbeddingProvider` and the new fields on `EmbeddingConfig`).
- [x] I have updated documentation (README.md / DEVELOPER.md) if needed — `README.md` gains a dedicated LiteLLM section and a configuration table; env-var table updated for `litellm`-specific mandatory values.
- [ ] I have addressed all [CodeRabbit](https://coderabbit.ai) review comments (or marked as resolved with explanation) — *to be confirmed once CodeRabbit runs against the opened PR.*

## Summary by CodeRabbit

* **New Features**
  * Added LiteLLM as a supported embedding provider, enabling access to 100+ providers via a single environment-variable selection.

* **Documentation**
  * Expanded configuration docs with LiteLLM setup, example JSON, required env vars, and optional dimension-forwarding guidance.

* **Tests**
  * Added unit tests covering LiteLLM configuration, provider selection, readiness, and health-check scenarios.

* **Bug Fixes / Validation**
  * Improved fail-fast validation and clearer error messages for missing LiteLLM-related configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,7 +242,7 @@ On VS Code's 2.45M‑line codebase, SocratiCode answers architectural questions 
 - **Hybrid code search** — Built on Qdrant, a purpose-built vector database with HNSW indexing, concurrent read/write, and payload filtering. Each chunk stores both a dense vector and a BM25 sparse vector; the Query API runs both sub-queries in a single round-trip and fuses results with Reciprocal Rank Fusion (RRF). Semantic search handles conceptual queries like "authentication middleware" even when those exact words don't appear in the code. BM25 handles exact identifier and keyword lookups. You get the best of both in every query with no tuning required.
 - **Configurable Qdrant** — Use the built-in Docker Qdrant (default, zero config) or connect to your own instance (self-hosted, remote server, or Qdrant Cloud). Configure via `QDRANT_MODE`, `QDRANT_URL`, and `QDRANT_API_KEY` environment variables.
 - **Configurable Ollama** — Use the built-in Docker Ollama (default, zero config) or point to your own Ollama instance (native install -GPU access-, remote server, etc.). Configure via `OLLAMA_MODE`, `OLLAMA_URL`, `EMBEDDING_MODEL` and `EMBEDDING_DIMENSIONS` environment variables.
-- **Multi-provider embeddings** — Switch between Local Ollama (private, GPU access), Docker Ollama (zero-config), OpenAI (`text-embedding-3-small`, fastest), Google Gemini (`gemini-embedding-001`, free tier), or LM Studio (local OpenAI-compatible server) with a single environment variable. No provider-specific configuration files.
+- **Multi-provider embeddings** — Switch between Local Ollama (private, GPU access), Docker Ollama (zero-config), OpenAI (`text-embedding-3-small`, fastest), Google Gemini (`gemini-embedding-001`, free tier), LM Studio (local OpenAI-compatible server), or LiteLLM (proxy gateway in front of 100+ providers) with a single environment variable. No provider-specific configuration files.
 - **Private & secure** — Everything runs on your machine — your code never leaves your network. The default Docker setup includes Ollama (embeddings) and Qdrant (vector storage) with no external API calls. No API costs, no token limits. Suitable for air-gapped and on-premises environments. Optional cloud providers (OpenAI, Google Gemini, Qdrant Cloud) are available but never required.
 - **AST-aware chunking** — Files are split at function/class boundaries using AST parsing (ast-grep), not arbitrary line counts. This produces higher-quality search results. Falls back to line-based chunking for unsupported languages.
 - **Polyglot code dependency graph** — Static analysis of import/require/use/include statements using ast-grep for 18+ languages. No external tools like dependency-cruiser required. Detects circular dependencies and generates visual Mermaid diagrams.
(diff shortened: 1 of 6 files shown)


airmonitor:
Addressed CodeRabbit's pagination concern on `client.models.list()` in commit 6c67965.

**Concern (lines 144–181 of `src/services/provider-litellm.ts`):** `modelList.data` only contains the first page of `/v1/models`. Today's LiteLLM proxy returns the full `model_list` from `config.yaml` in a single response, but a future LiteLLM build — or an upstream proxy in front of it — that paginates `/v1/models` would cause `ensureReady()` and `healthCheck()` to throw a spurious *"alias not registered"* error for any alias landing on a non-first page.

**Fix:** Both `ensureReady()` and `healthCheck()` now use the SDK's documented async-iteration pattern (`for await (const m of client.models.list())`), which transparently fetches subsequent pages on demand. Model ids are accumulated into a single array before the membership check, so the "Currently registered models" log line in the missing-alias error reflects the merged result across all pages. An inline comment explains why iteration matters even though today's LiteLLM doesn't paginate, so the pattern survives future drive-by simplifications.

Diff: 1 file changed, +19 / −6. No behavioural change against the current LiteLLM proxy; future-proofs against paginated `/v1/models` responses.

PR #67, opened 2026-06-02
feat(graph): Lua symbol/call extraction + fix file discovery for whitelist .gitignore

SIRTHEO:
﻿## Summary
Two small graph-engine fixes that together make the code graph work for Lua
projects (and unblock any repo using a whitelist-style `.gitignore`).

### 1. Fix: getGraphableFiles skips whitelisted directories
`getGraphableFiles` calls `shouldIgnore(ig, relPath)` with the bare relative
path for every entry, directories included. For a repo that ignores everything
and re-includes a folder by its trailing-slash form — e.g. `/*` then `!/src/` —
`ignores("src")` is true while `ignores("src/")` is false. The walk skips the
`src` directory and never descends, so the graph is empty for EVERY language on
such repos. Fix: pass the directory form (trailing slash) for directory entries,
matching gitignore directory semantics. Safe for normal ignores.

### 2. Feature: dedicated Lua symbol extractor
Lua had no node-kind-specific extractor and fell through to the regex fallback,
which records `Mod` for `function Mod.parse()`. Adds `extractFromLua`, walking
the ast-grep Lua tree: `function T.f()`/`T:m()`/`local function f()` (name = the
direct child before `parameters`), `T.f = function() ... end` assignments (RHS
must be directly a function), and call sites. Registers `@ast-grep/lang-lua`.
Namespace-table style now resolves to precise qualified symbols (`Table.method`).

## Testing
- `npm run build` (tsc strict) passes.
- On a ~40-file Lua project: 0 -> 42 file nodes, 0 -> 683 symbols with qualified
  names + call edges. Same-file callees resolve; cross-file caller resolution is
  still bounded by the import-based resolver (Lua has no imports) — out of scope.

## Notes
No breaking changes; other languages unaffected. The two changes are independent
but bundled since the Lua graph is only observable once discovery works.

## Summary by CodeRabbit

* **New Features**
  * Added Lua language support for code analysis, symbol extraction, and call tracing.

* **Bug Fixes**
  * Improved directory path handling in file discovery to ensure accurate traversal and ignores.

* **Tests**
  * Added regression and unit tests covering Lua symbol extraction and file-discovery behavior.

* **Chores**
  * Added a runtime dependency required for Lua parsing.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,6 +16,7 @@
         "@ast-grep/lang-go": "^0.0.5",
         "@ast-grep/lang-java": "^0.0.6",
         "@ast-grep/lang-kotlin": "^0.0.6",
+        "@ast-grep/lang-lua": "^0.0.7",
         "@ast-grep/lang-php": "^0.0.6",
         "@ast-grep/lang-python": "^0.0.5",
         "@ast-grep/lang-ruby": "^0.0.6",
@@ -165,6 +166,24 @@
       "version": "0.0.6",
       "resolved": "https://registry.npmjs.org/@ast-grep/lang-kotlin/-/lang-kotlin-0.0.6.tgz",
       "integrity": "sha512-wMUECzl0Gzle5Rj/1Vu/EBRYpySzmhLrAvbdOCLtxq/IPz8f9YBzZjHe/o2HCdTj6EkW5Z6E7XPbSsgT1rj2pQ==",
+      "hasInstallScript": true,
+      "license": "ISC",
+      "dependencies": {
+        "@ast-grep/setup-lang": "0.0.6"
+      },
+      "peerDependencies": {
+        "tree-sitter-cli": "0.25.8"
+      },
+      "peerDependenciesMeta": {
+        "tree-sitter-cli": {
+          "optional": true
+        }
+      }
+    },
+    "node_modules/@ast-grep/lang-lua": {
+      "version": "0.0.7",
+      "resolved": "https://registry.npmjs.org/@ast-grep/lang-lua/-/lang-lua-0.0.7.tgz",
+      "integrity": "sha512-xGJpC06HTGSb2KPD5YEuzdj7wo4ZRCkMPWWUCzXUqP8stkoJYp2SW6VExeInSIMLnGyEMQAGdjCWttgyRzTGOQ==",
       "hasInstallScript": true,
       "license": "ISC",
       "dependencies": {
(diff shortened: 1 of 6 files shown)


giancarloerra:
Thanks for this, it is a clean and well-scoped contribution, and CI is green across Node 18/20/22.

I verified both halves against the engine:

Discovery fix: correct and, importantly, complete. I tested our ignore@7.0.5 with the exact /* then !/src/ pattern: ignores("src") is true (the old walk bailed here, hence the empty graph), ignores("src/") is false (your fix descends), and ignores("src/foo.lua") is false, so the files under the re-included directory are actually picked up, not just the directory entered. I also confirmed the search/embedding discovery path (getIndexableFiles, which uses glob(nodir:true) plus a per-file check) is unaffected, so scoping this to the graph walk is right.

Lua extractor: faithful to the existing extractors. Good calls on taking the direct child before parameters as the name, requiring the assignment RHS to be directly a function_definition, and filtering keywords on call sites.

A few things before merge:

Add real Lua tests. tests/unit/graph-symbols.test.ts has a per-language block with concrete assertions for every other extractor. Please add a describe("Lua") that locks in the new behavior: qualified Table.method / T:m() names, local function, the T.f = function() ... end form, and call attribution to the enclosing function.

Fix the now-misleading existing test. it("handles Lua via regex fallback", ...) (currently in the "Regex fallback" block) no longer exercises the fallback, since Lua now routes to extractFromLua. It still passes because it only asserts <module> exists, but the name and placement are now wrong. Please move it into the new Lua block and give it real assertions.
Add a regression test for the discovery fix. A whitelist .gitignore fixture (/* then !/src/) that asserts files under src are discovered. Note getGraphableFiles is not currently exported, so this likely means either exporting it for the test or asserting through the graph build.

Minor style: every other dynamically-registered grammar calls parse("x" as unknown as Lang, source). Please match that for parse("lua", ...) for consistency. Not blocking since it compiles.

The discovery fix is valuable also on its own. Thanks again.

giancarloerra:
@coderabbitai review

SIRTHEO:
Thanks for the thorough review — all addressed in `8be929a`:

1. **Lua tests** — new `describe("Lua")` block locking in qualified `Table.method` / `T:m()` names, `local function`, both the `T.f = function() … end` and `local f = function() … end` forms, plus call attribution to the enclosing function (and top-level calls falling back to `<module>`).
2. **Misleading test** — moved the former `"handles Lua via regex fallback"` into the Lua block with real assertions, and dropped "Lua" from the regex-fallback block title.
3. **Discovery regression** — added a whitelist `.gitignore` fixture (`/*` then `!/src/`) asserting files under the re-included dir are discovered; exported `getGraphableFiles` for the test.
4. **Style** — `parse("lua" as unknown as Lang, source)` to match the other dynamic grammars.

Local suite green for the touched areas; lint clean.

giancarloerra:
@coderabbitai review

Issue #45, opened 2026-05-04
fix: Go codebase_graph_query returns 0 edges — module path resolution not implemented


mrsuit92:
## Summary

Go projects produce 0 dependency edges in `codebase_graph_query` / `codebase_graph_stats`, even though import extraction works correctly. This mirrors the JVM multi-module (#22), C# (#33), and PHP (#43) cases that have already been fixed.

## Root Cause

`src/services/graph-resolution.ts:195-199`:

```ts
case "go": {
  // Go imports are package paths; only track if it looks like a local module
  // (contains the module name or starts with ./)
  return null; // Go resolution requires go.mod analysis
}
```

All Go imports are unconditionally returned as `null`. ast-grep correctly extracts the import path (`graph-imports.ts:214-224`), but resolution is explicitly skipped — so no file-level edges are ever produced for Go.

## Reproduction

Index any Go project (single- or multi-module). Example: 21 Go files in a project with a valid `go.mod`, standard service layout — `codebase_graph_stats` reports 0 dependency edges and every Go file shows 0 connections.

Symbol-level tools (`codebase_impact`, `codebase_flow`, `codebase_symbol`) work correctly because they use a different code path.

## Proposed Solution

1. Parse `go.mod` at graph-build time to learn the module path (e.g. `github.com/user/repo`).
2. For each Go import, strip the module prefix to get the local package path.
3. Map package path → directory containing `*.go` files.
4. Treat imports without the module prefix (and the standard library) as external — same approach as the existing logic for other languages.

This is analogous to `buildJvmSuffixMap` from PR #22: a one-time scan at graph-build time, then O(1) lookups during resolution.

## Workaround

None at the file-graph level. Symbol-level analysis still works.

## Environment

- SocratiCode 1.8.4
- macOS, Docker, Ollama (`nomic-embed-text`)
- Go 1.21+ project, standard layout with `go.mod` at root

PR #84, opened 2026-07-20
fix(graph): resolve Go imports when go.mod is nested in a monorepo (#82)

### LINKED ISSUE #82: [Bug]: Go graph + impact still return 0 edges when go.mod is nested (monorepo) — module-path  resolution only finds root-level go.mod.
### Describe the bug

On SocratiCode 1.8.18 (latest), in a Go project whose go.mod is nested below the indexed root (a monorepo layout), the Go dependency graph and symbol-level Impact Analysis produce no edges and no callers for Go files, even though the imports and call sites clearly exist. Non-Go files in the same index (TypeScript/Vue) resolve correctly, so  the graph builder works in general — it specifically fails to resolve Go when go.mod is not at the indexed project root.

 After indexing such a repo:
 - codebase_graph_stats lists the Go files in the language breakdown, but every .go file appears as an orphan with 0 connections;  the total edge count reflects only the non-Go files.
 - codebase_graph_query <any .go file> → No dependency information found for this file.
 - codebase_impact <Go symbol> → Total impacted files: 0 — No callers found., even for a symbol that is plainly called from other files in the module.

 This is the same class of bug as #45 ("Go codebase_graph_query returns 0 edges — module path resolution not implemented"), which  was fixed for projects with go.mod at the root (commits c156da1 + 8c26ed8). That fix resolves Go imports by parsing the go.mod module path, but it does not appear to locate go.mod when it sits in a subdirectory of the indexed path — so no module path is learned, no Go edges are produced, and cross-package symbol resolution silently fails. Note #45 states that "symbol-level tools use a different code path and work"; in the nested-go.mod case they do not work either, because cross-package call resolution still depends on the module-path map that was never built.

Expected: Go imports resolved via the nearest go.mod (found by searching subdirectories from the indexed root, or by indexing per-module), producing file-graph edges and correct symbol blast-radius — matching the behavior of a single-module repo where go.mod is at the root.

### Steps to reproduce

 1. Create a monorepo whose Go module is nested one level below the root:

 ```
   my-monorepo/                         # <-- this path is passed to codebase_index
   ├── docker-compose.yml
   ├── frontend/                        # any TS/Vue app (graph resolves fine here)
   │   └── src/app.ts
   └── backend/                         # Go module lives here, NOT at the root
       ├── go.mod
       ├── cmd/server/main.go
       ├── internal/middleware/auth.go
       └── internal/service/user.go
 ```

 Minimal file contents:

 backend/go.mod

 ```
   module github.com/example/myapp-backend

   go 1.22
 ```

 backend/internal/middleware/auth.go

 ```go
   package middleware

   // Authorize returns true if the role may perform the action.
   func Authorize(role string) bool {
      return role == "admin"
   }
 ```

 backend/internal/service/user.go

 ```go
   package service

   import "github.com/example/myapp-backend/internal/middleware"

   func CanDeleteUser(role string) bool {
      return middleware.Authorize(role)
   }
 ```

 backend/cmd/server/main.go

 ```go
   package main

   import (
      "github.com/example/myapp-backend/internal/middleware"
      "github.com/example/myapp-backend/internal/service"
   )

   func main() {
      if middleware.Authorize("admin") {
         _ = service.CanDeleteUser("admin")
      }
   }
 ```

 2. Index the monorepo root and wait for the graph:

 ```
   codebase_index   { "projectPath": "/abs/path/to/my-monorepo" }
   codebase_status  { "projectPath": "/abs/path/to/my-monorepo" }   # → green, graph ready
 ```

 3. Observe the broken Go graph:

 ```
   codebase_graph_query { "projectPath": "...", "filePath": "backend/internal/middleware/auth.go" }
 ```

 → No dependency information found for this file.

 ```
   codebase_graph_stats { "projectPath": "..." }
 ```

 → Go files present under Languages, but all listed under "Orphan files"; total edges cover only the frontend.

 ```
   codebase_impact { "projectPath": "...", "target": "Authorize" }
 ```

 → Total impacted files: 0 — No callers found. (incorrect — Authorize is called from cmd/server/main.go and
 internal/service/user.go.)

 4. Contrast — non-Go files in the same index resolve normally, confirming the builder works except for the nested Go module:

 ```
   codebase_graph_query { "projectPath": "...", "filePath": "frontend/src/app.ts" }
 ```

 → returns normal import/dependent edges.

 5. Confirmation that the go.mod location is the cause — index the backend alone, where go.mod is at the indexed root:

 ```
   codebase_index  { "projectPath": "/abs/path/to/my-monorepo/backend" }
   codebase_graph_stats { "projectPath": "/abs/path/to/my-monorepo/backend" }
   codebase_impact { "projectPath": "/abs/path/to/my-monorepo/backend", "target": "Authorize" }
 ```

 → Go edges populate and codebase_impact Authorize returns the real callers — proving the module path resolves correctly when  go.mod is at the root, and fails only when nested below it.

### Expected behavior

it should index Go projects when go.mod is in a nested directory.

### Actual behavior

It does not index Go projects when go.mod is in a nested directory.

### Embedding provider

Ollama (Docker — default)

### Qdrant mode

Managed (Docker — default)

### MCP host

VS Code (GitHub Copilot)

### Operating system

Linux

### Node.js version

v22.23.1

### Relevant log output

```shell

```

### Additional context

_No response_

### Checklist

hirisov:
The Go module resolver read go.mod only from the indexed project root, so any Go project whose go.mod sits in a subdirectory (e.g. <root>/backend/) produced an empty file graph and 0 impact edges.

 Discover every go.mod under the root via an independent, ignore-aware tree walk (go.mod is never in the graphable file set), then attribute each .go file to its deepest owning module and resolve imports against the matching module using a structural prefix check. A single root-level go.mod (#45) keeps resolving exactly as before.

 Refs #82

 Summary

 When a Go project's go.mod lives below the indexed root — the common monorepo layout — the module resolver never found it: it read <projectPath>/go.mod from exactly one location, got back null/empty, and resolved every Go import to null. The result was an empty Go file graph: every .go file an orphan with 0 connections, codebase_graph_query returning "No dependency information found", and codebase_impact reporting 0 callers even for plainly-called symbols. Non-Go files were unaffected, proving the graph builder itself was fine — it specifically failed for Go.

 This generalizes the single root-level read (from #45) to discover every owning go.mod in the tree, so both a root-level module and nested/monorepo modules resolve correctly. Because the symbol-level Impact Analysis path consumes these file-import edges, it recovers automatically — no changes were needed there.

 Changes

 - src/services/graph-resolution.ts
     - buildGoModuleInfo now returns GoModuleInfo[] (one entry per discovered module) instead of GoModuleInfo | null.
     - New findGoModFiles() walks the tree to discover every go.mod, independent of the graphable file set (go.mod has no AST grammar and is never admitted by getGraphableFiles — relying on fileSet silently produced 0 edges for every Go project). It reuses the same createIgnoreFilter / shouldIgnore conventions as getGraphableFiles, with the trailing-slash-for-dirs rule, so go.mod under node_modules/, .git/, or gitignored paths is skipped.
     - Each .go file is attributed to its deepest owning module (by directory depth, not string length — so the root module . never beats a short-named nested module like z).
     - resolveImport's go case picks the matching module by structural prefix (import equals the module path, or is module-path/...), preventing overlapping paths from misrouting — e.g. github.com/x/yother is not swallowed by module github.com/x/y.
     - packageMap keys are now module-relative (a Go import strips the module path down to a module-relative dir); values stay project-relative so no further translation is needed for nested modules.
 - src/services/code-graph.ts — Updated the call site and surrounding comments to reflect the array return type.
 - tests/unit/graph-resolution.test.ts — Added coverage for nested modules, deepest-owner attribution, structural-vs-textual prefix matching, and module-relative package lookup.
 - tests/unit/graph-discovery.test.ts — Added an e2e test that exercises the real file-discovery → graph-build pipeline (where go.mod is absent from the file set) for both root-level and nested layouts, asserting Go edges are produced.

 Type of change

 - [ ] Documentation update

 Testing

 - [x] Unit tests pass (npm run test:unit)
 - [x] Integration tests pass (npm run test:integration) — if applicable
 - [x] TypeScript compiles cleanly (npx tsc --noEmit)

 Added unit tests for: nested-module discovery, deepest-owner attribution (including the root-vs-nested tie-break), structural prefix matching (github.com/x/y vs github.com/x/yother), and module-relative package resolution. Added an e2e test that runs the real pipeline with go.mod absent from the graphable file set for both root-level and monorepo layouts.

 Checklist

 - [ ] I have addressed all CodeRabbit review comments (or marked as resolved with explanation)

 Related issues

 Fixes #82
 Relates to #45 (root-level go.mod support — this PR generalizes that single-read approach and preserves its behaviour)

## Summary by CodeRabbit

* **New Features**
  * Enhanced Go import resolution to support multi-module repositories, including nested `go.mod` discovery and correct module-root vs subpackage mapping.
* **Bug Fixes**
  * Corrected module ownership and import matching using deterministic, depth-based selection and structural (segment) prefix matching.
  * Fixed edge cases for symlinked `go.mod` files and ensured ignored directories (e.g., build folders) can’t override real modules.
* **Tests**
  * Added/updated regression and unit tests covering nested modules, tie-breaking behavior, missing/invalid `go.mod`, and end-to-end graph resolution.
* **Documentation**
  * Refreshed inline documentation for Go module-resolution behavior.

### DIFF
--- a/src/services/code-graph.ts
+++ b/src/services/code-graph.ts
@@ -703,11 +703,14 @@ export async function buildCodeGraph(
   const hasCs = files.some((f) => path.extname(f).toLowerCase() === ".cs");
   const csNamespaceMap = hasCs ? buildCsNamespaceMap(fileSet, resolvedPath) : undefined;
 
-  // Build Go module-resolution info from go.mod (issue #45). Without this,
-  // every Go import resolved to null and Go projects produced an empty
-  // file graph. The info is null when go.mod is missing or unparseable;
-  // the resolver treats null as "no Go resolution available" and behaves
-  // exactly as it did before this PR for those cases.
+  // Build Go module-resolution info from every go.mod in the tree (issue
+  // #45 for a root-level go.mod; issue #82 for nested modules in a
+  // monorepo). Without this, every Go import resolved to null and Go
+  // projects produced an empty file graph. buildGoModuleInfo discovers
+  // go.mod itself (independently of the graphable file set) and returns
+  // one entry per module, or an empty array when none parse; the resolver
+  // treats an empty/undefined result as "no Go resolution available" and
+  // behaves exactly as it did before this feature for those cases.
   const hasGo = files.some((f) => f.endsWith(".go"));
   const goModuleInfo = hasGo ? buildGoModuleInfo(fileSet, resolvedPath) : undefined;
 
--- a/src/services/graph-resolution.ts
+++ b/src/services/graph-resolution.ts
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 // Copyright (C) 2026 Giancarlo Erra - Altaire Limited
-import { type Dirent, readdirSync, readFileSync } from "node:fs";
+import { type Dirent, readdirSync, readFileSync, statSync } from "node:fs";
 import path from "node:path";
 import { toForwardSlash } from "../constants.js";
 import type { PathAliases } from "./graph-aliases.js";
@@ -300,8 +300,22 @@ function findGoModFiles(projectPath: string): string[] {
       if (shouldIgnore(ig, entry.isDirectory() ? `${relPath}/` : relPath)) continue;
       if (entry.isDirectory()) {
         walk(path.join(dir, entry.name));
-      } else if (entry.isFile() && entry.name === "go.mod") {
-        results.push(relPath);
+      } else if (entry.name === "go.mod") {
+        // readdirSync Dirents do not follow symlinks: a symlinked go.mod
+        // reports isFile()===false, so without this it would be neither
+        // recorded nor followed and a root-level symlinked go.mod would
+        // regress (the old single readFileSync followed the link). statSync
+        // resolves the link so a symlinked go.mod is discovered like a real
+        // one; broken links and non-file targets are skipped.
+        let isFile = entry.isFile();
+        if (!isFile && entry.isSymbolicLink()) {
+          try {
+            isFile = statSync(path.join(dir, entry.name)).isFile();
+          } catch {
+            continue;
+          }
+        }
+        if (isFile) results.push(relPath);
       }
     }
   };
--- a/src/services/graph-resolution.ts
+++ b/src/services/graph-resolution.ts
@@ -1,9 +1,10 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 // Copyright (C) 2026 Giancarlo Erra - Altaire Limited
-import { readFileSync } from "node:fs";
+import { type Dirent, readdirSync, readFileSync } from "node:fs";
 import path from "node:path";
 import { toForwardSlash } from "../constants.js";
 import type { PathAliases } from "./graph-aliases.js";
+import { createIgnoreFilter, shouldIgnore } from "./ignore.js";
 
 // ── Module resolution ────────────────────────────────────────────────────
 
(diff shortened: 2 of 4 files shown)


giancarloerra:
This is good, thanks.

What I ran through the real `getGraphableFiles` -> `buildCodeGraph` pipeline:

- Root-level `go.mod` (the #45 case): `main.go -> pkg/foo.go`. No regression.
- Nested `backend/go.mod` (your original #82 layout): 3 edges, including `backend/internal/service/user.go -> backend/internal/middleware/auth.go`.
- The tie-break case: `z/main.go -> z/svc/bar.go`, so a single-character nested module dir no longer loses to the root module.

In every case the probe printed `go.mod in graphable fileSet? false` while still discovering the module, which is the whole point: the walk is genuinely independent of the file set. `tsc`, `biome`, and the full unit suite (937 tests) are clean, and CI is green.

The e2e test in `graph-discovery.test.ts` that drives the real discovery -> graph-build path is what makes this trustworthy going forward, so thanks for adding it. That was the gap that let the previous attempt through with green units.

Three small things I think is worth fixing:

1. **A symlinked `go.mod` is no longer found (small regression).** `findGoModFiles` gates on `entry.isFile()`, but under `readdirSync({ withFileTypes: true })` a symlink Dirent reports `isFile() === false`, so it is neither recorded nor followed. Repro: make `<root>/go.mod` a symlink to a file outside the tree. On this branch the graph comes back with 0 edges; on `main` the same fixture resolves, because the old `readFileSync` followed symlinks. Accepting `entry.isSymbolicLink()` and confirming with `statSync` should cover it.

2. **The ignore-filter branch has no test.** I stubbed it to a no-op and all 937 tests still passed, yet it is load-bearing: a stray `go.mod` under a default-ignored directory (say `build/`) can win module selection with an empty package map and silently drop every edge under that prefix, which is the same silent-zero-edge class this PR fixes. One more case in the e2e block would lock it in.

3. **Cost on vendored repos (follow-up, not for this PR).** `findGoModFiles` builds its own ignore filter rather than reusing the one `getGraphableFiles` just created, and `createIgnoreFilter` walks the whole tree with a skip list that doesn't include `vendor`. On a plain nested monorepo that is roughly 5% of `buildCodeGraph`, but on a large vendored tree I measured 28% (5.7k dirs) and 37% (26k dirs), about 1.2 to 1.9s. Building the filter once and threading it into both callers would fix it. Happy for this to be a separate issue.

Let me know what you think.

hirisov:
Thanks for the thorough review — really appreciate you running it through the real pipeline, especially catching that the green-unit-but-broken-pipeline gap was what let the first attempt through.

 All three addressed:

 1. Symlinked go.mod — good catch, that's a genuine regression I introduced. Fixed in 7b966e7: findGoModFiles now accepts symlinked go.mod entries and confirms the target via statSync (broken links skipped), so it matches the old follow-the-link behaviour. Added an e2e test that symlinks go.mod to a file outside the tree and asserts the edge resolves.

 2. Ignore-filter branch test — added an e2e case in graph-discovery.test.ts: a stray go.mod under build/ declaring the same module path as the root, so if discovery ever bypassed shouldIgnore it shadows the root module (empty package map) and silently drops every edge — exactly the failure mode you described. I confirmed it's load-bearing the way you did: stubbing shouldIgnore to a no-op fails it with expected false to be true.

 3. Double filter construction / vendor cost — agreed, and happy to keep it out of this PR. 

 tsc, biome, and the full suite are clean (939 unit / 166 integration).

giancarloerra:
@coderabbitai full review

PR #88, opened 2026-07-23
fix(graph): label extensionless files by detected language in graph stats and visualization

shaitourchin:
## What & why

Grammar-bearing extensionless files (e.g. `#!/bin/bash` scripts, no-shebang Python) are indexed and labelled with the correct language in search and the symbol graph, but three **display-only** sites re-derive the language from the file *path* with no content in hand. Since `path.extname("")` is `""`, `getLanguageFromExtension("")` returns `"plaintext"` — so a detected shell script shows as `shell` in `codebase_search` yet `plaintext` in the graph stats and visualisation:

- `getGraphStats().languageBreakdown` — surfaced by `codebase_graph_stats`
- `generateMermaidDiagram` node colour-coding — `codebase_graph_visualize`
- `buildFileVizData` node language — `codebase_graph_visualize mode="interactive"`

This is the cosmetic follow-up called out as a non-goal in the extensionless-file indexing work (#81, released in 1.9.0).

## Change

- Add an optional `language?: string` to `CodeGraphNode`.
- Populate it once in `buildCodeGraph`, at the grammar-bearing node-creation site, from the already-computed **post-detection** language — no extra I/O.
- Add a `nodeLanguage(node)` helper that prefers `node.language` and falls back to path-based derivation for nodes without it (older persisted graphs and import-target-only leaf nodes). The three display sites now go through it, which also removes the path→language derivation that was previously copy-pasted across all three.

Rejected alternative: head-reading file content at render time — adds async I/O and file re-reads to synchronous display/stat paths.

## Backward compatibility

`language` is optional and additive. Graphs persist as whole-graph JSON (`saveGraphData`/`loadGraphData`), so the field round-trips with **no persistence change**, and graphs saved before this change deserialise with `language` absent → fallback → identical to current behaviour. Extensioned files are unaffected (their stored language equals the path-derived fallback under a stable extension map). The corrected label surfaces for extensionless nodes after a graph rebuild.

## Scope

Display/stats only. Search, `codebase_symbol` / `codebase_impact` / `codebase_flow`, graph edges, symbols, chunking, and embeddings are already correct and untouched.

## Tests

Unit tests, no Docker required:

- `nodeLanguage` — stored-language and both fallback branches.
- `buildCodeGraph` stamps `shell` on a real `#!/bin/bash` fixture (real content detection, no mocks).
- Stats breakdown, mermaid colour, and interactive-viz language — each for both the stored-language and fallback cases.

Full unit suite green (`npm run test:unit`), `npm run lint` (biome) and `npm run build` (tsc) clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved language detection for extensionless files, preventing incorrect `plaintext` labels when a language can be identified.
  - Updated language statistics, Mermaid diagrams, and interactive graph displays to use detected languages consistently.
  - Added safe fallback behavior for files without stored or path-derived language information.

- **Tests**
  - Added coverage for language detection, statistics, diagram coloring, and interactive graph metadata.

### DIFF
--- a/src/services/code-graph.ts
+++ b/src/services/code-graph.ts
@@ -778,6 +778,10 @@ export async function buildCodeGraph(
     }
     const node = nodesMap.get(relPath);
     if (!node) continue;
+    // Record the (post-detection) language so display/stats sites don't have to
+    // re-derive it from the path — which silently mislabels extensionless files
+    // as plaintext.
+    node.language = language;
 
     // Extract imports using ast-grep
     const importInfos = extractImports(source, lang, ext);
--- a/src/services/graph-analysis.ts
+++ b/src/services/graph-analysis.ts
@@ -2,7 +2,17 @@
 // Copyright (C) 2026 Giancarlo Erra - Altaire Limited
 import path from "node:path";
 import { getLanguageFromExtension, toForwardSlash } from "../constants.js";
-import type { CodeGraph } from "../types.js";
+import type { CodeGraph, CodeGraphNode } from "../types.js";
+
+/**
+ * Resolve a graph node's language for display/stats. Prefers the language stored
+ * on the node at build time (correct for extensionless files, whose path carries
+ * no extension); falls back to deriving it from the path extension for nodes with
+ * no stored language (older persisted graphs and import-target-only leaf nodes).
+ */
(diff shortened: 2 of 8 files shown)


giancarloerra:
Thank you, well found and nicely done, merging now!

PR #90, opened 2026-07-26
fix(graph): count skipped files and label extensionless nodes

shaitourchin:
## Summary

`buildCodeGraph` dropped files from the graph with no trace. The per-file read guard swallowed every
`fs` error in a bare `catch`, and neither the oversized skip nor that catch incremented
`filesProcessed` — so a build that dropped files reported a percentage that never reached 100, and
nothing anywhere reported that the graph was incomplete.

Fixing that surfaced a second problem in the same loop: extensionless files were head-read a second
time to recover their language, and the label an *import-target* placeholder got depended on whether
the target happened to be processed before its importer. A grammar-bearing extensionless file that
was an import target and also oversized or unreadable showed up as `plaintext` in graph stats and
visualization.

This PR makes every abandoned file counted and logged with a reason, carries content detection from
discovery into the build instead of re-reading, and routes every extensionless-language decision
through one shared byte window, so discovery, chunking, the graph build, the watcher, and the
incremental symbol-graph update cannot disagree about the language of the same file. That is four call
sites of the new helper — one of them the disk-side resolver that discovery and the incremental update
both go through.

## Reviewing this

Five commits, each self-contained and each typechecking on its own, so they read in order:

1. `fix(graph): carry detected language from discovery into graph build`
2. `fix(graph): count and log files skipped during code graph build`
3. `fix(graph): log unreadable files in the incremental symbol-graph update`
4. `docs: document graph-build skip accounting and discovery ordering`
5. `fix(watcher): score the extensionless watch check on the shared byte window`

Commit 1 is the largest and the one worth reading closely — it carries the discovery→build handoff, the
shared detection window, and the sort. Commit 2 is the skip accounting proper. The rest are small.

## Changes

- **Detected language flows from discovery into the build.** `getGraphableFiles` now returns
  `{ files, detectedExts }`; the build loop reuses that detection instead of head-reading again, and
  import-target placeholders take their language from the map at creation time (so the label no
  longer depends on processing order). Discovery output is also sorted — see reviewer notes.
- **One shared detection window.** New `detectExtensionFromSource` scores an 8192-**byte** UTF-8
  window (bytes, not characters, so multibyte content cannot shift the score). The chunker's inline
  copy of that computation and the watcher's raw `readFileHead` call both fold into it. Detection also
  re-runs on the bytes about to be parsed rather than on a separate head-read, closing the window the
  old re-detect left open; what happens when it disagrees with discovery is the next bullet.
- **Skip accounting.** Every path that abandons a file routes through one `recordSkip` helper: logs
  the reason at `debug`, counts the file toward `filesProcessed`, and increments a new optional
  `filesSkipped`. Four reasons: `oversized`, `vanished` (ENOENT between discovery and read),
  `read-failed` (any other fault — EACCES/EIO/EMFILE, logged with the error text), `content-changed`
  (extensionless only: re-detection on the parsed bytes no longer yields a grammar-bearing language,
  so there is nothing to parse it with).
- **Re-detection adopts the fresh answer.** An extensionless file is re-detected on the bytes about to
  be parsed and parsed under the grammar *those* bytes call for, so content that changed since
  discovery is followed rather than dropped — the fresh detection describes what is in hand, so it
  supersedes discovery's rather than being compared against it. This keeps the graph aligned with the
  chunker, which has always adopted the fresh detection and bails only when detection fails.
- **Where the counts surface.** Per-reason breakdown plus total in the existing `"Code graph built"`
  log (which runs at `info`, the default level); total in `codebase_graph_status` while building, in
  its READY block, and in the background-build completion log. The count is omitted entirely when
  nothing was skipped.
- **A directory the walk cannot read is no longer silent.** `getGraphableFiles`' `walk` swallowed
  `readdir` failures in a bare `catch`, so one EACCES/EIO directory removed its entire subtree from the
  file list — a larger blast radius than any per-file reason, and the files leave the walk before they
  have a path of their own, so none of them can reach `recordSkip`. It now logs the directory and the
  error at `debug` (ENOENT stays quiet: a directory removed mid-walk has nothing left to graph).
  Control flow is unchanged — the parent keeps walking, as before.
- **Watcher path.** The `readFile` catch in the incremental symbol-graph update's changed-files loop
  swallowed every error, so a file whose symbols could not be refreshed kept a stale payload with no
  trace — on every change. Control flow is deliberately unchanged; only the ENOENT split was
  missing, mirroring the catch forty lines above it. ENOENT stays quiet because a real delete is
  reconciled through `removedRelPaths`.
- **Docs.** Four `DEVELOPER.md` passages: the `filesSkipped` counter and its four reasons, that
  discovery returns detected extensions and sorts its output, and the skipped count in
  `codebase_graph_status`.
- Also drops an unreachable node-lookup guard in the build loop by holding the node reference instead
  of re-reading it from the map.

### Behaviour a reviewer may notice

- `codebase_graph_status` gains a skipped count (only when non-zero); `GraphBuildProgress` gains an
  optional `filesSkipped`.
- Extensionless import targets that are oversized or unreadable now carry their detected language
  instead of `plaintext`.
- `buildJvmSuffixMap`'s duplicate class-path tie-break moves from traversal order to lexicographic.
- **Extensionless files with invalid UTF-8 in their head are admitted on a narrower window than
  before, at every detection site.** This one is worth a close look, since it is the only place the PR
  narrows behaviour rather than preserving it. A lossily-decoded head re-encodes to three bytes per
  U+FFFD, so the shared window covers roughly `DETECT_HEAD_BYTES / 3` characters of such a file: a
  latin-1 script whose only code markers sit past ~2730 characters detected as code before and does
  not now. Measured on a 3051-byte latin-1 shell script that fits entirely inside the raw 8 KB head —
  `main` answers `.sh`, this branch answers `null`, and an ASCII file of identical length still
  answers `.sh`. The narrowing is not incidental: a decoded string cannot recover the bytes it came
  from, so the in-memory and on-disk sides cannot be made to score the same span of a lossy file, and
  the only way to stop them disagreeing is to put both on the narrower window. Pinned by a test, and
  the reasoning is in `detectExtensionFromSource`'s JSDoc.

## Type of change

- [x] Documentation update

## Testing

Full suite on Node 22: **53 files / 1160 tests passing**, `npm run lint` clean with zero warnings,
`npm run build` (tsc + asset copy) clean.

- [x] Unit tests pass (`npm run test:unit`)
- [x] Integration tests pass (`npm run test:integration`)

New coverage: `tests/unit/code-graph-skips.test.ts` (oversized with no importer; oversized
extensionless import target — asserts both the skip count and that the placeholder keeps its detected
language; unreadable file, including that it is classified `read-failed` and not `vanished`; a file
whose head and tail read as different languages is kept under its head's language; clean project
leaves `filesSkipped` undefined), plus additions to `code-graph-progress.test.ts`,
`code-graph-node-language.test.ts`, `extensionless.test.ts`, `graph-discovery.test.ts`,
`watcher.test.ts`, and `tests/integration/symbol-graph-incremental.test.ts`.

`graph-discovery.test.ts` also covers the unreadable directory, asserting both that its subtree is
omitted and that the omission is logged — the log being the only observable, since those files cannot
reach the skip counter.

Each new ordering/window test was checked by mutation, not just by passing: removing `files.sort()`
fails the interleaving and JVM tie-break tests; reverting either detection call site to raw scoring
fails the lossy-decode tests; dropping the window in the graph's re-detection fails the head/tail test;
restoring the bare `catch` fails the unreadable-directory test.

## Checklist

## Reviewer notes

Decisions below were each made deliberately during implementation, and several of them are the kind a
reviewer would otherwise have to reverse-engineer. Pushing back on any of them is welcome — but the
reasoning is here so it does not have to be rediscovered.

**Why discovery sorts.** Not because any filesystem was observed returning hash order — that premise
did not reproduce on APFS, where each `readdir` is already sorted. Two other reasons stand on their
own: Node documents *no* `readdir` ordering at all, and a depth-first walk is not lexicographic even
when every `readdir` is, because it yields a directory's contents before the sibling entries that sort
after it (`/` 0x2F sorts after `.` 0x2E, so `a/x.ts` comes before `a.ts`). That second property is
filesystem-independent, which makes it testable anywhere — so the sort is pinned by a fixture of
`a/x.py` plus `a.py` that fails on APFS with `files.sort()` removed, rather than only by the
reverse-lexicographic-creation test, which cannot fail where `readdir` is pre-sorted (that one is kept
for platforms yielding creation order). `buildGoModuleInfo` and `buildCsNamespaceMap` already sorted
their own views defensively; this hoists the guarantee up to discovery, where processing order drives
node insertion order and `buildJvmSuffixMap`'s duplicate-class tie-break.

**The JVM tie-break is now pinned.** Sorting discovery output means `buildJvmSuffixMap`'s tie-break for
a class path two modules both provide is defined (lexicographic) rather than arbitrary-by-traversal —
pre-existing nondeterminism this PR settles as a side effect. The test runs through `buildCodeGraph`
with modules `mod` and `mod-b`: `mod-` sorts before `mod/` while the walk descends `mod` first, so the
two orders disagree and the assertion fails if either the sort or the tie-break changes.

**`src/services/graph-analysis.ts` and `src/services/indexer.ts` are touched even though neither is
about skip accounting.** Both are consequences of the main change, not drive-by edits:
`graph-analysis.ts` (and the `CodeGraphNode.language` JSDoc in `types.ts`) documents when the stored
language is absent, and that set changed once placeholders started carrying a detected language;
`indexer.ts` loses its inline copy of the head-window computation to the new shared helper — the
whole point of introducing the helper was that the copies could drift.

**Four skip reasons, not five — the `!detected || !detectedLang` guard is type-narrowing only.** An
earlier iteration counted that guard as a fifth reason. Discovery admits an extensionless path *only*
together with a grammar-bearing detection and records it in `detectedExts` in the same step, so the
state cannot occur; the guard exists because TypeScript cannot infer `detected` is defined from
`detectedLang` being non-null. It `continue`s without counting, and its comment says so. Accepted
consequence: if the invariant were ever broken, that path would be an uncounted skip (and would leave
`filesProcessed` short of `filesTotal`) rather than a counted one.

Worth not confusing it with the re-detect guard further down the loop, which is shaped almost
identically but *does* count, as `content-changed`. The difference is reachability: the re-detect guard
tests content the build actually read, which can genuinely have changed since discovery, while this one
tests a state discovery cannot hand it.

**`vanished` has no direct test, and `content-changed` is pinned only in one direction.** `vanished` is
not reachable through the public API and shares its catch with `read-failed`, which is covered
end-to-end including its log payload. For `content-changed`, the direction that is reachable without a
real race — a file whose head and tail read as different languages must *not* be skipped, because both
sides score the same window — has a test; actually firing the reason needs the content to change
between two reads inside one build, so it is exercised through `detectExtensionFromSource`'s own unit
tests rather than through `buildCodeGraph`.

**One adjacent gap left deliberately unfixed: an unreadable *extensionless* file is not counted.** At
discovery, a head-read failure on an extensionless file collapses into `null`, which the call site
cannot distinguish from "readable, but prose" — so the file is never admitted and never counted, while
the byte-identical file *with* a suffix reaches the build's read guard and is counted as `read-failed`.
Whether a permission fault reaches the skip total therefore depends on whether the file has an
extension.

It is not silent, which is why it is not urgent: `resolveExtensionlessExtension` already logs the
non-ENOENT fault at `debug` with the path and the error, at the same level as every other per-file skip
detail. What is missing is only the aggregate count, and adding it means threading a second counter out
of `getGraphableFiles` and deciding how a discovery-phase failure should render in
`codebase_graph_status` — a design question this PR's per-file funnel, which keys on a project-relative
path, deliberately does not answer. Happy to open it as a follow-up.

**Two related fixes were found while doing this and are deliberately held back.** Both are language-
labelling bugs in the same area this PR touches, and both are being kept out of it so this stays one
reviewable change rather than a grab-bag:

1. **`fix/shell-source-resolution`** — shell `source` / `.` directives are extracted as imports but do
   not resolve to a project file, so the file graph loses those edges. A three-line fix in
   `src/services/graph-resolution.ts` plus tests.
2. **`fix/stylus-language-and-indexing`** — `.styl` files are labelled `plaintext` instead of `stylus`
   and are not indexed at all. A `src/constants.ts` change plus tests and doc updates.

Neither depends on this PR and this PR does not depend on either — all three sit independently on the
same base, so they can be reviewed and merged in any order. Two textual overlaps to be aware of, both
mechanical keep-both-sides resolutions for whichever lands second:
`tests/unit/code-graph-node-language.test.ts` (each branch appends fixtures to the same `beforeAll` and
cases to the same `describe`) and, for the Stylus one, `DEVELOPER.md` (this PR rewrites the
skip-accounting and discovery passages; that one edits the extension table and its count).

I can open either as its own PR on request, or fold one in here if you would rather review them
together.

**Knowingly left alone.** These were raised in review and declined as out of scope — each is a
pre-existing shape this PR does not make worse:

- `graph-tools.ts` renders "N skipped" in three places rather than through one formatter, and those
  output strings have no assertions.
- The three `CodeGraphNode` object literals in `buildCodeGraph` are not unified behind a factory; one
  of the three is pre-existing code this PR does not otherwise touch.
- `isIndexableFile` still re-implements `resolveExtensionlessExtensionStrict` inline. Collapsing it is
  a ~29-line rewrite of a function this PR otherwise leaves alone, and it would merge two distinct
  debug messages into one.
- The two catch blocks in `symbol-graph-incremental.ts` carry near-identical rationale comments. The
  mirroring is deliberate — the second catch's whole justification is that it matches the first.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling and language detection for extensionless files during indexing and graph building.
  * Preserved existing symbol data when temporary file-read failures occur.
  * Added clearer handling for skipped, oversized, unreadable, or vanished files.
  * Stabilized file ordering and dependency resolution results.

* **New Features**
  * Graph build progress and completion status now report skipped file counts and reasons.
  * Placeholder graph nodes retain detected language information when source files are skipped.

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -456,12 +456,20 @@ When `codebase_graph_build` is called:
 1. BACKGROUND EXECUTION (fire-and-forget)
    ├── Tool returns immediately with "build started" message
    ├── Actual build runs asynchronously on the event loop
-   ├── Progress tracked via GraphBuildProgress { filesTotal, filesProcessed, phase }
+   ├── Progress tracked via GraphBuildProgress { filesTotal, filesProcessed, filesSkipped, phase }
    └── Client polls codebase_graph_status for progress %
 
 2. FILE DISCOVERY (phase: "scanning files")
    ├── Get graphable files from project (same ignore filters as indexing)
-   └── Include files with AST-grep grammar + files with extra extensions
+   ├── Include files with AST-grep grammar + files with extra extensions
+   ├── Extensionless files: admitted only when content detection on the first
+   │   8 KB yields a grammar-bearing extension; that extension is returned
+   │   alongside the path so the parse step need not head-read them again
+   └── Sort the file list lexicographically — Node documents no readdir ordering,
+       and a depth-first walk additionally interleaves a directory's contents with
+       the sibling entries that sort after it (a/x.ts before a.ts), while
+       processing order drives node insertion order and the JVM suffix map's
+       duplicate-class tie-break
 
 3. PARSE IMPORTS (phase: "analyzing imports", per file, via ast-grep)
    ├── Determine AST-grep language from file extension
@@ -486,6 +494,21 @@ When `codebase_graph_build` is called:
    │   └── CSS/SCSS/SASS/LESS: @import/@import url()/@require regex extraction
    ├── Tag CSS imports with isCssImport flag (for correct resolution extensions)
    ├── Update progress: filesProcessed++
+   ├── Extensionless files are re-detected on the bytes about to be parsed and
+   │   parsed under the grammar those bytes call for, so content that changed
+   │   since discovery is followed rather than dropped
+   ├── Skipped files count toward filesProcessed too and increment filesSkipped,
+   │   with the reason logged at debug and a per-reason breakdown in the
+   │   "Code graph built" log: oversized (> MAX_GRAPH_FILE_BYTES), vanished
+   │   (ENOENT between discovery and read), read-failed (any other fault, e.g.
+   │   EACCES/EIO/EMFILE), content-changed (extensionless only: re-detection on
+   │   the parsed bytes no longer yields a grammar-bearing language, so there is
+   │   no grammar to parse it with). A skipped file builds no node of its own and
+   │   contributes no outgoing edges; if another file imports it, the importer's
+   │   placeholder node still represents it — labelled by the discovery-detected
+   │   language when the target was extensionless, and left unlabelled otherwise
+   │   so nodeLanguage() derives it from the path, since detectedExts holds no
+   │   entry for an extensioned file.
    └── Return ImportInfo[] with module specifiers
 
 4. RESOLVE IMPORTS
@@ -1164,8 +1187,8 @@ Parameters:
   projectPath?: string  — Absolute path (defaults to cwd)
 
 Returns:
-  If build in progress: Status BUILDING with phase, progress %, elapsed time
-  If ready: Status READY with node/edge count, last built time, cache status, last build duration
+  If build in progress: Status BUILDING with phase, progress % (and a skipped count when non-zero), elapsed time
+  If ready: Status READY with node/edge count, last built time, cache status, last build duration, and files skipped when non-zero
   If not found: Instructions to build
 ```
 
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -401,9 +401,10 @@ export function getLanguageFromExtension(
 // ── Extensionless file detection ──────────────────────────────────────────
 
 /**
- * Bytes of file head inspected for extensionless language detection. Shared by
- * the discovery head-read (`readFileHead`) and the chunk-time re-detection so
- * the two paths always look at the same window and agree on the language.
+ * Bytes of file head inspected for extensionless language detection. Read by the
+ * two functions that bound a detection window — `readFileHead` on disk and
+ * `detectExtensionFromSource` in memory — so every site that decides an
+ * extensionless file's language scores the same bytes and cannot disagree about it.
  */
 export const DETECT_HEAD_BYTES = 8192;
 
--- a/src/services/code-graph.ts
+++ b/src/services/code-graph.ts
@@ -11,7 +11,7 @@ import type {
   CodeGraph, CodeGraphEdge, CodeGraphNode,
   SymbolEdge, SymbolGraphFilePayload, SymbolGraphMeta, SymbolNode, SymbolRef,
 } from "../types.js";
-import { resolveExtensionlessExtension } from "./extensionless.js";
+import { detectExtensionFromSource, resolveExtensionlessExtension } from "./extensionless.js";
 import { loadPathAliases } from "./graph-aliases.js";
 import { extractImports } from "./graph-imports.js";
 import { buildCsNamespaceMap, buildGoModuleInfo, buildJvmSuffixMap, resolveImport } from "./graph-resolution.js";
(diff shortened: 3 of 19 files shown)


giancarloerra:
Thanks for this, merged!

PR #93, opened 2026-07-29
fix(indexer): never emit chunks with empty content

Dewars30:
## Summary

`chunkFileContent` can emit chunks whose content is `""`. Because such a chunk embeds as a
path-only document, BM25 length-normalisation floats it to **rank 1** — so a good query returns a
blank top hit.

On a real four-repo corpus this was **1,371 of 25,065 chunks (5.5%)**, and it was invisible: the
chunks index and search cleanly, they just carry no text.

Three code paths produce them. Two are the same tautology — guarding on **array** length where
**content** length was meant:

```ts
// epilogue — lines.slice(lastEnd) on a trailing "\n" yields [""],
// a non-empty ARRAY of empty CONTENT, so this is always true
if (epilogueLines.length > 0) { … }        // → content: "" at line N+1

// preamble — lines.slice(0, startLine) with startLine > 0 is never empty
if (preambleLines.length > 0) { … }        // → content: "" at 1-1
```

The third is zero-byte and whitespace-only files (Python package markers like `__init__.py` are
the common real case), which reach `chunkByLines` and yield one blank chunk.

Minimal repro against `v1.9.2`:

```js
import { chunkFileContent } from "socraticode/dist/services/indexer.js";
import { ensureDynamicLanguages } from "socraticode/dist/services/code-graph.js";
await ensureDynamicLanguages();

const fns = Array.from({ length: 40 }, (_, i) =>
  `export function fn${i}(): void {\n  const v = ${i};\n}`).join("\n\n");

chunkFileContent("/t/a.ts", "a.ts", fns + "\n")          // epilogue  → 1 chunk with content ""
chunkFileContent("/t/b.ts", "b.ts", "\n" + fns + "\n")   // preamble  → 1 chunk with content ""
chunkFileContent("/t/__init__.py", "__init__.py", "")    // zero-byte → 1 chunk with content ""
```

## Changes

- Fix the epilogue guard to test content, not array length: `epilogueLines.some((line) => line.length > 0)`
- Fix the preamble guard the same way: `preambleLines.some((line) => line.length > 0)`
- Add a terminal invariant in `applyCharCap`: `chunks.filter((c) => c.content.trim().length > 0)`.
  Four of the five `return` paths in `chunkFileContent` pass through `applyCharCap` (the fifth
  returns `[]`), making it the one place that can guarantee the property for every chunking
  strategy — including `chunkByLines`, which the two guard fixes do not reach.
- Add four regression tests: one per site, plus one guarding against an over-broad filter.

I went with all three rather than only the two guards because fixing the guards alone still leaves
zero-byte files emitting blank chunks — the invariant would be "almost never empty", which is not
worth asserting. If you would prefer the guards alone without the `applyCharCap` change, that is an
easy trim; happy to resubmit.

## Type of change

- [ ] Documentation update

## Testing

- [x] Unit tests pass (`npm run test:unit`) — 44 files, 992 tests
- [ ] Integration tests pass (`npm run test:integration`) — not run; requires Docker, and this
      change is confined to pure chunking logic with no I/O or Qdrant surface
- [x] `npm run lint` (biome) clean — 100 files

The three new tests fail on `v1.9.2` and pass with the fix. The fourth passes both ways and exists
so an over-broad filter cannot silently empty the index:

```
× emits no blank chunk for a newline-terminated file (epilogue guard)
× emits no blank chunk for a file whose first line is blank (preamble guard)
× emits no chunks at all for zero-byte and whitespace-only files
  Tests  3 failed | 46 passed (49)     ← before
  Tests  49 passed (49)                ← after
```

### No re-index required

`chunkId` is `sha256(relativePath + ":" + startLine)` — position-derived, not sequential — so
dropping a chunk never renumbers another. Every surviving chunk keeps its exact id, which means an
existing index can be repaired with a filtered delete instead of a re-embed.

Measured by chunking four repos via `getIndexableFiles` under both builds, keyed **per repo**
(chunk ids are only unique within a collection — pooling repos produces false collisions):

| | pristine | fixed |
|---|---:|---:|
| chunks | 25,065 | 23,694 |
| removed | — | 1,371 (**all blank**) |
| added | — | **0** |
| contents changed | — | **0** |
| endLines changed | — | **0** |
| residual blank | — | **0** |

Per repo: 21,489→20,356 · 2,974→2,762 · 345→322 · 257→254. Cross-validated against a live Qdrant
instance — predicted post-fix counts of 2,762 / 322 / 254 match the actual point counts exactly for
the three collections that had been re-indexed.

### A variant I did not submit

There is a `lines.pop()` form that also corrects a genuine `endLine` off-by-one on small files. It
is arguably more correct, but measured **1,946 changed endLines, 466 changed chunk contents and 31
restructured files** on the same corpus — churn that only pays for itself alongside a re-index. I am
flagging it because it is the better fix if you are willing to take that cost, and submitting the
cheaper one silently would misrepresent the tradeoff. Happy to send it instead.

## Checklist

- [ ] I have updated documentation (README.md / DEVELOPER.md) if needed — no user-facing behaviour
      or configuration option changes, so nothing appeared to warrant an edit
- [ ] I have addressed all [CodeRabbit](https://coderabbit.ai) review comments — none yet

## Related issues

None that I could find — `gh issue list --state all` returns 32 issues for this repo, and none
mention empty chunks, blank results, or the epilogue/preamble guards. Filing as a fresh report
rather than linking an existing one.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured generated chunks never include blank or whitespace-only content, including cases with trailing newlines and leading blank lines.
  * Prevented empty chunks from being produced during chunking and after character-level truncation.
  * Preserved chunking for large inputs by truncating when needed rather than dropping content.
* **Tests**
  * Added unit tests to enforce the “no empty chunks” invariant and cover truncation and chunking-order edge cases.

### DIFF
--- a/src/services/indexer.ts
+++ b/src/services/indexer.ts
@@ -360,13 +360,19 @@ function applyCharCap(chunks: FileChunk[]): FileChunk[] {
   // which reach chunkByLines and would otherwise yield a single blank chunk.
   // Safe to drop: chunk ids are sha256(relativePath + ":" + startLine), so
   // removing a chunk never renumbers any other.
-  chunks = chunks.filter((c) => c.content.trim().length > 0);
-  if (chunks.every((c) => c.content.length <= MAX_CHUNK_CHARS)) return chunks;
-  return chunks.map((c) =>
-    c.content.length > MAX_CHUNK_CHARS
-      ? { ...c, content: c.content.substring(0, MAX_CHUNK_CHARS) }
-      : c,
-  );
+  //
+  // ORDER MATTERS: cap first, then filter. Filtering first lets a chunk whose only
+  // non-whitespace content sits past MAX_CHUNK_CHARS pass the filter and then be
+  // truncated back into a blank chunk, so the invariant would not actually hold.
+  const capped =
+    chunks.every((c) => c.content.length <= MAX_CHUNK_CHARS)
+      ? chunks
+      : chunks.map((c) =>
+          c.content.length > MAX_CHUNK_CHARS
+            ? { ...c, content: c.content.substring(0, MAX_CHUNK_CHARS) }
+            : c,
+        );
+  return capped.filter((c) => c.content.trim().length > 0);
 }
 
 /**
--- a/src/services/indexer.ts
+++ b/src/services/indexer.ts
@@ -353,6 +353,14 @@ function findAstBoundaries(source: string, lang: Lang | string): AstRegion[] {
  * defence; this cap ensures chunks are already within bounds before that.
  */
 function applyCharCap(chunks: FileChunk[]): FileChunk[] {
+  // Terminal invariant: never emit a chunk with no non-whitespace content.
+  // Four of the five `return` paths in chunkFileContent pass through here (the
+  // fifth returns []), so this is the one place that can guarantee the property
+  // for every chunking strategy — including zero-byte and whitespace-only files,
+  // which reach chunkByLines and would otherwise yield a single blank chunk.
+  // Safe to drop: chunk ids are sha256(relativePath + ":" + startLine), so
+  // removing a chunk never renumbers any other.
+  chunks = chunks.filter((c) => c.content.trim().length > 0);
   if (chunks.every((c) => c.content.length <= MAX_CHUNK_CHARS)) return chunks;
   return chunks.map((c) =>
     c.content.length > MAX_CHUNK_CHARS
@@ -506,7 +514,7 @@ function chunkByAstRegions(
   // Preamble: everything before the first declaration (imports, constants, comments)
   if (regions[0].startLine > 0) {
     const preambleLines = lines.slice(0, regions[0].startLine);
-    if (preambleLines.length > 0) {
+    if (preambleLines.some((line) => line.length > 0)) {
       chunks.push({
         id: chunkId(relativePath, 1),
         filePath,
@@ -592,7 +600,7 @@ function chunkByAstRegions(
   const lastEnd = regions[regions.length - 1].endLine;
   if (lastEnd < lines.length) {
     const epilogueLines = lines.slice(lastEnd);
-    if (epilogueLines.length > 0) {
+    if (epilogueLines.some((line) => line.length > 0)) {
       chunks.push({
         id: chunkId(relativePath, lastEnd + 1),
         filePath,
--- a/tests/unit/indexer.test.ts
+++ b/tests/unit/indexer.test.ts
@@ -232,7 +232,7 @@ describe("indexer utilities", () => {
     // chunk embeds as a path-only document, BM25 length-normalisation floats it
     // to rank 1 — a blank top hit for an otherwise good query.
 
-    it("emits no blank chunk for a newline-terminated file (epilogue guard)", () => {
+    it("emits no blank chunk for a newline-terminated file", () => {
       // `lines.slice(lastEnd)` on a trailing "\n" yields [""], which is a
       // non-empty ARRAY of empty CONTENT. Guarding on array length is therefore
       // always true and emits content:"" at line N+1.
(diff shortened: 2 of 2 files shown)


giancarloerra:
Thanks for this, it's an unusually well-diagnosed report: the corpus measurement, the exact repro, and the reasoning about `applyCharCap` being the one place that covers every strategy all made this quick to verify.

I confirmed the bug against v1.9.2 with your repro, and reproduced all three sites:

| Case | v1.9.2 | this PR |
|---|---|---|
| newline-terminated file (epilogue) | 1 blank chunk | 0 |
| leading blank line (preamble) | 2 blank chunks | 0 |
| zero-byte `__init__.py` | 1 blank chunk | 0 chunks |
| whitespace-only file | 1 blank chunk | 0 chunks |

I also drove a real end-to-end index with those files: nothing blank is stored, zero-chunk files are simply absent, and a following `updateProjectIndex` reports 0 added / 0 updated / 0 removed, so their hashes are still recorded and they don't churn. Unit and integration suites are green.

**One thing to fix before I merge.** The blank filter runs before the character cap, so a chunk whose non-whitespace content sits past `MAX_CHUNK_CHARS` passes the filter and is then truncated back into a blank chunk. The terminal invariant doesn't quite hold. Constructed case (100 lines of 23 spaces, code on the last line):

chunk line 1-100 len=2000 trimmed="" <-- survives
chunk line 91-101 len=228 trimmed="realCode();"


CodeRabbit posted the same finding with the right shape: cap first, then filter (or filter on both sides). It's a two-line change to `applyCharCap`.

**One observation, no action needed unless you want it.** The two guard fixes turn out to be behaviourally redundant with the terminal filter. I reverted both guards and all 49 tests still passed, because `applyCharCap` catches those chunks regardless; only removing the filter fails a test. Your PR description already reasons about this layering, so I'm not asking you to drop the guards, I think they're worth keeping since they avoid building a chunk only to discard it. It does mean the "epilogue guard" and "preamble guard" tests don't isolate the guards they're named after. If you want them to bite, they'd need to assert on `chunkByAstRegions` output before the cap, or you could rename them to describe the observable behaviour they actually pin.

To answer your question in the description: keep all three. Guards alone would leave zero-byte files emitting blanks, and you're right that "almost never empty" isn't worth asserting.

A thing to keep in mind:

- Existing indexes won't self-heal, since re-index skips unchanged files by hash. I'll mention in the release notes that a full re-index clears previously stored blank chunks.

Push the cap-ordering fix and I'll get this merged. Nice find, thanks for this!

Dewars30:
Thanks — you were right, and the repro was exact. I reproduced it against v1.9.2 before fixing:

```
chunks: 2
  lines 1-100  len=2000  trimmed=""          <-- survives
  lines 91-101 len=228   trimmed="realCode();"
```

99 lines of 23 spaces with `realCode()` on line 100. My invariant held at the filter and was violated by the time the chunk was returned, which is a worse failure than the one I set out to fix — it looked correct in the code and in the tests.

Pushed `a5d7edc`:

**Cap first, then filter.** `applyCharCap` now truncates and *then* drops blanks, so nothing the cap creates can slip past.

**Two tests, not one.** The first fails on the old ordering (verified: `1 failed | 50 passed` with filter-first, `51 passed` after). The second asserts oversized chunks are still *truncated* rather than dropped — without it, a future reordering could trade this bug for silent content loss, which would be harder to notice.

**Renamed the two guard tests.** You're right that they don't isolate the guards they were named after — I reverted each guard independently and confirmed they stay green, because `applyCharCap` catches those chunks either way. Since `chunkByAstRegions` isn't exported, naming them for the behaviour they actually pin seemed more honest than asserting through a seam that doesn't exist. Keeping the guards themselves, for the reason you gave: not building a chunk only to discard it.

Gate: `npm run test:unit` 44 files / 994 tests, `npx tsc --noEmit` clean, `npm run lint` clean across 100 files.

Good call on the release-notes point about existing indexes not self-healing — hash-skip means they won't. Worth noting the blanks are removable without a re-embed: `chunkId` is `sha256(relativePath + ":" + startLine)`, so a filtered delete on `content == ""` clears them and every surviving chunk keeps its id. That's what we did on a 25k-chunk index here rather than rebuild.

giancarloerra:
@coderabbitai full review

PR #98, opened 2026-07-30
fix(search): rank cross-project results by cosine, not intra-collection rank

### LINKED ISSUE #94: [Bug]: cross-project search ranks by intra-collection rank only — a weak hit in a small repo outranks a strong hit in a large one
# [Bug]: cross-project search ranks by intra-collection rank only — a weak hit in a small repo outranks a strong hit in a large one

## What happens

With `includeLinked: true` across four repos of very different sizes, ranking does not track
relevance. In one real search, an irrelevant benchmark JSON dump came back **above** the document
literally headed "The four-stage pipeline" for a query about that pipeline.

The cause is not the embeddings — it is the client-side fusion. `mergeMultiCollectionResults`
scores purely by a document's **rank inside its own collection**, so a weak hit that happens to be
rank 0 of a 254-chunk repo outranks a much stronger hit that is rank 1 of a 23,206-chunk repo.

## Reproduction

Unit-level, no index required. Every list below is sorted descending, which is the shape the
function actually receives — `searchChunksWithVector` uses `query: { fusion: "rrf" }`, so Qdrant
returns each collection already sorted by score.

```js
import { mergeMultiCollectionResults } from "socraticode/dist/services/qdrant.js";

const hit = (relativePath, score) => ({ relativePath, score, content: "…", filePath: "/" + relativePath });

const fused = mergeMultiCollectionResults([
  // large collection: a strong, on-topic match — but rank 1 in its own list
  { label: "big-repo",   results: [hit("docs/OVERVIEW.md", 0.91), hit("docs/PIPELINE.md", 0.83)] },
  // small collection: one weak, off-topic match — but rank 0 in its own list
  { label: "small-repo", results: [hit("benchmarks/raw/dump.json", 0.41)] },
], 10);

for (const r of fused) console.log(`${r.score.toFixed(6)}  [${r.project}]  ${r.relativePath}`);
```

Actual output on `v1.9.2`:

```
0.016393  [big-repo]    docs/OVERVIEW.md
0.016393  [small-repo]  benchmarks/raw/dump.json     ← 0.41 ties the 0.91 hit
0.016129  [big-repo]    docs/PIPELINE.md             ← 0.83 ranks BELOW the 0.41 hit
```

The 0.41 document ties the 0.91 document and beats the 0.83 document. The incoming score is present
on every record and is never consulted.

*(The paths above are synthetic stand-ins chosen to mirror the live case; this is a unit-level
repro and does not replay the original query.)*

## Expected

Cross-project ranking should be influenced by match strength, not only by position within each
source list. At minimum, a 0.41 hit should not outrank a 0.83 hit from a larger collection.

## Mechanism

`src/services/qdrant.ts:440`:

```ts
435  /** Merge results from multiple collection queries using client-side Reciprocal Rank Fusion.
436   * Deduplicates by `label::relativePath` so that files with the same relative path
437   * in different projects are kept as separate hits. Within a single project,
438   * the first (higher-priority) occurrence wins on conflict.
439   * Exported for unit testing. */
440  export function mergeMultiCollectionResults(
…
444    const RRF_K = 60;
…
450        const key = `${label}::${r.relativePath}`;
451        const rrfContribution = 1 / (RRF_K + rank + 1);
452
453        const existing = scored.get(key);
454        if (existing) {
455          existing.rrfScore += rrfContribution;
456          // Keep the version from the higher-priority (earlier) collection
457        } else {
458          scored.set(key, { ...r, project: label, rrfScore: rrfContribution });
459        }
```

**To be clear: including `label` in the key is deliberate.** The docstring at 436-438 says so
explicitly, and keeping same-path files from different projects as separate hits is a defensible
choice — you often *want* to see both the canonical file and a vendored copy, each labelled. This
report is about an unintended consequence of that choice, not a careless mistake.

The consequence: because the key is namespaced per collection, **a document can never appear in
more than one list**, so line 455 is unreachable across collections. RRF's core mechanism —
documents corroborated by several rankers accumulating score — is therefore inert in exactly the
case fusion exists for. A single-chunk result therefore receives exactly one `1 / (60 + rank + 1)`
term, and the fused score becomes a pure function of intra-collection rank. (Multi-chunk files are
the exception — see below.)

Two consequences follow:

1. **Collection size is discarded.** Rank 0 of a 254-chunk repo and rank 0 of a 23,206-chunk repo
   both score `1/61`, though they are wildly different evidence.
2. **Score magnitude is discarded.** `r.score` is carried through untouched and never read.

Line 455 *is* still reachable **within** a single collection, when one file matches at several
ranks as multiple chunks — so the accumulation branch is not dead code. In one of our repos, 32 of
118 files have more than one chunk (max 14), and such a file legitimately accumulates
`1/61 + 1/62 = 0.0325`. That is worth noting because it means multi-chunk files are implicitly
favoured over single-chunk ones, independent of relevance.

## Filed separately

There is a second, sharper consequence of the same `1/61` ceiling: the documented `SEARCH_MIN_SCORE`
default of `0.10` is six times higher than any cross-project hit can score, so cross-project search
returns nothing at default settings. I have filed that separately, since it has an obvious fix while
this one needs design work — see the companion issue linked in the comments.

## Possible directions

Offered as options, not a prescription — you know the retrieval design far better than I do:

- **Normalise scores per collection before fusing** (e.g. min-max or z-score within each list), then
  fuse on the normalised value. Keeps RRF's robustness while letting magnitude matter.
- **Weight the RRF contribution by collection size**, so rank 0 of a large corpus outweighs rank 0
  of a tiny one.
- **Fuse on similarity rather than pure rank** for the cross-project path, since Qdrant has already
  done rank fusion server-side within each collection — the client-side pass is arguably fusing
  something that is already fused.
- Independently, consider making `SEARCH_MIN_SCORE` scale-aware, or applying it before fusion
  rather than after.

Each has tradeoffs — normalisation is noisy for very small result sets, and size weighting needs a
cheap way to know collection cardinality at query time.

## Environment

| | |
|---|---|
| socraticode | 1.9.2 |
| OS | macOS 26.6 (arm64) |
| Node | v20.20.0 |
| Embedding | `unclemusclez/jina-embeddings-v2-base-code` @ 768d via Ollama |
| Qdrant | 1.17.0 |
| MCP host | Claude Code and Codex CLI (stdio) |
| Corpus | 4 linked projects: 23,206 / 2,762 / 322 / 254 points (a 91× spread) |

Thanks for socraticode — it is genuinely good, and this only surfaced because we lean on
cross-repo search heavily.

giancarloerra:
Fixes #94.

## The bug

Cross-project search (`includeLinked: true`) ordered results by each hit's **position within its own collection**, so relevance did not survive the merge. Reported case: an irrelevant benchmark dump from a 254-chunk repo outranked the document literally headed "The four-stage pipeline" from a 23,206-chunk repo.

The same rank fusion capped every cross-project score at `1/(60+0+1) = 0.0164`, **six times below** the documented `SEARCH_MIN_SCORE` default of `0.10`. So at stock settings the threshold discarded everything and `includeLinked` returned nothing at all.

## Why the obvious fix does not work

Ranking on the score Qdrant already returns looks right but is not: the query asks for `fusion: rrf`, so those scores are rank-derived too. Measured against three real indexed projects.

Two irrelevant hits score 0.5000 against a perfect match's 0.6111. Cosine discriminates cleanly, and does so in both directions (a weather query puts astroweather top at 0.7661 while socraticode drops to 0.5855).

## The fix

Request the dense vector alongside the existing hybrid query (**by name**, so the sparse BM25 vector stays behind), compute cosine client-side, and rank the merged set on it. Retrieval still uses hybrid RRF per collection, so lexical recall is unchanged; only the ordering *across* collections changes.

This resolves all three symptoms in the issue at once:

- **Ranking** tracks relevance instead of intra-collection position.
- **`SEARCH_MIN_SCORE`** works as documented: scores land at ~0.5–0.77, so cross-project search returns results at stock settings instead of nothing.
- **Multi-chunk bias** disappears: a file's best chunk is kept rather than accumulating `1/61 + 1/62 + …`, which previously favoured files simply for having more chunks.

Verified end to end against three real projects: the correct project wins each query and **5/5 results now survive the 0.10 default** (previously 0).

## Backwards compatibility

- **`mergeMultiCollectionResults` keeps its exact signature.** It uses cosine when every hit carries one and otherwise runs the original fusion untouched, so any existing caller passing plain `SearchResult`s (including scripts importing it from `dist`) gets precisely what it got before. All 11 pre-existing tests pass unmodified.
- The requirement is **all-or-nothing**: ranking a cosine against an RRF value would compare two different quantities, so a single hit without one sends the whole query down the fusion path.
- **No re-index, no data or schema change**, existing indexes work as-is. `SearchResult.denseScore` is optional and additive, and is stripped before results are returned.
- **Single-project search is untouched.**

## Behaviour change to note in the release

Cross-project search goes from returning nothing to returning ranked results, and its score scale moves from `≤ 0.0164` to ~0.5–0.77. Anyone who worked around the bug by setting `SEARCH_MIN_SCORE` very low (say `0.001`) will find that threshold is now effectively a no-op and should reset it to the default.

## Cost

One dense vector per hit on the Qdrant response: **21.5KB → 270KB at `limit=20`**. Local HTTP, no measurable latency change (18ms vs 20ms), and none of it reaches the model, so there is no token cost.

## Type of change

- [x] Documentation update

## Testing

- `npx tsc --noEmit` clean, `npm run lint` (biome) clean, `npm run build` succeeds
- Unit **1015/1015**, integration **167/167**
- 9 new tests: ranking by strength, threshold survival, multi-chunk best-not-sum, same-path-two-projects preserved, `denseScore` not leaked, limit, plus three fallback/compat cases
- **Mutation-checked**: forcing the cosine path off fails 4 tests; dropping the all-or-nothing guard fails the mixed-input test; a keep-first dedupe fails the best-chunk test
- Local CodeRabbit review: no findings (four earlier findings fixed, including a real one where mismatched vector dimensions were silently truncated instead of falling back)

## Summary by CodeRabbit

* **New Features**
  * Improved cross-project and multi-collection search ranking using cosine-similarity when dense similarity data is available.
  * File paths are handled consistently across projects during merged results.
* **Bug Fixes**
  * Prevents mixed scoring scales by automatically falling back to rank fusion when cosine can’t be computed.
  * Ensures `SEARCH_MIN_SCORE` thresholding follows the correct scoring mode.
* **Documentation**
  * Updated guidance for cross-project ranking/scoring and the `SEARCH_MIN_SCORE` semantics, including cosine→fallback behavior.
* **Tests**
  * Added unit coverage for cosine ranking, dense-score extraction/fallback scenarios, and merge edge cases.


giancarloerra:
@coderabbitai full review

PR #103, opened 2026-08-05
fix(graph): name Ruby calls from the grammar's method field

giancarloerra:
## Summary

Ruby is the second language whose call sites were named by the JavaScript extractor, and it loses even more than PHP did (#101): `extractCalleeNameJs` locates the callee as the identifier before a `(`, and idiomatic Ruby overwhelmingly omits parentheses — Rails DSL declarations (`has_many :posts`), callbacks (`run_callbacks :save`), plain receiver calls (`logger.info msg`). Every one of those resolved to `null` and was silently dropped, so Ruby files contributed almost no call edges and `codebase_impact` under-reported callers throughout Ruby codebases.

Measured on a Rails-style class body containing `has_many`, `validates`, `run_callbacks`, `logger.info` and one parenthesised `persist()`: **before, only `persist` was extracted; after, all five.**

## The fix

Take the callee from the `call` node's **`method` field** instead of parsing text: tree-sitter-ruby names it for every call shape — parenthesised or not, command style, safe navigation (`a&.b`), block calls, and each link of a fluent chain as its own node. The JS text parse remains only as a fallback when the field is absent. Structural beats textual here; there is no bracket-matching to get wrong.

## Known limitation (grammar-level, documented in the code)

A bare receiverless, argumentless call (`helper` alone on a line) parses as a plain `identifier`, indistinguishable from a variable read, so it is not a `call` node and still produces no edge. Recording it would fabricate edges from every variable mention, which is worse than the missing edge.

## Backwards compatibility

- Every other language produces **byte-identical** extraction output (diffed against `main` with fixtures for ts/py/go/php/java): only the Ruby path changes.
- More Ruby names now reach `resolveCallSites`; names with no in-project definition (DSL methods like `has_many`) resolve to `unresolved` exactly as unknown names always have.

## Testing

- 9 new tests in `tests/unit/graph-symbols-ruby-calls.test.ts`: paren-less receiver calls, command-style DSL, parenthesised calls unchanged, fluent chains (with counts — each link its own call), block calls, safe navigation, scope attribution, the bare-identifier non-case, and a realistic Rails body asserting count and set
- **Mutation-checked**: reverting to the text parse fails 7 of the 9
- `tsc` clean, biome clean, full unit suite passes
- Adversarially verified in an isolated worktree (heredocs, string interpolation, `save!`/`valid?` names, setter sends, endless methods): no wrong-name cases found

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Ruby call detection so more kinds of method calls are recognized, including command-style calls, safe navigation, block calls, fluent chains, and parenthesis-less usage.
  * Preserved existing behavior when detailed call information is unavailable, reducing missed results in Ruby code analysis.

* **Tests**
  * Added broader coverage for Ruby call extraction, including common Rails-style and DSL patterns.

### DIFF
--- a/src/services/graph-symbols.ts
+++ b/src/services/graph-symbols.ts
@@ -1220,7 +1220,17 @@ function extractFromRuby(
 
   const rawCalls: ExtractedSymbols["rawCalls"] = [];
   for (const node of safeFindAll(root, "call")) {
-    const calleeName = extractCalleeNameJs(node.text());
+    // Ask the grammar, not the text. tree-sitter-ruby's `call` node carries the
+    // callee in its `method` field for every call shape — parenthesised or not,
+    // command style (`has_many :posts`), safe navigation (`a&.b`), block calls,
+    // and each link of a fluent chain as its own node. The previous
+    // extractCalleeNameJs(text) parse required a `(` before the name, so every
+    // parenthesis-less call — the dominant Ruby idiom — was silently dropped
+    // and Ruby files contributed almost no call edges. (A bare receiverless,
+    // argumentless `helper` parses as a plain identifier, indistinguishable
+    // from a variable read, so it is not a `call` node and stays out.)
+    const methodNode = node.field("method");
+    const calleeName = methodNode ? methodNode.text() : extractCalleeNameJs(node.text());
     if (!calleeName) continue;
     const r = node.range();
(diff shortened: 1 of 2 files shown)


giancarloerra:
@coderabbitai full review

Issue #111, opened 2026-08-19
DEVELOPER.md: graph-resolution.ts section documents a 5-parameter resolveImport and none of the build*Map helpers


gregoryfoster:
## Summary

The `graph-resolution.ts` section of `DEVELOPER.md` has drifted from the source. On current `main` it lists one function with a five-parameter signature:

> \| `resolveImport` \| `(specifier, importerPath, projectRoot, fileSet, language) → string \| null` \| Resolve module specifier to file path \|

The actual signature takes eleven:

```ts
export function resolveImport(
  moduleSpecifier: string,
  sourceFile: string,
  projectPath: string,
  fileSet: Set<string>,
  language: string,
  aliases?: PathAliases,
  jvmSuffixMap?: Map<string, string>,
  csNamespaceMap?: Map<string, string[]>,
  goModuleInfo?: GoModuleInfo[] | null,
  phpPsr4Map?: Map<string, string[]>,
  dartPackageMap?: Map<string, string>,
): string | null
```

## The larger gap

The six missing parameters are not incidental — each is the mechanism behind a language-specific resolution fix, and the functions that build them are absent from `DEVELOPER.md` entirely:

| Helper | Introduced for | In DEVELOPER.md |
|---|---|---|
| `buildJvmSuffixMap` | JVM multi-module layouts | no |
| `buildCsNamespaceMap` | C# `using` resolution | no |
| `buildGoModuleInfo` | Go modules, incl. nested `go.mod` (#82, #45) | no |
| `buildPhpPsr4Map` | PHP PSR-4 (#43) | no |
| `buildDartPackageMap` | Dart `package:` URIs (#106) | no |

They share a design worth recording in one place: each is called once per graph build, each discovers its manifests by walking the filesystem with the graphable walk's ignore filter rather than scanning `fileSet` (manifests are never in it — the trap behind #82), and each is threaded into `resolveImport` as a trailing optional parameter so existing callers stay source-compatible. A contributor adding the next language currently has to reconstruct that pattern by reading the source; it is exactly what `DEVELOPER.md` is for.

## Why file this separately

Noticed while preparing a fix for #107, which adds a twelfth parameter in the same shape. Correcting the table there would have meant an unrelated documentation change inside a bug-fix PR, so it is split out rather than folded in.

## Suggested fix

[...]

Issue #46, opened 2026-05-04
fix: Python codebase_graph_query returns 0 edges for sibling-flat imports — sourceDir not tried


mrsuit92:
## Summary

Python projects that use flat imports between sibling files (`import config` where `config.py` lives in the same directory as the importer) produce 0 dependency edges. This is a common pattern in service-style monorepos where each top-level directory is a runnable Python application root.

## Root Cause

`src/services/graph-resolution.ts:168-192`:

For non-relative imports, resolution tries only:

- `<projectPath>/<module>.py`
- `<projectPath>/src/<module>.py`
- `<projectPath>/lib/<module>.py`

It does **not** try `<sourceDir>/<module>.py` — the directory of the importing file. Sibling-flat imports therefore never resolve, even though they are valid at runtime when the file is executed from its own directory.

## Reproduction

Layout:

```
project/
├── service-a/
│   ├── main.py        # contains: `import config`
│   ├── config.py
│   └── ...
└── service-b/
    └── ...
```

`main.py` runs via `python main.py` from inside `service-a/`, so `import config` resolves to `service-a/config.py` at runtime. SocratiCode does not check this path and the edge is lost.

Real-world impact: 49 Python files in a service-style project, hundreds of intra-service imports, **0 dependency edges** reported.

## Proposed Solution

After the existing project-root and `src`/`lib` checks fail, fall back to:

- `<sourceDir>/<module>.py`
- `<sourceDir>/<module>/__init__.py`

This preserves current behavior for project-root layouts while covering the multi-service pattern.

Alternatively (similar to `buildJvmSuffixMap` from PR #22): build a `<basename>.py → [files]` index at graph-build time and disambiguate by proximity to the importer.

## Workaround

Rewrite imports as relative (`from . import config`). Relative imports already resolve via the existing `sourceDir`-based code path.

## Environment

- SocratiCode 1.8.4
- macOS, Docker, Ollama (`nomic-embed-text`)
- Python 3.11+, service-style monorepo, flat sibling imports throughout

PR #112, opened 2026-08-19
fix(graph): resolve Python imports through pyproject-declared import roots

### LINKED ISSUE #107: Dependency graph resolves ~0 edges on Python src layout (uv workspace) while reporting READY
## Summary

On a `uv` workspace using the standard Python **src layout**, the dependency graph builds
successfully and reports `Status: READY`, but resolves almost no edges. First-party imports
between workspace packages are not resolved, so nearly every file is an orphan and
`codebase_graph_query` / `codebase_impact` answer with "No dependency information found for
this file."

This is the same *shape* as the closed #43 (PHP namespace imports → all files orphans), but
for Python, and it survives in 1.6.1.

## Environment

- socraticode **1.6.1** (via `npx -y socraticode`, Claude Code plugin `socraticode@socraticode`)
- Node 22, Docker + Qdrant + Ollama `nomic-embed-text`, all green
- Repo: `uv` workspace, 9 packages, 362 Python files, layout
  `packages/<dashed-name>/src/<underscored_module>/…`

## Observed

```
codebase_health       → all green
codebase_graph_build  → completes in 5.1s
codebase_graph_status → Status: READY
                        Files (nodes): 374
                        Dependencies (edges): 3          ← expected ~1500+
                        Symbols: 3767
                        Call edges: 23237
                        Unresolved: 81.8%
codebase_graph_stats  → Average dependencies per file: 0.0
                        350+ orphan files
```

`codebase_graph_query` on `packages/usa-wa-adapter-sos/src/usa_wa_adapter_sos/house/build.py`
— a file with 25 import statements, 16 of them first-party — returns
`No dependency information found for this file.`

`codebase_search`, `codebase_symbol`, `codebase_symbols` and the context-artifact tools all
work correctly on the same repo. The defect is confined to the file-dependency graph.

## Which edges did resolve (the clue)

Only 3, and both patterns avoid the src layout entirely:

- `scripts/tests/systemd_units.py` ← 2 importers, via a **bare sibling import**
  (`from systemd_units import …`) — same directory, no package prefix, no `src/`.
- `packages/clearinghouse-core/src/clearinghouse_core/db/ulid.py` — 2 connections.

Everything of the form:

```python
from usa_wa_adapter_legislature.tenure_spans import Observation
```
→ `packages/usa-wa-adapter-legislature/src/usa_wa_adapter_legislature/tenure_spans.py`

failed. The resolver appears not to handle the three-way mismatch between the **dashed
distribution directory**, the intervening **`src/`**, and the **underscored module name**.

That is the standard layout produced by `uv init --lib`, hatchling and setuptools
`src-layout`, so it affects most modern Python monorepos — precisely the repos where a
dependency graph is worth having.

## Suggested direction

When resolving a first-party import `a_b_c.d`, also consider `*/src/a_b_c/d.py` (and
`*/src/a_b_c/d/__init__.py`), independent of the containing directory's name. The
distribution directory name is not derivable from the module name and should not be required
to match it. `pyproject.toml`'s `[tool.uv.workspace] members` / `[project] name` +
`[tool.hatch.build.targets.wheel] packages` give the mapping explicitly if a heuristic is
unwelcome.

## Secondary: `READY` with ~0 edges is indistinguishable from a healthy graph

`codebase_graph_status` reports `READY` in this state, and nothing downstream can tell it
apart from a graph that resolved normally. Any consumer that gates on status — as ours did —
directs its users to tools that return empty rather than erroring, and empty reads as
"nothing depends on this" rather than "the resolver failed". A yield signal in
`codebase_graph_status` (or a `DEGRADED` status when edges/nodes falls below some floor)
would let consumers route around it rather than silently trust it.

## Downstream mitigation

`gregoryfoster/skills` (the `init-socraticode` skill) now measures edges-per-file after
indexing and, below 0.1, writes a degraded exploration policy that routes
imports/dependents/blast-radius questions to `grep` and states explicitly that empty graph
output means the tool failed. That workaround retires when this is fixed.

Happy to supply a minimal reproduction repo.

*Filed from https://github.com/gregoryfoster/skills/issues/107.*

gregoryfoster:
## Summary

Fixes both halves of #107, together in one PR as you proposed there.

**Primary — Python `src` layout resolves nothing.** A packaged Python project puts its importable modules under its manifest's directory in one of two layouts: `src/` (what `uv init --lib`, hatchling and setuptools generate) or flat, beside the manifest. The resolver probed `src/` and `lib/` only at the *project* root, which reaches a single-package repo and nothing else. In a workspace, `from a_b_c.d import X` into `packages/a-b-c/src/a_b_c/d.py` resolved nothing — dashed distribution directory, underscored module, `src/` between them — and, as you confirmed, even a package's own absolute self-imports failed.

**Secondary — `READY` with ~0 edges is indistinguishable from a healthy graph.** `codebase_graph_status` now states the share of captured imports that resolved when it collapses, so a consumer can read the yield instead of inferring it.

## Changes

- `buildPythonImportRoots` registers every import root the tree's `pyproject.toml` manifests imply, root and nested. Discovery walks the filesystem with the graphable walk's ignore filter (`pyproject.toml` is never in `fileSet` — the #82 trap), and skips `site-packages/` and `dist-packages/` unconditionally, since every installed distribution ships a manifest that would register a root over vendored code shadowing the project's own modules.
- `resolveImport` gains a trailing optional `pythonImportRoots` parameter, keeping every existing caller source-compatible; omitting it reproduces the old behavior exactly.
- `saveGraphData` records `importCount` alongside the existing `edgeCount`; `codebase_graph_status` compares the two and adds an advisory below a 2% ratio. Reading it from the metadata point keeps status free of loading and walking the whole graph.

### One deviation from your fix direction, and why

You suggested deriving importable module names from the directories present under each root, rather than from `[project] name`. I registered the **roots** and try each in turn instead — same rejection of `[project] name`, for the reason you gave (Pillow imports as `PIL`), but it also covers **single-module distributions**, where `src/mymodule.py` is the entire importable surface and no directory bears the module's name. Enumerating directories would miss those. Trying roots also covers PEP 420 namespace packages for free, since it never asks what a directory contains — and it means no manifest is ever read, so no TOML parser enters the dependency tree.

### A deliberate behavior change worth your attention

The new roots are tried **after** the project-root and `src/`+`lib/` probes, so anything they resolved is untouched — but **before** the sibling-flat fallback from #46, on the grounds that a root a build manifest declares is evidence where the sibling guess is a heuristic about how a script happens to be run. In a repo that has both a manifest and the #46 sibling shape, an import both could satisfy now resolves through the declared root instead of the sibling. That is intended, and pinned by `prefers a manifest root over the sibling-flat guess`. The sibling fallback still fires wherever no declared root matches. Happy to flip the order if you'd rather #46 keep precedence.

## Type of change

- [x] Documentation update

## Testing

Verified against the **real `buildCodeGraph`**, not just unit fixtures.

On a real 618-file uv workspace whose 5 manifests yield 10 import roots:

| | edges | orphan `.py` files | captured imports resolved |
|---|---|---|---|
| `main` | 35 | 595 of 616 | 35 / 2959 = **1.2%** |
| this branch | **2103** | 67 | 2103 / 2959 = **71%** |

The captured-import count is identical on both refs, confirming the change alters resolution only and never capture.

On a synthetic uv-workspace fixture, 1 edge on `main` → 6 here, with each class landing: cross-package `src/` layout, absolute self-import, PEP 420 namespace package, single-module distribution, and flat layout — while the #46 sibling edge still resolves to the same file.

The 2% threshold is calibrated on measurements rather than intuition: 1.2% for the broken repo above, 71% for the same repo fixed, 42% for a second Python repo, and 7.6% for a doc-heavy repo whose imports are largely external. A floor of 20 captured imports keeps the ratio from being reported where it carries no signal. It is an advisory line and not a `DEGRADED` token, per your guidance — a repo of scripts importing only stdlib does trip it, so the text states the measurement first and names the benign explanation alongside the resolver one.

- [x] Unit tests pass (`npm run test:unit`) — **1165 passed**, 68 new (1097 on `v1.12.0`)
- [x] Integration tests pass (`npm run test:integration`) — **168 of 169**, see note below
- [x] New tests added for new/changed functionality — every new guard mutation-tested; each mutant fails at least one test, including both ordering decisions and each half of the `site-packages`/`dist-packages` skip

**Note on the one integration failure:** `qdrant.test.ts › lists collections including the test collection` fails when `QDRANT_COLLECTION_PREFIX` is set, because the test asserts on the literal unprefixed `codebase_test_qdrant_integration` while `listCodebaseCollections()` filters by prefix. It fails identically on `main` — unrelated to this PR, and mentioned only so the run is reproducible. I ran with a prefix to isolate the suite from other indexed projects on the same Qdrant.

## Checklist

## Related issues

Closes #107.

Related: #111 — while working on this I found the `graph-resolution.ts` section of `DEVELOPER.md` documents a five-parameter `resolveImport` (the real one takes eleven on `main`) and none of the five `build*Map` helpers. This PR adds a twelfth parameter in the same shape. I filed that separately rather than folding an unrelated documentation change into a bug fix — glad to take it if you'd like a PR.

Structurally this follows #108 (Dart `package:` URIs through pubspec-derived roots), which you named as the same disease class.

## Summary by CodeRabbit

* **New Features**
  * Improved Python import resolution across workspace, `src`, flat, namespace-package, and single-module layouts.
  * Added import-resolution diagnostics to codebase graph status, including resolved and captured import counts.
  * Graph metadata now preserves import counts for diagnostic reporting.

* **Bug Fixes**
  * Improved handling of nested manifests, exclusions, comments, malformed declarations, and unsupported workspace patterns.
  * Added clearer warnings when unresolved imports may limit dependency and impact results.

* **Documentation**
  * Expanded guidance for Python import layouts and low-resolution graphs.


giancarloerra:
CI is green and CodeRabbit is clean on 5fd0305. Two defects block the merge. Both sit in the new root matching, not in the advisory half, and both appear on layouts more common than the ones the added tests cover.

## 1. Registered roots have no locality

`buildPythonImportRoots` returns one flat sorted list, and the new loop tries it in that order for every source file, first hit wins, with no preference for the root that contains the importing file. Fixture: a per-service monorepo, two flat `uv init --app` services, each with its own `config.py`, `handlers.py`, and `main.py` doing `import config` / `from handlers import handle`.

```
main      services/beta-svc/main.py -> services/beta-svc/config.py
          services/beta-svc/main.py -> services/beta-svc/handlers.py
5fd0305   services/beta-svc/main.py -> services/alpha-svc/config.py
          services/beta-svc/main.py -> services/alpha-svc/handlers.py
```

Roots on the branch are `[".", "services/alpha-svc", "services/alpha-svc/src", "services/beta-svc", "services/beta-svc/src", "src"]`, and beta's own root sorts after alpha's. Edge count is 4 on both refs, so `importCount` and the advisory cannot surface this: the graph reads healthy while pointing at another package's file. Those edges also feed `resolveCallSites`, so `codebase_impact` inherits the inversion.

Direction: order candidates by proximity to the importing file, deepest root that is a prefix of its directory first.

## 2. Any manifest anywhere contributes roots

Only `site-packages/` and `dist-packages/` are excluded, so a manifest in a subtree that is on no `sys.path` still registers one. Fixture: root package plus `examples/demo/pyproject.toml`.

```
main      src/mainpkg/app.py     import settings stays unresolved
5fd0305   src/mainpkg/app.py ->  examples/demo/settings.py
```

A null becomes a fabricated edge, and proximity ordering does not address it since no root competes. The same shape reproduces for a checked-in `third_party/` sdist and for an environment directory not named `site-packages` (`.venv312/...`) holding an editable checkout. Sample apps, cookiecutter templates, docs projects and per-fixture manifests are ordinary contents of a Python repo. This needs a rule for which manifests count, for example only manifests on the ancestor path from the project root to the importing file, or only declared workspace members.

## README contradicts the code

`README.md:973` ends "Relative imports and the project-root `src/`/`lib/` and sibling-flat conventions are unchanged." The comment above the new loop states the opposite: "That IS a behavior change for a repo that has both a manifest and the #46 sibling shape: such an import can now resolve to a different file than it did before." The monorepo fixture above shows the comment is the accurate one.

## Test gap

None of the 28 added cases builds two manifests carrying the same top-level module name, and none builds a manifest in a non-importable subtree. Both shapes belong in the suite alongside the fixes.

## Decisions

- **Precedence versus #46: flip it.** CPython 3.14.4 puts the script's own directory at `sys.path[0]`, ahead of every installed-distribution entry, so where both candidates are importable Python resolves the sibling. This is separate from defect 1 and neither fix substitutes for the other.
- **Threshold: keep 2% with the floor of 20.** Boundary behaviour matches the documented contract at every edge case, and this repo measures 51.5%, well clear of it.
- **Advisory scope, no change requested now:** the ratio sums every language, so a polyglot repo whose Python subtree resolves nothing can still read around 42% and stay silent, which is the #107 symptom itself. Worth a follow-up issue rather than expanding this PR.
- **Merge will be a squash.** The first commit body carries bare `#82` and `#46` references; conventional-changelog renders bare `#N` as closes annotations, so the release notes would claim to close both, while `Refs #107.` yields no reference at all because of the trailing period. The repo also does not carry AI co-authorship trailers in commit messages. A squash with a corrected message settles both.

## What holds

Test counts are exact (1097 on `main`, 1125 here, the stated +28), tsc, lint and build are clean, and the mutation-testing claim checks out on spot-reverted mutants. Captured-import counts are identical across refs, so the change is resolution-only as stated. Non-Python resolution is unchanged across the full language matrix. Persisted-graph compatibility is clean in both directions: a graph written by 1.12.0 without `importCount` prints no advisory rather than a wrong one, and older code ignores the new key. `buildPythonImportRoots` costs about 60 ms on a 5280-file tree, once per build, behind the `hasPython` gate. The `QDRANT_COLLECTION_PREFIX` integration failure is pre-existing and identical on `main`, and only fires when that variable is set.

Can you push the two fixes?

gregoryfoster:
Both defects reproduced here before anything was changed, on your fixtures, and both are fixed in `1f22f00`. Pushed as a commit on top rather than a rewrite so this diffs against what you reviewed.

## 1. Locality

`buildPythonManifests` now records each manifest's roots alongside the workspace members it declares, and a new `pythonRootsForFile` selects and orders the roots for one file. Roots containing the file come first, deepest first; the rest follow lexicographically, so the cross-package tie — where nothing in the import names a package — breaks identically on every machine.

Scoped roots are computed once per directory inside the build, so `resolveImport`'s parameter is still `string[]` and the ~100 positional call sites are untouched.

## 2. Manifest scope

A manifest applies to a file only when it sits on that file's ancestor path, or when an ancestor manifest declares it a `[tool.uv.workspace]` member.

Ancestry alone would have dropped the cross-package imports this PR exists to resolve — a sibling package is never an ancestor of the importing file — so member declarations are the exception, on the grounds that a member declaration is precisely the statement that the sibling is importable from here. Your other two shapes fall out of the same rule: a checked-in `third_party/` sdist and an editable checkout in a `.venv312/` are members of nothing.

This costs the "no manifest is ever read" property I claimed. The section is located by header and read only as far as the next top-level table, so another tool's `members` key cannot be mistaken for uv's, and only `*` and `**` are interpreted in a glob — every other character, `?` included, matches literally, which is stricter than uv and can only drop a member rather than invent one. Happy to swap it for a different rule if you would rather not read manifests at all; the tradeoff is that non-uv monorepos then lose cross-package edges.

## Precedence versus #46: flipped

Sibling-flat now runs first. `sys.path[0]` settles it, and the argument applies whether or not the package is installed, which my "declared roots are evidence" reasoning did not account for.

That also settles the README contradiction you found: the sibling-flat convention genuinely is unchanged now, so the sentence is true as written. I extended it to describe the scoping and ordering rules as well.

## Test gap

Eleven further cases, including the two shapes you named — two manifests carrying the same top-level module name, and a manifest in a non-importable subtree.

Worth flagging one thing I found while mutation-testing the fix: my first proximity test had the alphabetically *earlier* package doing the importing, so lexicographic order produced the right answer by accident and the test could not distinguish proximity from luck. Reverting the containment rule left the whole suite green. The test now imports from the later package, and that mutant dies against two cases. It is the same blind spot that let the flat list through in the first place.

## Verification

Your two fixtures now match `main` exactly — four correct edges in the monorepo, none in the subtree case. The headline result is unchanged: 2,103 edges on the real 618-file uv workspace, and all six edge classes on the synthetic fixture. 1136 unit tests, `tsc`, lint and build clean.

## On your decisions

Threshold and advisory scope: agreed, nothing changed. I have left the advisory's polyglot blind spot for you to file if you want it — it is your call whether that is worth an issue.

Squash: understood, and the reference and trailer defects are real. I have left the earlier commits alone so this push stays reviewable against your read of them; say the word if you would rather I rewrite the branch into a single clean commit instead, and I will force-push it in whatever shape you want to merge.

gregoryfoster:
Fixed in `36a8d39`, along with two more instances of the same defect that looking for it turned up.

**The reported one.** TOML permits a comment after a table header; the end-of-line anchor rejected that form. Reproduced as a failing test first, and writing that test surfaced the second instance immediately.

**A quoted word inside a comment in the array** was read as a member glob. On a manifest whose `members` array carried `# not "examples/demo"`, `examples/demo` was admitted as a workspace member — turning the scoping rule that fixes defect 2 back into the fabricated edge it exists to prevent. Comments are now stripped before the array is read.

**A `]` inside any member string** ended the array span mid-string. The truncated span has no closing quote, so *every* member in the array was lost, not just the offending entry — a glob character class such as `packages/[ab]*` is enough to trigger it, taking a well-formed `packages/*` beside it down too. The span now walks to the first `]` outside a string.

All three fail the same way: the manifest reads as declaring nothing, scopes to its own subtree, and cross-package imports return to `null` with nothing on screen to say why. Same silent-degradation shape as the original issue.

I tried bracket counting for the third and dropped it. It survives a balanced character class but not `"packages/a]b"`, and once strings are opaque it never fires — a guard no test could justify keeping. Mutation testing is what exposed that: with both mechanisms present, each masked the other and neither mutant died, which also meant my test wasn't discriminating. The test now uses an unbalanced bracket and the remaining guard fails without it.

The docstring now states what the reader does *not* understand — `#` inside a member string, escaped quotes, multi-line strings, glob syntax beyond `*` and `**`. Each is legal TOML that degrades silently, so the boundary seemed worth naming rather than leaving to be rediscovered. All of them drop members rather than invent them.

Also in this commit, from a review pass on my own side: the two array-key patterns are hoisted to module constants (neither carries `/g`, so no `lastIndex` state), and the `PythonManifest` docstring is about the interface now — it had been the old function's docstring moved wholesale, still opening by describing a sorted list of directories.

1140 unit tests, `tsc` and lint clean. Both your fixtures and the 618-file workspace result are unchanged: four correct edges in the monorepo, none in the subtree case, 2,103 edges on the real repo.

Standing offer from my last comment still applies — say the word and I will force-push this branch as a single clean commit in whatever shape you want to merge, rather than leaving you to squash four.

giancarloerra:
Both blockers are gone at `36a8d39`, on fresh fixtures rather than the two I posted: locality holds when the alphabetically-first service is not the importer and when two teams' subtrees carry the same module names, and manifest scoping holds while the cross-package imports this PR exists to resolve still resolve. The precedence flip matches `main` and matches CPython. 1140 tests, tsc and lint clean.

One new defect, introduced by `36a8d39` itself, plus two older ones in the same reader. All three are the same shape: the reader can admit something the manifest does not declare.

## Blocking: `exclude` runs through a drop-only reader

`listOf(WORKSPACE_EXCLUDE_KEY)` shares the members path. For `members`, dropping an entry is safe. For `exclude`, dropping an entry removes an exclusion, so the package the manifest explicitly excludes becomes a member and draws a cross-package edge. Root manifest with `members = ["packages/*"]` over `packages/{alpha,beta,legacy}`, resolving `legacy.core` from `packages/alpha/src/alpha/main.py`, checked against uv 0.10.0 on the same bytes:

```
exclude entry                            uv 0.10.0        36a8d39
["packages/legacy"]                      legacy excluded  no edge          correct
["packages/leg[a]cy"]                    legacy excluded  edge to legacy   wrong
["packages/#legacy", "packages/legacy"]  legacy excluded  edge to legacy   wrong
```

uv honours `[a]` as a glob class, so the first wrong row is a plain miss. The second is the regression: the comment stripping added in `36a8d39` truncates the array at the `#`, so the following `"packages/legacy"` entry never exists. One case worth naming because it looks like a defect and is not: `exclude = ["packages/#legacy"]` alone. uv reads it as a literal path that matches nothing and keeps `legacy` a member, so the edge is correct there.

## Two more cases where the reader invents rather than drops

- A `[tool.uv.workspace]` header inside a triple-quoted string is read as a real section. The header regex has no string context, so a manifest whose only workspace text sits in `description = """..."""` yields members and widens roots, while `tomllib` reports no such table. Present before `36a8d39`.
- A mixed quote inside a member truncates it to a prefix that can name a real directory: `members = ["it's/*"]` yields member `it`, a directory the manifest never names. uv selects only the root for the same manifest.

Both falsify the docstring sentence that "can only ever drop a member rather than invent one", which is the stated justification for not parsing TOML.

## What would settle it

Keep the hand-rolled reader and make its invariant true: void the whole `[tool.uv.workspace]` section whenever the scanner meets something it cannot represent faithfully, whether in `members`, in `exclude`, or before the header. A `#` or a mixed quote inside a string, a glob character beyond `*` and `**`, a `"""` or `'''` span. Voiding degrades to ancestor-path scoping, which is the conservative behavior, and it makes "drop, never invent" hold for both arrays instead of only one. Adding a TOML parser dependency is the alternative and is a bigger call than this PR needs.

Please add the `#`, character-class and `?` shapes to the `exclude` tests. The existing case at `tests/unit/graph-resolution.test.ts:703` covers only the plain literal, which is why the regression passed.

## Two constraints for the merge

Neither is currently violated, and both should stay true after the exclude fix.

- **Nothing existing may change behavior.** No import that resolved before this PR may resolve to a different file after it, in any language. The precedence flip restored that for the #46 shape, and non-Python resolution is byte-identical across the language matrix with a populated `pythonImportRoots` passed on every call.
- **No user action may be required.** A graph written by 1.12.0 has no `importCount`, and on this branch it prints no advisory rather than a wrong one; 1.12.0 reading a graph written here ignores the extra key. Worth confirming in the PR text that the only user-visible sequencing is the usual one: the new Python edges appear when the graph is next rebuilt, and nothing needs re-indexing for correctness.

## One wording fix

`README.md:973` says "so uv workspaces and monorepos get cross-package edges". Only `[tool.uv.workspace]` is read, so poetry, pdm and hatch path-dependency monorepos get no cross-package edges. Narrow the clause to uv workspaces. Treating a path or editable dependency as a member declaration would cover the others and belongs in its own change.

gregoryfoster:
Fixed in `07ca25b`. All three reproduced first, and your table now matches uv on every row.

## The asymmetry is the fix

You are right that `exclude` cannot share a drop-only reader, and the reason generalises past the three instances: dropping a member costs an edge that should have resolved, while dropping an *exclusion* admits a package the manifest explicitly excludes and draws an edge uv would not. Patching each shape was never going to hold, so the reader now has the invariant instead — **narrow what a manifest declares, never widen it** — and voids the whole section the moment it meets anything it cannot represent exactly, in either array.

| exclude entry | uv 0.10.0 | `07ca25b` | |
|---|---|---|---|
| `["packages/legacy"]` | legacy excluded | no edge | correct |
| `["packages/leg[a]cy"]` | legacy excluded | no edge (section voided) | correct |
| `["packages/#legacy", "packages/legacy"]` | legacy excluded | no edge | correct |
| `["packages/#legacy"]` | legacy a member | edge to legacy | correct |

Two of these resolve more precisely than voiding would, because they are represented rather than approximated. A `#` inside a string is read as the literal path character it is, so the fourth row keeps its edge — voiding there would have dropped a real one. And a `"""` block sitting beside `members` in the same table is blanked rather than fatal, so that manifest still declares its members.

The header is now matched only in text with multi-line string spans masked out, so prose in a `description = """…"""` no longer poses as a table. Strings are opaque to the scanner, so `members = ["it's/*"]` is the single member `it's/*` rather than the invented `it`.

Voiding degrades to ancestor-path scoping, which resolves strictly fewer imports, so the fallback direction is the conservative one.

## Tests

The `exclude` side now covers the `#`, character-class, `?` and non-scalar shapes alongside the plain literal, plus the `#`-alone row that looks like a defect and is not. Two earlier cases asserted the weaker "keep the well-formed neighbours" behaviour and now assert voiding, since the invariant they encoded was the wrong one.

Two guards came out rather than in, both unable to change an observable outcome: bracket-depth counting, which never fires once strings are opaque, and a newline check inside a quoted value, since no directory name can contain one so such an entry matches nothing either way. Every remaining guard is mutation-tested. Two survived a first pass because they masked each other — reading the section from masked text and masking at all — and the tests that separate them were written for exactly that.

## Your two constraints

**Nothing existing changes behaviour.** Both fixtures from your earlier review, the six synthetic edge classes and the 618-file workspace are unchanged at 2,103 edges. Non-Python resolution is untouched.

**No user action is required**, and the only user-visible sequencing is the ordinary one: the new Python edges appear when the graph is next rebuilt, by `codebase_graph_build` or the watcher, and nothing needs re-indexing for correctness. A graph written by 1.12.0 carries no `importCount` and prints no advisory rather than a wrong one; 1.12.0 reading a graph written here ignores the extra key. Neither direction requires a rebuild to stay correct — a rebuild only adds the edges that were missing.

## Wording

`README.md:973` no longer claims monorepos generally get cross-package edges. Only `[tool.uv.workspace]` is read, so poetry, pdm and hatch path-dependency layouts get ancestor-path scoping and no cross-package edges; the clause now says so. Agreed that treating a path or editable dependency as a member declaration is its own change.

1150 unit tests, `tsc`, lint and build clean.

gregoryfoster:
@giancarloerra, it seems like we're getting to a point of diminishing returns on building a "good enough" inline TOML parser, LMK if you want me to continue hacking at these edges vs. adding an ideally lightweight dependency.

giancarloerra:
> @giancarloerra, it seems like we're getting to a point of diminishing returns on building a "good enough" inline TOML parser, LMK if you want me to continue hacking at these edges vs. adding an ideally lightweight dependency.

Ok, I think let's add the dependency. Your read on diminishing returns is right, and the decision comes down to what the remaining cases are.

To me they seem not edges. They are the lexer. `# """` masking a valid table and `members = ["a" "b"]` inventing members both come from the scanner not tracking comment and string state, which is the first thing a TOML parser does and the last thing a pattern-matching reader can bolt on. Each further fix in that direction adds code whose only purpose is to approximate exactly what a parser already does, and the invariant you settled on makes that cost visible: a hand-rolled scanner can only keep "narrow, never widen" by voiding more often, and every void is real edges lost in a manifest that was perfectly valid. That trade gets worse with each round, not better.

Swapping also loses nothing you gained in `07ca25b`. The precision you argued for is preserved: a `#` inside a string stays the literal path character it is, so `exclude = ["packages/#legacy"]` alone still keeps `legacy` a member and still draws that edge.

## Which one

`smol-toml`, checked against the registry and run here rather than taken from its README:

- zero dependencies, BSD-3-Clause, `engines: >= 18` matching the repo floor, 109 KB unpacked, ships its own types, published this month;
- parsed identically on Node 18.20.8, 20.20.2 and 22.23.2 in containers;
- every shape from both review rounds resolves correctly, including two legal spellings that are invisible to the current reader:

```
# """ then a valid table                       members ["packages/alpha"]   (voided today)
members = ["a" "b"]                            throws, invalid TOML         (invents members today)
exclude = ["packages/#legacy", "packages/legacy"]   both entries kept
members = ["it's/*"]                           kept verbatim
[tool.uv.workspace] inside a """ block         not a table
[ tool.uv.workspace ]                          members ["packages/*"]       (missed today)
[tool.uv] + workspace = { members = [...] }    members ["packages/*"]       (missed today)
```

The last two matter beyond tidiness: both are ordinary TOML that a user can write today and get silently no cross-package edges, with nothing on screen to say why. That is the same failure shape as the original issue.

## Two conditions on the swap

- A malformed or unreadable manifest must be caught and treated as declaring nothing, never thrown out of the graph build. A manifest that fails to read is skipped today and the build continues; that has to stay true when the parser starts throwing on invalid input.
- The invariant stays as you wrote it, narrow what a manifest declares and never widen it, so the failure direction after the swap is still fewer edges rather than wrong ones.

Keep this round's tests pointed at behavior rather than at scanner internals so they survive the swap and prove it.

## On `07ca25b` itself

Confirmed here before any of the above: all four `exclude` shapes match uv 0.10.0 on identical manifests, including the row that looks like a defect and is not, and both earlier fixtures are unchanged with four correct edges and no fabricated one.

```
exclude entry                            uv 0.10.0        07ca25b
["packages/legacy"]                      legacy excluded  no edge
["packages/leg[a]cy"]                    legacy excluded  no edge
["packages/#legacy", "packages/legacy"]  legacy excluded  no edge
["packages/#legacy"]                     legacy kept      edge to legacy
```

So the branch is correct as it stands. The swap is about what happens to the next valid manifest nobody has written yet, not about fixing something broken now.

Once it lands I will approve and squash.

gregoryfoster:
Done in `e239742`. Agreed on the diagnosis — the remaining cases were the lexer, and the invariant made the cost of approximating it visible rather than hiding it.

## The swap

`smol-toml` 1.8.0, checked here rather than taken from the README: BSD-3-Clause, no transitive dependencies (the lockfile gains exactly one entry), `engines: >= 18`, 109 KB unpacked, its own types, ESM and CJS. I ran your three container versions and got identical parses on 18.20.8, 20.20.2 and 22.23.2. Recorded in `THIRD-PARTY-LICENSES`.

Everything you listed reproduces. I also checked the uv column myself against **uv 0.11.8**, locking throwaway workspaces and reading `[manifest] members` out of the lock rather than reasoning about globset:

| manifest | uv 0.11.8 | `e239742` |
|---|---|---|
| `[ tool.uv.workspace ]` | members resolved | resolved (missed before) |
| `[tool.uv]` + `workspace = { members = [...] }` | members resolved | resolved (missed before) |
| `# """` above a valid table | members resolved | resolved (voided before) |
| `members = ["a" "b"]` | fails to parse | voids (invented two before) |
| `members = "packages/*"` | fails the lock | voids |
| `members = ["packages/[al]*"]` | matched **both** alpha and legacy | voids |
| `members = ["packages/alph?"]` | matched alpha only | voids |
| `exclude = ["packages/#legacy", "packages/legacy"]` | legacy excluded | no edge |
| `exclude = ["packages/#legacy"]` | legacy kept | edge to legacy |

The character class and the `?` confirm the glob check has to stay: uv really does compute a different set than a literal reading, so those still void under the invariant.

## Two things beyond your list

**A third silently-missed spelling.** A top-level `tool.uv.workspace.members` dotted key is a workspace to uv, and the old reader found nothing in it. The table walk covers it for free — and the same key written *after* `[project]` belongs to `[project]`, in uv and in the parser alike, so that one correctly stays not-a-workspace.

**A BOM would have been a regression, not a gap.** uv skips a leading byte-order mark and locks the workspace normally. TOML's grammar has no place for one, so `tomllib` and `smol-toml` both reject the whole document — whereas the old header regex matched a later line and never noticed. Swapping without stripping it would have taken every member off any manifest saved from an editor that writes a BOM. It is stripped before parsing, so uv's behaviour is what gets modelled.

## Your two conditions

**A malformed manifest declares nothing and never leaves the build.** Caught at the parse, same contract an unreadable file already had, now pinned by a test that puts a broken manifest at the root above a valid member so the walk has to carry on past it.

**The invariant is unchanged.** Voiding still degrades to ancestor-path scoping, which resolves strictly fewer imports.

## What came out

The scanner, the multi-line-string masking and the three hoisted patterns are gone. The walk then lost a redundant own-key check — a missing key yields undefined, which the next step already rejects, so no test could kill it — and removing it made the two guards it had been masking killable. Net guards went down again, 8 of 8 mutation-tested.

One of those turned out to matter more than it looked: the check on the value the walk ends on is load-bearing for `[tool.uv]` with no workspace under it, the commonest uv manifest there is. Without it that manifest crashes the whole graph build. No test covered it, which is exactly what the mutation pass was for.

**On the tests you asked to survive the swap:** I ran all 185 against the parser before editing any of them, and all 185 passed. The 13 added since cover the new spellings, the BOM, the non-array and non-string shapes, and the malformed manifest; one earlier test was rewritten because the behaviour it asserted was a scanner limitation rather than a rule.

1162 unit tests, integration 168/169 (a pre-existing collection-prefix artifact, unrelated), `tsc`, lint and build clean. Both your earlier fixtures, the six synthetic edge classes and the 618-file workspace are unchanged at 2,103 edges.

giancarloerra:
The swap is clean. Both constraints hold, and the three spellings, the BOM strip and the unparseable-manifest path all behave as your commit message says: a broken manifest declares nothing and the build finishes, `[tool.uv]` with nothing under it does not crash, and every earlier fixture is unchanged, including all four `exclude` rows against uv. One lockfile entry, no transitive dependencies, and the tests were behavior-level enough to survive the swap untouched.

Two things to fix before this goes in. Neither comes from `e239742`.

## 1. `exclude` globs under-exclude, and that fabricates an edge

`toRe` translates `*` to `[^/]*`, so it cannot cross a path separator, while uv's `*` can. On the members side that only ever narrows, which is safe. On the exclude side it admits a package the manifest excluded. Same fixture as before, `members = ["packages/*"]` throughout, alpha importing `legacy.core`, checked against uv 0.10.0 on identical manifests:

```
exclude entry              uv 0.10.0                     this branch
["packages/*legacy"]       legacy excluded               no edge
["*legacy"]                legacy excluded               edge to legacy
["*"]                      everything excluded, root only  edge to legacy
```

`exclude = ["*"]` is not an exotic spelling. This lands on the invariant the file states at the top of the section, that the reader may narrow what a manifest declares and never widen it, so the comment currently asserts the opposite of what the code does. It arrived in `1f22f00` rather than in the parser swap, and it is not on `main`, so it ships with this PR either way.

Two fixes fit the existing design: extend the unsupported-glob check so an exclude pattern whose `*` would have to span a `/` voids the section, or translate `*` as `.*` for exclude patterns only. Include-side `*` needs no change, since `[^/]*` is a strict subset of what uv matches there.

## 2. `README.md:1493` now contradicts the file it points at

The sentence lists the dependency licenses as `(MIT, Apache 2.0, ISC)`, and this PR adds `smol-toml   BSD 3-Clause License` to `THIRD-PARTY-LICENSES`. Adding BSD 3-Clause to the parenthetical is the whole fix. Nothing else about licensing changes: BSD 3-Clause is permissive and asks only for the attribution line this PR already adds.

## Smaller

- `src/services/graph-resolution.ts:752` contains two literal `\u2014` escape sequences inside a comment, six ASCII characters each, where em dashes were presumably meant. Nothing interprets an escape in a comment, so it renders verbatim. It is the only occurrence in `src/`.
- The description says 198 tests. The branch reports 197: 185 at `07ca25b` plus 13 added minus 1 renamed away, with no coverage lost.
- Not blocking, worth knowing: `buildPythonManifests` applies no size cap to a manifest read, unlike `MAX_GRAPH_FILE_BYTES` for graphable files, and smol-toml's parse time grows superlinearly in key count where the scanner was linear. Reaching that regime takes on the order of a hundred thousand keys in one `pyproject.toml`. At realistic scale the two are indistinguishable, and a 50 MB manifest parsed in 177 ms.

Push the two fixes and this is ready.

gregoryfoster:
Both fixed in `0a63b84`, plus the two smaller ones.

## 1. The exclude side

Reproduced, and you were right that it lands on the invariant rather than beside it: the comment at the top of that section forbids exactly what the code did.

Digging into why, uv does not use one matcher for the two lists. It expands `members` by globbing the filesystem, where a lone `*` selects a single path segment, and matches `exclude` as a pattern against the member's whole path, where it does not stop at a separator. That is why `members = ["packages/*"]` leaves `packages/alpha/inner` out while `exclude = ["*"]` empties the workspace — two code paths, two meanings for the same character.

So `*` now translates per list rather than per reader. I took the second of your two options: translating exclude exactly is better than voiding it, since voiding would have been correct and lossier.

Both versions, since I did not want the model read off one release — I pulled 0.10.0 alongside 0.11.8 and they agree on every row. uv counts the workspace root among its own members; this reader scopes the declaring manifest through ancestry instead, which is why it is absent on the right:

| exclude entry | uv 0.10.0 / 0.11.8 | `0a63b84` | |
|---|---|---|---|
| `["packages/legacy"]` | alpha | alpha | |
| `["packages/leg*"]` | alpha | alpha | |
| `["packages/*legacy"]` | alpha | alpha | |
| `["*legacy"]` | alpha | alpha | **was alpha + legacy** |
| `["*"]` | root only | none | **was alpha + legacy** |
| `["packages/*"]` | root only | none | |
| `["packages/#legacy", "packages/legacy"]` | alpha | alpha | |
| `["packages/#legacy"]` | alpha, legacy | alpha, legacy | |
| `["packages/leg[a]cy"]` | alpha | voids | narrows |
| `["packages/legac?"]` | alpha | voids | narrows |

The include side deliberately keeps the narrow `*`, and that is now pinned rather than assumed: widening it registers roots over packages the workspace never declared, so a test asserts `packages/alpha/inner` stays out. Both directions are mutation-tested — narrowing exclude and widening members each fail a test written for that one case, 10 of 10 mutants killed.

The translation's parameter names both spellings in its type, which rejects a fragment that is neither. It cannot stop the two being transposed, since each is valid at either call site, and the comment says so rather than implying the type carries more than it does.

## 2. README

Fixed — `BSD 3-Clause` added to the parenthetical at `README.md:1493`. Agreed that nothing else about licensing moves.

## Smaller

- The two `—` sequences are em dashes now. None left anywhere in `src/` or `tests/`.
- Test count: you are right, and the arithmetic was mine — 185 plus 13 added reads as 198, but one of the 13 replaced an existing test rather than adding to it, so 197, now 200 with this round's three. The PR description had a separate stale figure from the first commit; I measured the baseline by checking out `v1.12.0` (1097 unit tests, 140 in that file) and corrected it to **1165 passed, 68 new**.
- The manifest size cap: leaving it, per not-blocking. `MAX_GRAPH_FILE_BYTES` governs graphable files rather than manifests, and a hundred thousand keys in one `pyproject.toml` is far outside anything this resolver will meet — happy to be told otherwise if you would rather it were capped here.

1165 unit tests, integration 168/169 (the standing collection-prefix artifact), `tsc`, lint and build clean. Both earlier fixtures, the six synthetic edge classes and the 618-file workspace are unchanged at 2,103 edges.

PR #113, opened 2026-08-20
feat(index): add Elixir AST support

### LINKED ISSUE #110: [Feature]: Add Elixir support
### Problem or motivation

I have Elixir code I’d like to index properly. Right now I can add `.ex` and `.exs` through `EXTRA_EXTENSIONS`, but that only treats them as plaintext. I’d like the normal language-aware stuff too: better chunks, graph edges, and symbols.

### Proposed solution

Elixir looks like the easy next language to add since ast-grep already has `@ast-grep/lang-elixir`.

I think this would mean:

- add `.ex` and `.exs` as Elixir files;
- register the Elixir grammar with the existing dynamic grammar loader;
- use it for AST-aware chunking;
- pick up basic Elixir module relationships (`alias`, `import`, `require`, `use`);
- resolve in-project modules from `defmodule`;
- add basic `defmodule` / `def` / `defp` symbols and calls where practical.

HEEx/EEx would need more work because they need template-aware parsing and Phoenix component resolution, so they’d probably be out of scope for this first pass.

### Alternatives considered

`EXTRA_EXTENSIONS=.ex,.exs` works for search already, but the files stay plaintext and don’t get the graph or symbol features.

### Area

Indexing, Code graph, Testing, Documentation

### Additional context

ast-grep already lists Elixir support and has a published Node grammar package:

- https://astgrep.com/reference/languages
- https://www.npmjs.com/package/@ast-grep/lang-elixir

### Checklist

- [x] I have searched existing issues and this hasn't been requested before

AJMC2002:
## Summary

Adds language-aware Elixir source and template support: AST-aware chunks, graph edges, symbols, and HEEx/EEx/LEEx template parsing.

## Changes

- Add `.ex` / `.exs` and `@ast-grep/lang-elixir`
- Register Elixir with the dynamic grammar loader for AST-aware chunking
- Extract `alias`, `import`, `require`, and `use`; resolve in-project `defmodule`s
- Add basic `defmodule` / `def` / `defp` symbols and calls
- Parse HEEx/EEx/LEEx templates with dedicated grammars for AST chunking, remote components, and embedded Elixir calls

## Type of change

- [x] Documentation update

## Testing

- [x] Unit tests pass (`npm run test:unit`)
- [x] Integration tests pass (`npm run test:integration`) — if applicable

## Checklist

## Related issues

Closes #110

## Summary by CodeRabbit

* **New Features**
  * Added indexing and relationship extraction for Elixir `.ex` and `.exs` files.
  * Added support for HEEx, EEx, and LEEx templates, including embedded expressions, imports, components, and AST-aware chunking.
  * Added Elixir module resolution and improved Python workspace import resolution.
  * Added Elixir visualization colors in supported diagrams.
  * Added default exclusions for `_build` and `deps`.
* **Documentation**
  * Updated supported-language documentation with Elixir and template support.
* **Tests**
  * Added coverage for Elixir source and template parsing, indexing, chunking, and dependency resolution.


src/services/graph-resolution.ts
```diff
@@ -275,6 +275,23 @@ export interface GoModuleInfo {
...
+  const map = new Map<string, string>();
+  for (const file of [...fileSet].filter((f) => [".ex", ".exs"].includes(path.extname(f))).sort()) {
+    let source: string;
+    try {
+      source = readFileSync(path.join(projectPath, file), "utf8");
+    } catch {
+      continue;
+    }
+    for (const match of source.matchAll(/^\s*defmodule\s+([A-Z]\w*(?:\.[A-Z]\w*)*)\s+do\b/gm)) {
+      if (!map.has(match[1])) map.set(match[1], file);
```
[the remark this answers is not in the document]
AJMC2002: Fixed in 2a86bb2: normalize module-map extensions and support block plus keyword-form `defmodule` declarations. Added uppercase `.EX`/`.EXS` keyword-form fixtures; `npm test` and build pass.

giancarloerra:
Thanks for picking this up. The structure follows how the other languages are wired here, and the parts that touch shared code are safe: `.ex`/`.exs` are gated behind `hasElixir`, so a project with no Elixir pays nothing, and the `isTopLevel` change is inert for every other language, since none of the twenty installed grammars has root kind `source`:

```
program: javascript typescript tsx php bash dart java ruby
source_file: go rust kotlin swift        module: python
translation_unit: c cpp                  compilation_unit: scala csharp
chunk: lua   stylesheet: css   document: html
```

What blocks the merge is not missing support, it is wrong data. Extraction currently reports things that are not in the source, and those entries reach `codebase_symbol` and `codebase_impact` as if they were real. Output from this branch on a small fixture:

```
extractImports(consumer.ex)  ->  ["MyApp.Notes", "MyApp.Notes"]
   alias MyApp.Notes                captured
   alias(MyApp.Notes)               dropped
   alias MyApp.{Notes, Sub.Deep}    only Notes, the dotted member dropped

extractSymbolsAndCalls(notes.ex).calls  ->  ["moduledoc", "limit"]
   @moduledoc """..."""   reported as a call
   @limit 50              reported as a call, and resolves to any function named limit

buildElixirModuleMap  ->  MyApp.Ghost -> lib/notes.ex
   MyApp.Ghost appears only inside a @moduledoc heredoc example
```

Alongside those, reproduced on the same branch: `note.limit` (a map or struct read) becomes a cross-file call edge to a function named `limit`; `def fetch(id) when is_integer(id)` reports itself as its own caller and callee; `for i <- 1..10 do` emits a call that resolves to a real function named `for`; and a nested `defmodule MyApp.Outer do defmodule Inner do` registers `Inner` rather than `MyApp.Outer.Inner`, so `alias MyApp.Outer.Inner` finds nothing while a bare `Inner` key enters the project-wide namespace.

## Root cause, and the shape of the fix

Two mechanisms produce all of it.

1. **The directive and call detection keys on the first `identifier` child.** In tree-sitter-elixir almost everything is a `call` node, because `defmodule`, `def`, `alias` and friends are macros rather than syntax, and the node carries a `target` field naming what is being called. Reading `target` instead of scanning children separates a real invocation from a definition head, a module attribute (`@x` parses as `unary_operator > call`), and a dotted field read (require the dot's first child to be an `alias`, that is a module, before treating it as a remote call). Special forms and definition macros then need excluding by name: `for`, `with`, `case`, `cond`, `receive`, `try`, `quote`, `unquote`, and `defstruct`, `defguard`, `defmacro`, `defdelegate`, `defprotocol`, `defimpl`.

2. **`buildElixirModuleMap` regex-scans raw file text.** A `defmodule` inside a `@moduledoc` heredoc or a `~S` sigil is indistinguishable from a real one, and documented examples are idiomatic Elixir. The map should come from the AST `defmodule` nodes, which the symbol walk already visits, carrying the enclosing prefix so nested modules get their full name. `buildCsNamespaceMap` uses raw text, but C# has no doctest convention.

Smaller, same file: `extractImports` anchors its regex at the start of the `arguments` text, which begins with `(` for `alias(MyApp.Foo)`, so strip a leading paren and whitespace; and the multi-alias member filter `^[A-Z]\w*$` rejects dotted members, so `alias MyAppWeb.{Endpoint, Router.Helpers}` loses the second entry.

## Two constraints for the merge

- **Nothing existing may change behavior.** The `isTopLevel` line is the only shared code touched, and the grammar scan above shows it cannot affect another language. Please keep it that way if the fixes touch anything outside the Elixir paths.
- **No user action may be required.** Adding a language is additive, and a project with no Elixir is untouched. Worth stating in the PR text that anyone currently indexing `.ex` through `EXTRA_EXTENSIONS` keeps working as-is and gains AST chunking, symbols and edges when the project is next indexed.

## Open questions

- `buildElixirModuleMap` keeps the first path per module name; `buildCsNamespaceMap` keeps every candidate (`Map<string, string[]>`). With the first-wins shape, a vendored or legacy copy can shadow the canonical file and still report a unique match. Match the C# shape?
- `_build` and `deps` are not in `DEFAULT_IGNORE_PATTERNS`. Without gitignore processing (off by default) a stale build artifact or hex dependency source can win the module map. Add both entries here.
- `ENTRY_POINT_NAMES.elixir` is `{"main"}`. `main/1` is escript only; the conventions are `Application.start/2` and `Mix.Task.run/1`.
- Elixir is missing from the graph colour maps in `graph-analysis.ts` and `viewer-app.js`, so its nodes render grey and drop out of the Mermaid legend.
- `defprotocol` and `defimpl` produce no module symbol, so every `def` inside them becomes a top-level symbol with a duplicate unqualified name. Scoping that out for now is reasonable if the README says so.
- Chunking uses `elixir: ["call"]`, which at top level matches only the outer `defmodule`, so a module is one region and falls back to line windows. That matches how class-based languages already behave, but the README currently implies function-level chunking. The Dart paragraph above it is the model for stating grammar-level gaps plainly.

Happy to review the next round.

src/services/code-graph.ts
```diff
@@ -609,6 +610,7 @@ const EXTENSION_TO_AST_GREP_LANG: Record<string, Lang | string> = {
   ".php": "php",
   ".swift": "swift",
   ".dart": "dart",
+  ".ex": "elixir", ".exs": "elixir",
```
EevanW, line 618: What about
.eex (Embedded Elixir templates)
and
.leex (LiveView‑templates)?

```suggestion
  ".ex": "elixir", ".exs": "elixir",  ".eex": "elixir", ".leex": "elixir",
```

giancarloerra: Thanks for looking at this. Those two extensions should stay out, and the reason is that the grammar cannot read them. An EEx template parsed with `@ast-grep/lang-elixir`, the grammar this PR registers:

```
template.eex   rootKind=ERROR   ERROR_nodes=14   call_nodes=0
real.ex        rootKind=source  ERROR_nodes=0    call_nodes=3
```

EEx and HEEx are templates, HTML carrying `<%= %>` or `{ }` tags, not Elixir source files. Mapping them to `elixir` would claim them as parseable, produce zero symbols and zero call edges, and replace the line-based chunking they get today with one file-sized error node, so search over templates would get worse rather than better.

Two further points on `.leex` specifically: it was soft-deprecated in LiveView 0.16 and stateful LiveComponents dropped support for it, and the current line is 1.2.10, so `.heex` is the extension a template pass would need to target first.

Template support is real and worth having, and it needs its own path: parse as HTML, extract the embedded Elixir regions, and resolve Phoenix function components. That is the same shape the Svelte and Vue handling already uses here. @AJMC2002 scoped it out of this PR deliberately in #110 and that was the right call. 

Sources: https://hexdocs.pm/phoenix_live_view/0.17.11/changelog.html

AJMC2002: That can create false data. In the Eex snippet:

```eex
   <button onclick="save()">Save</button>
 ```

the Elixir parser recovers it as a call node targeting `button`, with this node text:
 
 ```elixir
 button onclick="save()">Save</button
 ```
 
A simpler safe approach is to add them as first-class grammarless template extensions: index them with the Elixir language label, use line-based chunking, and include them as graph leaf nodes without symbol or call extraction.

Full AST support could be added using dedicated HEEx/EEx grammars.

**Comment by AJMC2002** [comment]
I tried the suggested direct `.heex` / `.eex` / `.leex` → Elixir grammar mapping first, but tree-sitter's error recovery makes that unsafe for graph data. For example:

```eex
<button onclick="save()">Save</button>
```

is recovered by the Elixir grammar as a `call` node targeting `button`, while a real expression such as:

```eex
<%= render "form.html", assigns %>
```

can be swallowed inside an `ERROR` node and produce no `render` call.

I therefore added the published HEEx/EEx tree-sitter WASM grammars instead. Template markup is parsed by its template grammar; only grammar-confirmed embedded expression ranges are copied into a whitespace mask (preserving original line positions) and then parsed as Elixir. EEx control-flow fragments spanning multiple directives are combined. If either the template AST or the combined Elixir AST is invalid, extraction fails closed for that data rather than guessing. HEEx remote component modules also contribute file dependencies, and both formats now provide template-AST chunk boundaries.

`.leex` uses the EEx grammar on a best-effort basis because it is largely EEx-compatible. If parsing fails, it safely falls back to line chunking and emits no guessed symbol/call edges.

The parsers load lazily only when one of these template extensions is present, so non-template projects do not initialize the WASM runtime. Checks now cover markup text, HTML attributes, comments, escaped EEx, Unicode offsets, multi-directive blocks, malformed expressions, HEEx component resolution, and LEEx compatibility. Current verified behavior:

```text
<button onclick="save()">  -> no calls
<%# cleanup() %>            -> no calls
<%%= fake() %>              -> no calls
<%= render(...) %>          -> render
{load(@id)}                 -> load
valid LEEx                   -> live_patch, item_path
invalid embedded expression -> no guessed calls
```

All 1,105 unit tests pass, along with the build and lint checks.

giancarloerra:
Thanks for all the fixes!

The Python workspace PR just merged (was before yours), and this branch now conflicts with `main`. The collision is narrow: both PRs appended a trailing optional parameter to `resolveImport` (that PR added `pythonImportRoots`, this one `elixirModuleMap`) and extended the same call site in `code-graph.ts`.

Resolution on rebase: keep both, with `elixirModuleMap` after `pythonImportRoots` as the last parameter, and pass both at the `code-graph.ts` call site. `THIRD-PARTY-LICENSES` and `package-lock.json` may need the same both-sides treatment since both PRs added entries.

Once the rebase is up, the full review pass on the template work happens and this can move and I'll merge!

giancarloerra:
Full review of the rebased branch, everything below verified by execution on the built head.

**All eight items from the earlier review are fixed.** Attributes and field reads no longer emit calls, heredoc `defmodule` examples register nothing, `alias(...)` and dotted multi-alias members extract, nested modules carry their full prefix, special forms are excluded, guarded heads no longer self-call, and the module map is AST-derived and multi-candidate. Your seven-row behavior table reproduces row for row. The rebase also preserves the Python workspace behavior exactly: a uv-workspace fixture produces byte-identical graphs on `main` and this branch, and the other languages are unchanged across the resolution matrix.

**The dependency audit came back clean**: SLSA provenance on both `@lumis-sh` packages validates via npm audit signatures, zero transitive dependencies, zero install scripts, the WASM blobs import only memory primitives with no I/O capability and match their pinned hashes, and the lazy-load claim holds with a verified fail-safe when the packages are absent.

Three things before merge:

**1. One fail-closed gap.** `analyzeElixirTemplate` rejects masked Elixir only on ERROR nodes. An unclosed call in HEEx parses as a valid template, and tree-sitter recovers the expression with a zero-width MISSING token and no ERROR node, so the call is emitted:

```
<div>{load(@id)}</div>   ->  load        correct
<div>{load(((}</div>     ->  no calls    fail-closed works
<div>{load(@id}</div>    ->  load        MISSING token, slips through
```

Reproduced on Node 18, 20 and 22. The emitted name does match text present in the file, so it is not fabrication, but it contradicts the stated contract that invalid embedded expressions produce no guessed calls. Rejecting trees that contain MISSING tokens alongside ERROR nodes closes it; please add the unclosed-paren case as a regression test.

**2. The PR body is stale.** The Summary still says HEEx/EEx stays out of this first pass, while the branch ships the template service, three dependencies, and the template test suite. Update it to describe what is actually here.

**3. The CLA checkbox in the body is unchecked.** Please confirm it.

Two things recorded as documented limitations rather than blockers, since neither produces wrong data: fully-qualified calls without an `alias`/`import`/`require`/`use` directive resolve as unresolved (the README already scopes resolution to the four directives; worth an explicit sentence that qualified bare calls are not yet resolved), and `defprotocol`/`defimpl`/`defguard`/`defmacro`/`defdelegate` register no symbols (call emission for them is correctly suppressed). Add both to the README limitations note; follow-ups welcome as separate issues.

AJMC2002:
Thanks for letting me contribute! :)

PR #117, opened 2026-08-25
fix(context): exclude binary and ignored files from directory artifacts

### LINKED ISSUE #116: Directory context artifacts embed compiled bytecode: the binary guard in readArtifactContent() can never fire, and the walk honours no ignore file
Two defects in `readArtifactContent()` combine so that a **directory** context artifact silently embeds compiled bytecode and other build output. Read out of the published package, `socraticode@1.12.0`, `dist/services/context-artifacts.js` (lines 125–145).

Both are one-line-ish fixes; I'm happy to send a PR for either or both.

## 1. The directory walk has a hardcoded ignore list and imports no ignore service

```js
if (stat.isDirectory()) {
    // Find all files in the directory (recursively, skip hidden/dot-files)
    const files = await glob("**/*", {
        cwd: resolved,
        nodir: true,
        dot: false,
        ignore: ["**/node_modules/**", "**/.git/**"],
    });
```

`node_modules` and `.git`, and nothing else. Grepping the module for `ignore` returns exactly these two occurrences and no `.gitignore` / `.socraticodeignore` reference anywhere — so there is **no code path by which either ignore file can apply to a context artifact**. This is not a load-order or staleness effect; I confirmed it behaviourally first (adding `__pycache__/` and `*.pyc` to `.socraticodeignore` and re-indexing left the artifact's chunk count identical) and then in the source.

That is defensible as a deliberate choice, but it is not what the README's index-scope framing leads a user to expect, and it is the general case that lets the next consumer embed `dist/`, `.pytest_cache/` or coverage data.

## 2. The binary guard is already written, and it can never fire

The loop immediately below is what turns (1) into embedded mojibake rather than merely extra text files:

```js
for (const file of files) {
    const filePath = path.join(resolved, file);
    try {
        const content = await fsp.readFile(filePath, "utf-8");
        parts.push(`# ── ${file} ──\n${content}`);
    }
    catch {
        // skip unreadable files (binary, permissions, etc.)
        logger.debug(`Artifact: skipping unreadable file ${file}`);
    }
}
```

The comment states the intent exactly — *skip binary*. It never happens, because **`fsp.readFile(path, "utf-8")` does not throw on binary input**; it returns a string with U+FFFD replacement characters. Verified against a `.pyc` fixture on Node:

```
readFile utf-8 on binary: RETURNED (no throw), len 8, hasFFFD true
```

So every `.pyc` takes the *success* branch and is embedded, mojibake and all. Nothing is unreadable, so nothing is skipped, so nothing is logged — the `logger.debug` line is dead code.

## Measured consequence

A project declaring `{"name": "alembic-migrations", "path": "./alembic/versions"}` — a natural, documented-shape artifact. Every test run and every `alembic` invocation drops `__pycache__/` into that directory:

- **70 `.pyc` files embedded**, 352K of bytecode against 320K of source
- **32 of the artifact's 86 chunks** were compiled bytecode
- a `codebase_context_search` for *"what is the current alembic migration head revision"* returned **decompiled bytecode as its top hit**

| | chunks |
|---|---|
| baseline | 86 |
| after adding `__pycache__/` + `*.pyc` to `.socraticodeignore`, re-indexed | **86** |
| after `rm -rf alembic/versions/__pycache__`, re-indexed | **54** |

Top-hit score for the same query went 0.5417 → 0.6111 once the bytecode was gone.

## Why this is worth fixing rather than documenting

It fails silently and **upward**. Nothing errors, the chunk count *rises*, and `codebase_status` reports more artifact content indexed than before. Every available signal says the artifact got healthier while search quality got worse — so there is no observation a user can make that would prompt them to look. The only reason I found it was chasing a bad search result back to its chunk.

## Suggested fixes

1. **Make the binary guard actually detect binary.** This alone fixes the bytecode case and preserves the existing intent. Read as a `Buffer` and either sniff for a NUL byte in the first few KB, or decode with a fatal decoder:

   ```js
   const buf = await fsp.readFile(filePath);
   if (buf.includes(0)) { logger.debug(`Artifact: skipping binary file ${file}`); continue; }
   const content = buf.toString("utf-8");
   ```

   (`new TextDecoder("utf-8", { fatal: true }).decode(buf)` inside the existing `try` would also work and would make the `catch` live, at the cost of rejecting a few legitimately-latin1 text files.)

2. **Wire in the ignore chain** the code indexer already uses, so a directory artifact honours the built-in defaults + `.gitignore` + `.socraticodeignore`. This is the general fix and the one that stops the next unanticipated build directory.

A cheap partial alternative to (2), if wiring the ignore service into artifacts is unwanted for scope reasons: extend the hardcoded list with the common build-output names (`**/__pycache__/**`, `**/dist/**`, `**/build/**`, `**/*.pyc`). Less principled, but it covers the case above.

Worth noting for (1) alone: the artifact walk already passes `dot: false`, so dot-directories like `.pytest_cache/` are excluded — `__pycache__` is the notable build directory that is *not* hidden and therefore *is* walked.

gregoryfoster:
## Summary

Fixes both defects from #116 together, as agreed in your review: a **directory** context artifact no longer embeds compiled bytecode and other build output.

The walk honoured no ignore file beyond a hardcoded `node_modules`/`.git` list, and its binary guard could never fire — `readFile(path, "utf-8")` does not throw on binary input, it returns U+FFFD replacement characters, so every `.pyc` took the *success* branch and the `logger.debug` skip line was dead code. It failed upward: chunk counts and `codebase_status`'s artifact count both rose as search quality fell.

Per your instruction, the correction to the original writeup: Qdrant fusion scores are rank-derived, so the 0.5417 → 0.6111 movement is **not** a relevance delta and I make no claim from it. The defensible statement is the one the chunk counts already support — bytecode hits left the top results and source filled them (86 → 54 chunks on the reference case, 32 of which were compiled bytecode).

Both fixes live entirely inside `readArtifactContent`, so the indexed content and the staleness hash cannot diverge. Re-indexing is automatic on the next content-hash mismatch.

## Changes

- **Binary guard — the NUL sniff, not the fatal decoder.** Each file is read once as a `Buffer`, sniffed for a NUL byte in the first `DETECT_HEAD_BYTES` (8 KiB), then decoded with `toString("utf-8")`.
- **Ignore chain — post-glob filter, rooted at the artifact directory.** `createIgnoreFilter`/`shouldIgnore` over the glob results, mirroring how the indexer consumes the chain. Post-glob because glob 11's `ignore` option wants path-scurry objects, not the `ignore` package; rooted at the artifact directory because the `ignore` package throws `RangeError` on absolute or `../`-prefixed paths.
- **Exclusions are counted and returned, not logged in place.** `readArtifactContent` returns `{ ignored, binary, unreadable }`; `indexArtifact` logs the summary at info. Per-file skips stay at debug. This matters because `readArtifactContent` also serves the staleness check that runs on *every* search — logging the summary there would repeat it per search instead of reporting it once per index, and a shrinking chunk count deserves a visible cause at the moment it shrinks.
- **The "no readable files" throw now names the counts**, so the loud failure is actionable.
- **README**: the "read recursively" line gains the exclusion semantics and the dotfile rule; the Ignore Rules section, which implied the three layers already covered artifacts, now says how they do.

### Boundaries kept, as you specified

- The single-file branch is untouched — a declared single-file artifact pointing at a binary still works, with no silent skip.
- A directory of only binaries keeps the existing loud "no readable files" throw.
- `dot: false` stays, and is now documented as the reason `.pytest_cache/` was already excluded while `__pycache__/` was walked.

## Two consequences worth your explicit sign-off

**1. The Stage-0 parity argument is narrower than it first appears.** The *rule* is shared with the indexer's Stage-0 guard; the *scope* is wider. Stage-0 runs only on extensionless files, while the code index reads anything with an indexable extension through `fsp.readFile(absolutePath, "utf-8")` with no binary guard at all (`indexer.ts:808`, `:1120`; `chunkFileContent` adds none). So one class does diverge: a NUL-bearing file with an *indexable* extension — a `.sql` or `.yaml` holding embedded binary — is indexed as code but skipped in an artifact directory. I think that is right for artifacts, where a directory is swept rather than declared file by file, and the skip is counted and logged. But it is not the clean "no loss class the code index does not already have" that the sniff was chosen on, so I would rather you hear it from me than find it.

**2. A directory artifact now inherits the defaults in full, including names it might legitimately use.** Beyond `__pycache__`/`*.pyc`/`dist`/`build`, `DEFAULT_IGNORE_PATTERNS` also covers `env`, `vendor`, `target`, `out`, `coverage`, `*.map`, `*.log`. A k8s-manifest artifact with an `env/` subdirectory loses content it used to embed. That is a new silent-exclusion class in a PR whose purpose is removing one, so: the info-level summary makes it observable, and the README documents the escape hatch. Worth noting the escape hatch has one correct spelling — `!env` re-includes files under `env/`, while `!env/**` alone does not, since gitignore semantics cannot re-include a file whose parent directory is excluded. There is a test pinning that so the doc cannot rot.

## The question you deferred to review: root at the artifact directory, or at the project?

I implemented **artifact-directory rooting**. Three things surfaced while doing it, one of which I did not anticipate:

1. **Relative-path safety is structural, not incidental.** glob returns paths relative to `cwd: resolved`, so every path handed to `ignore` is already relative and inside. Project rooting would require `path.relative(project, artifactDir)` plus a guard for the `..` and absolute cases — including the global-config fallback, where the artifact directory is outside any project.
2. **A directory would be able to ignore itself.** Under project rooting, an artifact declared at `./build/openapi/` has every file matched by the default `build` pattern, and the artifact collapses to the "no readable files" throw. Artifact-directory rooting yields relative paths that no longer start with `build/`, so this cannot happen.
3. **Cost, on the hot path.** `createIgnoreFilter` walks recursively for nested `.gitignore` files. Rooted at the artifact directory that walk is bounded by the artifact; rooted at the project it is a full source-tree walk, per artifact, on every staleness check — i.e. on every `codebase_context_search`.

The cost: a project-root `.socraticodeignore` still does not reach artifacts — the exact remedy I tried in the issue before finding the cause. The defaults already contain `__pycache__` and `*.pyc`, so the reported case works with no new patterns, and the README now states the limitation rather than leaving it to inference. Happy to switch to project rooting if you weigh (3) differently.

## Ignore-filter construction on the staleness path

Resolved as you directed: **no memo**. The chain is rebuilt per read.

The numbers behind that, measured on a 3,600-file / 300-directory artifact: building the chain is ~10 ms of a ~243 ms read, about 4%, because reading and hashing file contents dominates and no memo touches it. A memo that stayed correct would have to fingerprint every nested `.gitignore`, which measured 8.0 ms against the 9.7 ms rebuild it avoids — so correctness would have cost roughly what it saved.

What did cost real time was reading each artifact **twice** on the stale path: `ensureArtifactsIndexed` reads to compute the staleness hash, then `indexArtifact` read again. It now threads that content through via an optional `preread` parameter, so a re-index walks and reads the directory once instead of twice — a full `readArtifactContent` saved per re-indexed artifact, 243 ms on the shape above against the memo's ~10 ms ceiling. It also gives the staleness decision and the write one content snapshot: what is indexed is exactly what the decision hashed, while the previous two-read path could make the decision on one version and index a later version. On `main`, `indexArtifact` still derived the stored hash and indexed content from the same second read, so there was no stored-hash/content divergence.

`createIgnoreFilter`'s two status lines drop from info to debug for the per-search noise.

## On the CHANGELOG

Not touched, deliberately. `.release-it.json` generates `CHANGELOG.md` from commit subjects via `@release-it/conventional-changelog`, so a hand-added entry would be misplaced at the next release. Correcting my own justification, as you noted: the history is not purely release commits — `c8d1e15 docs(changelog): correct scraped issue references in the 1.12.0 entry` is a hand edit, and I had that output in front of me when I wrote otherwise. It is a correction to a generated entry rather than a new one, so the generated-file rationale holds; the claim I hung it on did not. The operational note is in the commit body and the README instead. If you want it in the release notes verbatim, this is the line:

> Directory context artifacts re-index on their next content-hash check; expect chunk counts to **drop** once excluded files stop being embedded.

## Type of change

- [x] Documentation update

## Testing

- [x] Unit tests pass (`npm run test:unit`) — 1183 tests, 53 files
- [x] Integration tests pass (`npm run test:integration`) — `tests/integration/context-artifacts.test.ts`, 17 tests
- [x] TypeScript compiles cleanly (`npx tsc --noEmit`), `npx biome check src/ tests/` clean

Every test you asked for, plus two more. The ignore-chain test uses only *text* fixtures, so the binary guard cannot be what removed them and the chain is demonstrably live on its own:

| Test | What it pins |
|---|---|
| ignore-chain exclusion | all three layers, one fixture each, all fixtures plain text |
| top-level NUL binary, no ignore pattern | the guard firing on its own |
| latin1 file **kept** | fails if a fatal decoder is ever swapped in |
| UTF-16 with BOM skipped | parity with Stage-0 |
| binaries-only directory still throws | the loud failure survives |
| single-file binary artifact unchanged | the declared-path boundary |
| staleness hash == indexed-content hash | built by comparing against a directory holding only the survivors |
| build output does not mark an artifact stale | the hash covers post-exclusion content |
| `!env` negation re-includes | the documented escape hatch |
| dangling symlink counted `unreadable` | the third counter, which appears in the throw |
| preread content is indexed, not re-read from disk | the threading actually skips the second read |
| `indexArtifact` still reads when handed nothing | the parameter is optional, not required |
| staleness-read exclusions carry into what is indexed | one read, so hash and content cannot disagree |
| nested `.gitignore` added between ensure passes | ignore rules are rebuilt and the artifact re-indexes immediately |
| ensure call-site read count | the config and stale artifact are read once each; dropping the preread argument changes 2 reads to 3 |

Checked by mutation, not just by passing. With `src/services/context-artifacts.ts` reverted to `main`, **9 of the 12** new tests fail; the other three (latin1 kept, single-file binary untouched, `!env` re-include) pin properties that also hold on `main`, so they guard against regression rather than proving the fix. A stale count of 6 stood here until you measured it — that number was taken when the block held 7 tests and I did not re-measure as it grew. Separately: with a fatal `TextDecoder` swapped in for the sniff, the latin1 test fails; with the `.socraticodeignore` fixture removed, the three-layer test fails; with `indexArtifact` made to ignore its `preread` argument, the threading and read-count tests fail; and with a module-level ignore-filter cache restored, the nested-ignore freshness test fails.

## Checklist

## Related issues

Fixes #116

## Summary by CodeRabbit

* **New Features**
  * Context artifacts now automatically exclude ignored, hidden, binary, and unreadable files when processing directories.
  * Artifact processing reports how many files were excluded for each reason.
  * Hashing reflects only included content, preventing excluded build output from triggering unnecessary updates.
  * Single-file artifacts continue to be read verbatim, including binary files.

* **Documentation**
  * Updated developer and user documentation to explain artifact filtering, exclusions, and indexing behavior.


gregoryfoster:
Addressing both CodeRabbit items — pushed as `09716ea`.

**Nitpick, `tests/unit/context-artifacts.test.ts` — directory-local `.gitignore` test: taken.** Correct catch: the chain has three layers and I was exercising two, so a regression confined to the `.gitignore` layer would have passed everything in this PR. The test is now `applies all three ignore layers, with no binary involved`, carrying one fixture per layer, each matched by a pattern the other two layers do not hold (`coverage/` + `__pycache__/` for the defaults, `notes-draft.txt` for `.gitignore`, `scratch.txt` for `.socraticodeignore`). Verified by removing the `.gitignore` fixture: the test fails. It also pins `RESPECT_GITIGNORE` on for the duration, so a host shell setting cannot silently decide whether layer 2 is being exercised at all.

**Major, `src/services/context-artifacts.ts:229-234` — counting glob-filtered files: documented rather than changed, deliberately.** The finding is factually right, and I confirmed both consequences before deciding: an artifact directory containing only `node_modules/dep/a.js` throws with no exclusion detail, and a directory mixing a real file with a vendored `node_modules` reports `{"ignored":0,"binary":0,"unreadable":0}`.

I did not route those through the counted path, because doing so means deleting the glob-level `ignore`, and that option's cost is the reason it exists. A pattern ending in `**` registers with glob as a children-pattern (`node_modules/glob/dist/esm/ignore.js:72-83`), so `childrenIgnored()` prunes the subtree instead of enumerating it and discarding it afterwards. `readArtifactContent` is also the staleness path, which runs on every `codebase_context_search` — so exact counts would be paid for by fully enumerating a vendored `node_modules` on every search. That trade seemed clearly wrong for a number that appears only in a log line.

What I did instead is stop the counters from overclaiming: `ArtifactExclusions` now documents that these count what the walk *found* and then rejected, with `node_modules`/`.git` pruned before the walk yields anything and therefore counted nowhere, and the `ignore` option carries the pruning rationale and the trade-off inline.

@giancarloerra — this is a judgement call about your hot path rather than a mechanical fix, so if you would rather have exact totals and accept the walk, say so and I will make the swap.

gregoryfoster:
Taken — pushed as `33eff04`. `process.env.RESPECT_GITIGNORE = "true"` instead of deleting the variable, plus the PR body's test table updated to match the three-layer test.

One correction for the record, since the suggestion is right but the reasoning behind it is not. Deleting the variable *did* enable the layer: [`ignore.ts:74`](https://github.com/giancarloerra/SocratiCode/blob/main/src/services/ignore.ts#L74) reads `(process.env.RESPECT_GITIGNORE ?? "true").toLowerCase() !== "false"`, so unset falls through to enabled. The test was exercising the `.gitignore` layer — provable independently of the env question by removing the `versions/.gitignore` fixture, which fails the test; it could not fail that way if layer 2 were inert.

The suggestion is still worth taking, for a different reason than the one given: setting `"true"` explicitly means the test states its own precondition instead of inheriting it from a default defined in another module, so it keeps pinning layer 2 even if that default ever changes. Verified both ways after the edit — the test fails with the layer-2 fixture removed, and passes with `RESPECT_GITIGNORE=false` forced in the shell, which is the run that shows the assignment overriding a hostile environment rather than restating a default.

giancarloerra:
On the counting question: keep the pruning, and keep the documentation fix you made. `readArtifactContent` is the staleness path, so it runs on every `codebase_context_search`; paying a full enumeration of a vendored `node_modules` on every search to make a log-line number exact is the wrong trade, and the children-pattern pruning is exactly why that option is there. Documenting that the counters report what the walk found and rejected is the right resolution.

The rest of the review, all verified by execution on `33eff04`: conformance to the issue guidance on every point. The latin1 file is kept, the NUL sniff matches Stage-0 parity including the 8 KiB window and its documented residual, the single-file branch is byte-identical to `main`, the binaries-only throw and `dot: false` survive, the staleness hash moves only when kept content changes, and the upgrade path re-indexes exactly the artifacts that previously embedded now-excluded files, with no orphaned chunks. 1177 unit and 17 integration tests pass here alongside your suite.

Two fixes before merge, both small, and the first is the direct consequence of the decision above:

1. **README, two sentences.** The Ignore Rules section still ends "Whatever is left out is reported in that artifact's log line when it is indexed", which the pruning decision makes untrue for `node_modules`, `.git` and dot-files: they are counted nowhere. Scope that sentence to ignore-chain, binary and unreadable exclusions. Separately, "a directory artifact indexed by an earlier version re-indexes on its next hash check" needs the qualifier "when the walk previously embedded now-excluded files"; a directory with nothing to exclude hashes identically and correctly does not re-index.
2. **PR body, line 82.** "with the source change reverted, 6 of the new tests fail" measures as 9 of 12 here; the other 3 pin properties that also hold on `main`. Right direction, wrong number.

Two other decisions you surfaced:

- **Rooting: artifact-directory rooting is accepted.** The RangeError constraint is real, and the README now stating that project-root ignore files do not reach a directory artifact turns it into a documented semantic rather than a surprise.
- **Ignore-filter construction on every staleness check: memoise per artifact** rather than demoting the log lines. That removes both the repeated work on the hot path and the per-search log noise, while keeping the lines meaningful during real indexing.
- **CHANGELOG: your placement is accepted.** The file is generated; the chunk-count note stays in the README and goes into the release notes by hand at release time. One correction to the justification: the history does contain a non-release hand edit, so "only release commits" is not quite true, but the generated-file rationale stands.

No need to amend the commit for the footer or the trailer. This will be squashed with a maintainer-written message carrying the Closes footer, as with your previous PR.

gregoryfoster:
All three landed in `616a52d`, plus both corrections to my own claims.

**1. README, two sentences.** The Ignore Rules sentence now scopes itself to what is actually counted: *"Files dropped by the ignore rules, by the binary check, or because they could not be read are counted in that artifact's log line when it is indexed. `node_modules`, `.git` and dot-files are pruned before the walk sees them, so they appear in no count."* The upgrade sentence gained the qualifier and its converse: it re-indexes *if the walk previously embedded files that are now excluded*, and *"an artifact with nothing to exclude hashes identically and is left alone."*

**2. PR body, the mutation number.** Corrected to 9 of 12, and the body now says why it was wrong rather than just changing the digit: the 6 was measured when the block held 7 tests and I never re-measured as it grew to 12. Re-ran it against `f836e99` — 9 fail, and the three survivors are the ones you identified (latin1 kept, single-file binary untouched, `!env` re-include).

**3. Ignore-filter memoisation.** Per artifact directory, keyed on a fingerprint of that directory's own `.gitignore` and `.socraticodeignore` (mtime + size) plus `RESPECT_GITIGNORE`. `ignore.ts` is untouched, so the info lines keep their meaning and now mark a real rebuild instead of a search. It also removes a double-build I had not noticed: on a stale artifact, `ensureArtifactsIndexed` reads for the hash and `indexArtifact` reads again, which was two recursive walks per re-index and is now one.

Five tests in a new `context-artifacts-ignore-cache.test.ts`, pinning both halves of the bargain — mutation-checked in both directions: with the memo removed, the two build-once tests fail; with the memo made unconditional, the three rebuild tests fail. The `RESPECT_GITIGNORE` term turned out to be load-bearing rather than defensive; without it, flipping the flag serves a filter built under the old value.

The residual, stated in the JSDoc and the PR body: a `.gitignore` nested *below* the artifact root is not fingerprinted, because stat-ing the subtree would cost what the memo saves. Adding one deeper in the tree takes effect on the next rebuild rather than immediately. The two files a user actually edits — the ones the README points at — are both covered, and there is a test asserting an edited `.socraticodeignore` takes effect on the next read, which is the #116 symptom guarded directly.

**On the CHANGELOG justification:** you are right, and I had the evidence in hand when I wrote it — `c8d1e15 docs(changelog): correct scraped issue references in the 1.12.0 entry` was in the `git log -- CHANGELOG.md` output I ran. It is a correction to a generated entry rather than a new one, so the generated-file rationale stands, but the claim I hung it on did not. The PR body now says so.

Counting decision noted and unchanged; rooting acceptance noted. No commit amendment for the footer or trailer, as you said.

Verification here: `npx biome check src/ tests/` clean, `npx tsc --noEmit` clean, 1182 unit tests across 53 files, 17 integration tests. One incidental fix from my own review: the new test file was leaving five temp directories per run in `TMPDIR` — it now uses one root with per-test subdirectories and removes the tree, matching the sibling artifact tests.

gregoryfoster:
CodeRabbit's invalidation finding is taken — `8771aa0`.

**Superseding my previous comment on one point:** I described the memo fingerprint there as mtime + size. That is no longer what it is, and the PR body has been corrected in both places. The fingerprint now digests each ignore file's *bytes*.

The finding is right, and it lands on the one workflow this PR exists to fix. If `.socraticodeignore` is edited to the same length inside a single timestamp tick, a stat-based fingerprint holds still, the memo serves stale rules, and those rules decide which files get hashed into the artifact's content — reproducing "I edited .socraticodeignore and nothing happened", which is exactly how #116 started. Two small reads on a path that then walks the whole artifact directory is not a cost worth trading that for.

There is now a test for precisely that case: it swaps one pattern for another of equal length, pins the timestamp across the edit, asserts that `mtimeMs` and `size` are both genuinely unchanged, and requires the next read to honour the new rules. It **fails** against the mtime+size fingerprint and passes against the digest — verified by reverting the implementation and re-running.

Worth flagging how the first version of that test went wrong, since it bears on how much the surrounding mutation checks are worth. My first attempt captured the `mtime` from `stat` and fed the resulting `Date` back through `utimes` — which does not round-trip exactly. The sub-millisecond drift moved the old fingerprint by itself, so the test passed against *both* implementations and proved nothing. It only surfaced because reverting the implementation gave an inverted result I could not explain. Pinning to a whole-second timestamp, which round-trips exactly, made it bite.

Local verification: biome and `tsc --noEmit` clean, 1183 unit tests across 53 files, 17 integration tests, no temp-directory residue.

gregoryfoster:
CodeRabbit's second finding on `8771aa0` — nested `.gitignore` files not covered by the memo's fingerprint — is valid. I reproduced it, measured the ways out, and one of my own claims needs correcting first.

## Correcting my framing

The JSDoc I wrote says a nested ignore file "takes effect on the next rebuild rather than immediately", and the PR body and my earlier comment repeat that. That describes a delay. It is a permanent miss:

```
read 1:  NESTED_MARKER present = true    hash 376fc8273303f734
         edit sub/.gitignore to exclude note.md
read 2:  NESTED_MARKER present = true    hash 376fc8273303f734   <- unchanged
         25 further staleness checks:    still stale
         write an UNRELATED root .socraticodeignore
read 27: NESTED_MARKER present = false
```

The content hash never moves, so `ensureArtifactsIndexed` never re-indexes. Release comes only from an unrelated root ignore-file change, a cache clear, or a process restart. On a long-lived server that is indefinite. It is the #116 shape — edit an ignore file, nothing happens — reached through a narrower door, and my wording made it sound like a bounded lag. Whatever you decide below, that wording gets fixed in all three places.

## Measurements

Synthetic artifact directory: 3,600 files, 300 directories, 40 nested `.gitignore` files.

| Operation | Cost |
|---|---|
| `readArtifactContent`, cold | 243 ms |
| `readArtifactContent`, warm (memo hit) | 218 ms |
| `createIgnoreFilter` — everything the memo skips | 9.7–11.9 ms |
| glob walk — runs on every call regardless | 25 ms |
| reading + hashing file contents | ~200 ms |
| Fingerprinting nested files: discover + digest all of them | 8.0 ms |
| A root-only chain build (2 reads, no walk) | 0.18 ms |

Second correction: I described memoisation as removing "the repeated recursive walk on the hot path", which is true but implies more than it delivers. The memo's ceiling is the ~10 ms build, not the 25 ms cold/warm delta — that delta includes cache-clearing and run-to-run variance. Against a 243 ms read the memo saves roughly 4%. The dominant cost is reading and hashing every file, which no memo touches.

## Options

**A — fingerprint the nested files too.** Discover every `.gitignore`/`.socraticodeignore` under the artifact directory and digest them alongside the root pair. Fully correct, nested support unchanged. The measured cost is 8.0 ms to avoid a 9.7 ms rebuild, so the memo's remaining saving is about 1.7 ms in 243.

**B — drop the memo.** Fully correct, nested support unchanged, no invalidation surface to get wrong. Costs the ~10 ms build per artifact per staleness check, i.e. on every `codebase_context_search`, and reinstates the per-search `createIgnoreFilter` info lines that the memo was introduced to quiet.

**C — narrow the artifact chain to the artifact root, keep the memo.** No nested walk: the build drops from 11.9 ms to 0.18 ms, and the content digest already on the branch then covers 100% of the chain's inputs, so the memo becomes complete with no residual. Reads get faster than the current branch. The cost is scope — a `.gitignore` nested inside an artifact directory stops being honoured.

Three things bearing on C's scope cost, offered as facts rather than argument. On `main`, artifacts honour no ignore files at all, so root-only is still an increase over the current release rather than a reduction. Surveying two real repositories, every nested `.gitignore` I found sits either at a package root (`extension/.gitignore` here) or inside `.pytest_cache/` and `.ruff_cache/`, which `dot: false` already excludes — none under a plausible artifact path, though that is two repositories and not a survey. And every test on the branch places ignore files at the artifact root, so the suite passes unchanged under C; the nested case has no coverage today either way.

If C appeals, it can be a `nested: false` option on `createIgnoreFilter` so artifacts keep sharing your defaults and `shouldIgnore` rather than growing a parallel chain. A variant drops the memo as well, since a 0.18 ms build makes the cache pointless except for suppressing the repeated log lines — at the price of assembling the chain in the artifacts module.

## Where this leaves it

A and B keep the semantics you have already reviewed and differ in what they spend; C changes those semantics in exchange for being both correct and cheaper than the branch. That is a call about how far the feature reaches on your hot path, and you have already ruled once on this trade-off, so I would rather not pre-empt it.

Tell me which and I will implement it, with the wording corrections either way. Reproduction and benchmark scripts are throwaway probes rather than committed files; happy to paste either if you want to run them yourself.

gregoryfoster:
@giancarloerra LMK if you have a preference, otherwise I'm leaning towards option A for completeness.

giancarloerra:
> @giancarloerra LMK if you have a preference, otherwise I'm leaning towards option A for completeness.

Sorry, I think let's go for B. Your own numbers make the memo not worth its invalidation surface, and thread the content from ensureArtifactsIndexed into indexArtifact so the second full read disappears, which is worth more than the memo ever saved. Demote the two ignore log lines to debug for the noise.

gregoryfoster:
Option B, both additions, in `e703244`.

**Memo removed.** The chain is rebuilt per read. Gone with it: the fingerprint, `clearIgnoreFilterCache`, its DEVELOPER row, and the whole `context-artifacts-ignore-cache.test.ts` file. Net −95 lines against the previous head.

**Content threaded from `ensureArtifactsIndexed` into `indexArtifact`.** Optional `preread` parameter; `indexAllArtifacts` still reads for itself. You were right that this is worth more than the memo ever saved, and by a wider margin than I expected: the second read is a *full* `readArtifactContent` — 243 ms on the 3,600-file shape, 34 ms on the 480-file one — against the memo's ~10 ms ceiling.

It also closed a correctness gap I had not been framing as one. The hash used to come from one read and the indexed content from a second, so a file changing between them would store a hash that never matched what was actually indexed. One read, so they cannot disagree.

**Two `createIgnoreFilter` status lines demoted to debug**, with a line in that function's JSDoc saying why.

Tests: the six memo tests are gone, four threading tests replace them in `context-artifacts-preread.test.ts` — preread content is what gets indexed, `indexArtifact` still reads when handed nothing, staleness-read exclusions carry into what is indexed, and an unchanged artifact is left alone. Mutation-checked: making `indexArtifact` ignore its `preread` argument fails the first of those. 1181 unit tests across 53 files (1183 − 6 memo + 4 threading), 17 integration tests, biome and `tsc --noEmit` clean.

The PR body's staleness-path section is rewritten to describe the no-memo design and carries the measurements, so the thread's rationale does not have to be reconstructed from comments.

giancarloerra:
Option B is right and the staleness class is gone rather than relocated. Verified end to end on live Qdrant: adding `docs/sub/.gitignore` now drops the file on the very next read (`reindexed: ["docs"]`), editing one flips which file survives, and the same script against `main` and against the memo head keeps the excluded file indexed forever. A 26-case invalidation matrix passes, and there is no memo, fingerprint or module-level mutable state left in either file.

The threading is sound too: the indexed bytes reassembled from real Qdrant chunks hash identically to what the staleness read produced, `codebase_context_index` remains byte-for-byte identical to the ensure path on chunk count, ids, payloads and text, hostile prereads (wrong artifact, stale, fabricated counts, directory deleted mid-flight) all self-heal on the next ensure, and the second full read is genuinely gone: 3 reads down to 2, a median 434 ms saved on the 3,600-file shape.

Two things before merge.

**1. The round has no regression test.** Two separate mutations leave the whole suite green:

```
drop the 4th argument at the ensureArtifactsIndexed call site   ->  1181/1181 pass
reintroduce a module-level ignore-filter memo with no signature ->  1181/1181 pass
```

The first reverts the mechanism this round adds; the second is strictly staler than the memo we just removed. The four new tests call `indexArtifact` directly, so they pin the callee and never the call site, and the six deleted memo tests were the only coverage of ignore-rule freshness. Two assertions close both: index once, add `<artifact>/sub/.gitignore`, assert the next `ensureArtifactsIndexed` drops that file; and count `readFile` calls, which separates the heads exactly at 2 against 3. Without them CI stays green while a future refactor quietly restores the #116 symptom.

**2. One claim in the PR body is false.** "A file changing between them would have stored a hash that never matched the indexed content" does not hold: on `main`, `indexArtifact` derived both the indexed content and the stored hash from its own single read, so they could not diverge. Executed with a file flipped the instant the staleness read returned, all three heads store a hash that exactly describes the chunks written. What threading actually buys is one snapshot for both the decision and the write, plus one fewer full read, which is what your commit body already says. Only the PR body overstates it.

Nothing needed on the commit trailers: the squash message is written at merge time, as with your last one.

Two documentation items you may as well fold in, both verified: `README.md:999` lists `target` among the defaults to re-include with `!target`, but `findNestedGitignores` skips `target/` unconditionally, so a `.gitignore` inside a re-included `target/` is never read (the same layout under `env` behaves correctly). And `README.md:996` reads symmetrically while the two files are not: `.socraticodeignore` is read once at the artifact root and never recurses, while `.gitignore` is walked through the subtree.

giancarloerra:
I am going to finish the remaining merge work directly on this branch.

I will keep the accepted scope and limit the changes to the items in [the last review](https://github.com/giancarloerra/SocratiCode/pull/117#issuecomment-5452940078): the two missing regressions, the PR-body correction, and the two README corrections. No further action is needed from you while I complete that work.

giancarloerra:
@coderabbitai full review

PR #118, opened 2026-08-26
fix(graph): resolve Rust imports through Cargo-declared crate roots

SIRTHEO:
## Summary

On Rust projects the file graph is the list of a crate's private modules and nothing else. Two causes combine:

- `graph-imports.ts` extracts module declarations with `/^mod\s+(\w+)\s*;/`, which no `pub mod` matches, and `use` declarations with `/^use\s+/`, which no `pub use` matches. Neither reaches the resolver.
- `graph-resolution.ts`, `case "rust"`: only specifiers **without** `::` resolve. Every `use crate::…`, `super::…`, `self::…` and `other_crate::…` hits `return null`.

Rust was the only language marked Full Support with no manifest-derived name map — Go reads `go.mod`, PHP `composer.json`, Dart `pubspec.yaml`, Python `pyproject.toml` since #112. Rust has `Cargo.toml`, which nothing read.

The second commit closes the gap the first one left: **a Rust path means what it means from the point in the source that writes it**, and four things a file says about its own position were still being read off the file instead — an inline `mod` block, a `#[path]` attribute, a raw identifier, and the edition the manifest declares.

### Reading the manifest (first commit)

- `buildRustCrateMap` records every crate the tree declares: its importable name (`[lib] name` or `[package] name`, dashes turned into underscores, as Cargo does), its library root, and every target that is the top of its own module tree — library, binaries, integration tests, examples, benches, build script — declared in the manifest or autodiscovered the way Cargo does. Discovery walks the filesystem with the graphable walk's ignore filter (`Cargo.toml` is never in `fileSet` — the #82 trap) and skips `target/`, where Cargo unpacks a manifest for every dependency it builds. The manifest is parsed with `smol-toml`, already in the tree from #112.
- `resolveRustImport` resolves `crate::` against the file's own crate root, `super::` and `self::` against its position in the module tree, and a leading segment naming a crate in this project against that crate's library root. Trailing segments name items, not modules, so the longest prefix that names a file wins.
- The module-declaration and `use` regexes accept a visibility modifier, and a `use` tree is flattened into one path per leaf.

Two Rust layout rules the old resolver got wrong are now honoured. `mod bar;` inside `src/foo.rs` is `src/foo/bar.rs`, not `src/bar.rs` — a crate root and a `mod.rs` own the directory they sit in, every other file owns the directory named after it. And a binary at `src/bin/tool.rs` is its own crate root.

### Reading the position the source writes from (second commit)

- **An inline `mod` block is a module level the filesystem does not show.** `#[cfg(test)] mod tests { use super::*; }` sits in a large share of all Rust files, and every `super::` inside it was counted one level too high. `extractImports` now walks the chain of enclosing `mod` blocks and rebases the path: a `super::` inside one inline level names the file itself, which is no edge, and it takes one more to leave the file. A `mod` declared inside `mod outer { … }` lands under `outer/`, which is where rustc looks.
- **`#[path = "…"]`** is read off the attribute preceding the declaration and travels as the path it names. The two forms count from different places, which is the part that is easy to get wrong: a declared module resolves it against the directory the declaring file sits in, while one written inside an inline `mod` resolves it against that file's own module directory, one directory deeper per inline level. Both were checked against rustc.
- **A raw identifier names the module it escapes**: `mod r#async;` declares `async`, whose file is `async.rs`. Left alone, `crate::r#async::poll` resolved to nothing and fell back to the crate root.
- **`extern crate serde;`** is extracted. The crate it names cannot collide with a local module of the same name — rustc rejects that — so the bare name resolves the way a `mod` declaration does.
- **Three markers a `use` writes are kept.** Comments inside a `use` tree no longer become path segments, nor do they end the walk to a `#[path]` attribute above the `mod` it belongs to. A group leaf renamed with `as` is the same leaf — comparing it before removing the alias dropped `use crate::config::{self as cfg};` whole. And a leading `::` says the head names a crate rather than a module in scope, which matters precisely because a local module now wins otherwise.
- **A dependency renamed in the manifest** (`dep = { package = "real-name" }`) is followed. `RustCrate` gains `aliases`, read from `[dependencies]`, `[dev-dependencies]`, `[build-dependencies]`, `[target.*.dependencies]` and `[workspace.dependencies]` via `workspace = true`. The same crate answers to different names in two members of one workspace, so the map is consulted from the importing package.
- **Target autodiscovery is turned off from the manifest**: `autobins`, `autotests`, `autoexamples`, `autobenches`, `autolib`, and `build = false`. `RustCrate` gains `edition`, defaulting to `"2015"` when the key is absent as Cargo does — and in 2015 declaring one target by hand stops the discovery of the rest of its kind.
- **A build directory under another name** (`CARGO_TARGET_DIR`) is recognised by the markers Cargo writes into it (`CACHEDIR.TAG`, `.cargo-ok`) rather than by the name `target`. The comment that promised this defence now matches the code.

Two orderings now follow rustc rather than convenience (an earlier draft listed a third here, the unanchored 2015 path; later rounds turned it off and it is a known limit below):

- **A module in scope wins over a crate of the same name.** The extern prelude is consulted last. Reading the crate names first drew an edge into an unrelated sibling crate whenever a local module carried a published crate's name — `config`, `log` and `utils` are both.
- **A path that names the file it was written in is not an edge.** `use crate::db::Connection;` inside `#[cfg(test)] mod tests` in `db.rs` names that same file.

`rustRootForFile` is memoised per crate map: it answers the same for every import in a file, and a hundred-crate workspace asked it a hundred and twenty thousand times.

### What the review rounds added

- **A macro body is read where rustc reads it.** `cfg_if! { … mod unix; … }` and the `macro_rules!` wrappers tokio, libc and clap write around their module declarations hid every module they declare — of the 487 files under tokio's `src/` that are not a crate root, 323 had no edge into them on `main`; none has now. The invocation's head and braces are blanked so the body parses in place, and the pass is kept only while every parse error sits inside a macro that was unwrapped: recovery that reaches past the macro (which is what happens to a body inside an inline `mod`) drops the pass, first for that body, then for the file. The shape is pinned by the cargo-oracle test below, with `compile_error!` files where a spilled declaration would land.
- **The manifest is read the way Cargo reads it**: `[workspace.dependencies]` inherited by members, `[patch]` and `[replace]` redirecting a registry name to a project crate, a renamed dependency, a dash-named one imported under its underscore, and the edition a workspace declares — each checked against `cargo metadata`.
- **A leading `::`** is read by the edition that wrote it, and **`include!("…")`** draws an edge, since the file it names is compiled into this one.
- **The cargo-oracle test runs in CI** (`rust-graph-vs-cargo` job, needs a toolchain), rather than counting as coverage only where a toolchain happens to exist.
- **A package reaches only the crates it declares.** `use helper::Thing` in a package whose manifest names no `helper` is E0432 to rustc, yet the raw name used to be searched across every crate of the workspace and drew an edge into the undeclared sibling. The fence is what the manifest declares, admitted under the name rustc knows the crate by: a renamed alias, or the *library* name of a crate declared under its package name (`other = { path = … }` with `[lib] name = "otherlib"` is imported as `otherlib`). The package's own library is the one name reached without declaring it; a package whose manifest cannot be parsed is not fenced. Cargo-verified in a workspace whose undeclared sibling carries a `compile_error!` and whose declared one sets a `[lib] name`.
- **The manifest that governs a file is the innermost one containing it**, and the root's `"."` is the empty prefix rather than a one-character directory. Measured by length it tied with every member directory of one character and won the tie, so every file of `a/` was governed by the workspace root: with a `[workspace]`-only root that declares no dependency and no edition, `a`'s path dependencies went unreachable and its 2021 files were read as 2015 — which turned `use ::bb::Thing;` into an edge to `a`'s own `bb` module. Cargo-verified in a workspace with a one-letter member and a second package nested inside it, because depth decides two orderings and not one: root against member, and a member against the one nested in it. Ranking by raw length turns both halves of the first red; answering "0 for the root, 1 for everything else" turns the second red.
- **In edition 2015 a leading `::` is answered from the crate root**, never from the importing file's own declarations: a child's `#[path] mod foo;` used to capture `use ::foo::Item` that names the root's `foo`. A name at the root is a module of the root, else a crate the package declares (a registry crate resolves to nothing), else an item of the root — the root file, unless rustc provides the name itself. Cargo-verified with a child whose local `foo` defines no `Item`.

## Type of change

- [ ] Documentation update

## Testing

Measured against the **real `buildCodeGraph`**, on two Rust trees, at all three states:

| | `main` | first commit | now |
|---|---|---|---|
| A — 88 `.rs` files, 11-crate Cargo workspace | 34 edges | 680 | **701** |
| B — 149 `.rs` files, 10-crate Cargo workspace | 64 edges | 621 | **619** |

The second commit moves few edges, and that is the point: it removes the wrong ones and adds the right ones. On tree A, **six false edges go and fifteen true ones arrive** — every one of the six came from a `super::` written inside `#[cfg(test)] mod tests`, landing at the parent module the source never imports and closing a cycle with the `mod` declaration pointing the other way. The fifteen are the test-support modules those blocks really import, two levels up. Each was checked against the source: `entry.rs` writes `use super::super::test_support::{…}` inside its test block, and `flows.rs` declares `mod test_support;`. On tree B the edge set is unchanged and two redundant specifiers stop resolving.

**The crate map is checked against `cargo metadata --no-deps`**, which is the only authority on what a project's targets are. On both real workspaces and on a tree built to hold every awkward declaration — `[lib] name` different from the package, every target path declared by hand, a build script renamed, autodiscovery off, a proc-macro crate, a renamed dependency, edition 2015 — the targets and importable names agree **exactly**: 26/26, 15/15, 8/8.

Where the layout rules were in doubt, **rustc decided, not us**: each case was built as a crate with the file in one candidate position, and `cargo check` said which one compiles. That is how four things were settled — a binary in `src/bin/` owns the directory it sits in (so a module of `src/bin/tool.rs` is `src/bin/helper.rs`, not `src/bin/tool/helper.rs`); `#[path]` on a declared module is relative to the declaring file's own directory, while inside an inline block it is not; a module declared inside an inline `mod` lands under it; and a member under `[workspace] exclude` is still importable by a member that depends on it by path — which is why an earlier attempt to treat `exclude` as "not importable" was withdrawn rather than kept. It also corrected a test on this branch that asserted nine targets where Cargo reports six.

Every new guard is mutation-tested: **seventeen mutations, seventeen red batteries**. The one test the first commit left green on a vacuous assertion (`every(e => e.target.endsWith(".rs"))`, true on an empty set) now asserts the edges leaving the file that imports a third-party crate, and fails when the inline-module walk is disabled.

On three public trees against `main` at `3d99883`, edge lists sorted and diffed line by line — not counts:

| | `main` | this branch | this branch, distinct pairs |
|---|---|---|---|
| ripgrep | 76 | **612** | 261 |
| tokio | 167 | **3,922** | 2,210 |
| clap | 78 | **1,196** | 507 |

The third column is there because the first two count edge *records*, and a Rust `use` tree yields one per leaf — see the known limit on it below. `main` has no duplicates to speak of because almost nothing resolved.

The only edges `main` draws that this branch does not are four that pointed at a sibling where rustc reads the child. Python, TypeScript, JavaScript and Go trees come out identical edge for edge. Building tokio's graph costs 1,286 → 2,096 ms (nine paired runs, median), which is the price of reading the modules the macros declare.

**Cargo is the oracle** in `tests/integration/rust-graph-vs-cargo.test.ts`: a fixture crate is built, its dep-info (`target/debug/deps/*.d`) lists what rustc compiled, and every edge the graph draws must land on that list — files that must *not* be read carry `compile_error!`, so a wrong edge is one a clean build proves false. It runs in CI when a toolchain is present.

- [x] Unit tests pass (`npm run test:unit`) — **1,608 passed**, 4 skipped, on macOS and in `node:18`, `node:20` and `node:22` containers
- [x] Integration tests pass (`npm run test:integration`) — full suite **1,818 passed** with Ollama, cargo-oracle test included
- [x] Lint clean (`npm run lint`)

## Known limits

- A path with a bare head written inside an inline module is rebased onto that module only when the block declares a module by that name (`mod tests { mod fixtures; use fixtures::build; }`). Any other bare head is left alone, because the commoner shape by far is `mod tests { use some_crate::Thing; }` — how a test reaches another crate of the project — and rebasing that would lose the edge. A module in scope through some other route than a declaration in the same block is the case left uncovered.
- `crate::missing_module::Item` resolves to the crate root rather than to nothing: no prefix of the path names a file, and the root is where the module would have been declared. It is the same fallback that makes `crate::SomeType` reach the root correctly, and separating the two would take knowing which segments name items. In edition 2015 a leading `::` shares it for a name the manifest does not declare and rustc does not provide: `::RootItem` is the root, and so is `::nothing::Item`.
- A `mod` under `#[cfg(…)]` is always followed, whatever the feature flags say. The graph has no feature resolution, and a module behind a flag is still a file in the tree.
- **An unanchored path in edition 2015 is absolute from the crate root, and that reading is not acted on.** `use registry::write;` in `src/client.rs` compiles with `mod registry;` written only in `lib.rs`, and no edge is drawn. Acting on it means answering a path with no declaration behind it, and each way of bounding that produced a fresh edge rustc rejects — a bare `use b;` between two integration-test crates, an undeclared sibling file, a `mod` in a nested block, a virtual workspace's root claiming its members' declarations — every one found by a reviewer running cargo. `main` drew none of these edges either, so the gate costs nothing already published. The reasoning sits above `edition2015` in `graph-resolution.ts`.
- **One edge record per `use` leaf.** `use crate::io::{A, B, C};` resolves three times to the same file and pushes three records; nothing deduplicates, here or for any other language. On these trees that is 612 records for 261 distinct pairs (ripgrep), 3,922 for 2,210 (tokio). It matters to what reads `graph.edges` as connections — `graph_stats.totalEdges`, `avgDependencies`, `mostConnected`, `findCircularDependencies` and the HTML visualisation all count records — and to `node.imports`, which grows the same way: `qdrant.ts` sums it as `importCount`, and `graph-tools.ts` prints the same specifier twice under `Imports (N)`. Deduplicating at `edges.push` would change every language's numbers, so it is not folded into a Rust bug fix; happy to send it separately.
- **A package whose `Cargo.toml` cannot be parsed is not fenced.** `use helper::Thing;` in such a package still reaches an undeclared sibling. The alternative costs that package every cross-crate edge on a manifest we merely failed to read.
- **Two `#[path]` declarations of one name under opposite `cfg`s: the last one wins for paths through it.** Both declarations draw their own edge, so neither file is unreachable, but `pub use sys::Handle;` beside `#[cfg(unix)] #[path = "sys_unix.rs"] mod sys;` and its `windows` twin resolves to whichever was written last, on every platform. The graph has no feature or target resolution.
- A macro body is read only while blanking the invocation leaves the rest of the file parsing as it did. Where it does not — a `cfg_if!` inside an inline `mod` or `impl`, whose arms close the block early, or libc's `freebsd/mod.rs`, which parses as one error from line 1 — the body is left alone and its declarations draw no edge. A macro written inside a `cfg_if!` arm is not reached either: after blanking, the arm is the block of a recovered `if`, and a block is not where an item is read from. Measured over the 153 crates of a 1,256-crate registry cache that carry the shape, that is 56 dependencies not read; ripgrep, tokio and clap lose none. Reading each body in isolation with the invocation's scope would remove both limits and is a follow-up.

## Note on a duplication I did not introduce

`findCargoManifests` is the fifth near-identical manifest walk in this file, after the ones for `go.mod`, `composer.json`, `pubspec.yaml` and `pyproject.toml` — they differ only in the filename and the one directory each skips. I followed the existing convention rather than fold a refactor into a bug fix. Happy to send one that collapses all five into a parameterised walk, separately, if you want it.

## Checklist

- [ ] I have updated documentation (README.md / DEVELOPER.md) if needed
- [ ] I have addressed all [CodeRabbit](https://coderabbit.ai) review comments (or marked as resolved with explanation)

## Related issues

No issue open for this one — happy to file one if you would rather have the report separate from the fix.

Related: #111. This adds a thirteenth parameter to `resolveImport` and a sixth `build*Map` helper, so the count in DEVELOPER.md moves again; I left the docs alone rather than fold them in here.

Structurally this follows #112 (Python) and #108 (Dart): the same disease, a manifest the resolver never read.

## Summary by CodeRabbit

* **New Features**
  * Improved Rust project analysis across Cargo workspaces, editions, dependency aliases, custom targets, and `#[path]`-mapped modules.
  * Added broader support for Rust macros, inline modules, re-exports, `include!`, `extern crate`, raw identifiers, and conditional path attributes.
  * Added Elixir module and template analysis, plus improved PHP namespace and file inclusion resolution.
  * Graph status now reports the builder version.

* **Bug Fixes**
  * Prevented incorrect links to unrelated, undeclared, shadowed, or self-referencing modules.
  * Improved resolution for nested modules and edition-specific import paths.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,3 +61,27 @@ jobs:
           npm init --yes
           npm install --engine-strict "$package_file"
           test -f node_modules/socraticode/dist/index.js
+
+  # The one test that asks the compiler instead of asserting what we believe.
+  # It lives under tests/integration/ because it runs a real `cargo check`, and
+  # the unit job above does not reach it — so it gets a job of its own rather
+  # than being counted as covered while never running.
+  #
+  # It skips itself where cargo is absent, which is what makes this job's
+  # `rustc --version` step matter: without it a toolchain that quietly stopped
+  # being installed would turn the test into a silent no-op, and a silent no-op
+  # is how the two regressions this test exists to catch reached main in the
+  # first place.
+  rust-graph-vs-cargo:
+    name: Rust graph vs cargo
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+          cache: npm
+      - name: Fail if the Rust toolchain is missing
+        run: rustc --version && cargo --version
+      - run: npm ci
+      - run: npx vitest run tests/integration/rust-graph-vs-cargo.test.ts
--- a/src/services/code-graph.ts
+++ b/src/services/code-graph.ts
@@ -1012,11 +1012,21 @@ export async function buildCodeGraph(
     // and the declaration is the only evidence of that in the source: matching
     // a neighbouring file by name instead let a third-party head capture the
     // import whenever a file of that name happened to exist.
+    //
+    // The name comes from the declaration itself and not from its specifier.
+    // Reading the specifier's last segment got both ends wrong: a
+    // `#[path = "custom.rs"] mod foo;` recorded `custom.rs` and lost every
+    // `use foo::Item;` in that file, and a `mod bar;` written inside
+    // `mod outer { … }` recorded `bar` as if the file declared it — which
+    // handed a same-named neighbouring file back the capture this gate exists
+    // to stop. Both checked with cargo 1.70.0 and 1.98.0: the attribute form
+    // compiles with `foo` in scope, and at the file's own level a name
+    // declared inside an inline block reaches the dependency instead, or
+    // fails with E0432 when there is none.
     const declaredMods = new Set<string>();
     for (const imp of importInfos) {
-      if (imp.isModuleDeclaration) {
-        const segments = imp.moduleSpecifier.split("::");
-        declaredMods.add(segments[segments.length - 1]);
+      if (imp.isModuleDeclaration && imp.declaredName) {
+        declaredMods.add(imp.declaredName);
       }
     }
 
--- a/src/services/code-graph.ts
+++ b/src/services/code-graph.ts
@@ -1023,13 +1023,15 @@ export async function buildCodeGraph(
     // compiles with `foo` in scope, and at the file's own level a name
     // declared inside an inline block reaches the dependency instead, or
     // fails with E0432 when there is none.
-    // Only a declaration carries a declared name, so asking for the name is
-    // the whole question — pairing it with `isModuleDeclaration` would add a
-    // condition no tree can make false, and a guard nothing can fail is a
-    // guard nobody can check.
-    const declaredMods = new Set<string>();
+    // The map carries the file with the name, because half the declarations
+    // move it: `#[path = "custom.rs"] mod foo;` names `foo` and files it at
+    // `custom.rs`, and a resolver holding only the name looks for `src/foo.rs`,
+    // finds nothing, and falls through to a library called `foo` if the
+    // workspace has one — an edge into an unrelated crate. Where the
+    // declaration does not move anything, the value is the name itself.
+    const declaredMods = new Map<string, string>();
     for (const imp of importInfos) {
-      if (imp.declaredName) declaredMods.add(imp.declaredName);
+      if (imp.declaredName) declaredMods.set(imp.declaredName, imp.moduleSpecifier);
     }
 
     for (const imp of importInfos) {
@@ -1038,7 +1040,7 @@ export async function buildCodeGraph(
       // Try to resolve to a project file
       // CSS imports from <style> blocks use CSS resolution even when the source file is Svelte/Vue
       const resolutionLanguage = imp.isCssImport ? "css" : language;
-      const resolved = resolveImport(imp.moduleSpecifier, absolutePath, resolvedPath, fileSet, resolutionLanguage, aliases, jvmSuffixMap, csNamespaceMap, goModuleInfo, phpPsr4Map, dartPackageMap, pythonRootsFor(relPath), elixirModuleMap, phpFqcnMap, rustCrates, declaredMods);
+      const resolved = resolveImport(imp.moduleSpecifier, absolutePath, resolvedPath, fileSet, resolutionLanguage, aliases, jvmSuffixMap, csNamespaceMap, goModuleInfo, phpPsr4Map, dartPackageMap, pythonRootsFor(relPath), elixirModuleMap, phpFqcnMap, rustCrates, declaredMods, imp.isModuleDeclaration === true);
       if (resolved) {
         node.dependencies.push(resolved);
 
--- a/src/services/code-graph.ts
+++ b/src/services/code-graph.ts
@@ -1007,13 +1007,26 @@ export async function buildCodeGraph(
       });
     }
 
+    // Which modules this file brings into the tree by declaring them. A Rust
+    // path with an unanchored head can only reach a module the file declares,
+    // and the declaration is the only evidence of that in the source: matching
+    // a neighbouring file by name instead let a third-party head capture the
+    // import whenever a file of that name happened to exist.
+    const declaredMods = new Set<string>();
+    for (const imp of importInfos) {
+      if (imp.isModuleDeclaration) {
+        const segments = imp.moduleSpecifier.split("::");
+        declaredMods.add(segments[segments.length - 1]);
+      }
+    }
+
     for (const imp of importInfos) {
       node.imports.push(imp.moduleSpecifier);
 
       // Try to resolve to a project file
       // CSS imports from <style> blocks use CSS resolution even when the source file is Svelte/Vue
       const resolutionLanguage = imp.isCssImport ? "css" : language;
-      const resolved = resolveImport(imp.moduleSpecifier, absolutePath, resolvedPath, fileSet, resolutionLanguage, aliases, jvmSuffixMap, csNamespaceMap, goModuleInfo, phpPsr4Map, dartPackageMap, pythonRootsFor(relPath), elixirModuleMap, phpFqcnMap, rustCrates);
+      const resolved = resolveImport(imp.moduleSpecifier, absolutePath, resolvedPath, fileSet, resolutionLanguage, aliases, jvmSuffixMap, csNamespaceMap, goModuleInfo, phpPsr4Map, dartPackageMap, pythonRootsFor(relPath), elixirModuleMap, phpFqcnMap, rustCrates, declaredMods);
       if (resolved) {
         node.dependencies.push(resolved);
 
--- a/src/services/code-graph.ts
+++ b/src/services/code-graph.ts
@@ -506,6 +506,13 @@ export async function getGraphStatus(projectPath: string): Promise<{
 // ── Register dynamic language grammars ───────────────────────────────────
 
 let dynamicLangsRegistered = false;
+/**
+ * The declaration map a path written inside an inline `mod { … }` block is
+ * answered with: empty, because the file's own declarations are not in that
+ * block's scope. Shared and never written.
+ */
+const EMPTY_DECLARED_MODS: Map<string, string> = new Map();
+
 const loadedDynamicLanguages = new Set<string>();
 const failedDynamicLanguages = new Map<string, string>();
 
@@ -1029,6 +1036,8 @@ export async function buildCodeGraph(
     // finds nothing, and falls through to a library called `foo` if the
     // workspace has one — an edge into an unrelated crate. Where the
     // declaration does not move anything, the value is the name itself.
+    // Shared and never written: one empty map for every inline-block path in
+    // the build, instead of one per import.
     const declaredMods = new Map<string, string>();
     for (const imp of importInfos) {
       if (imp.declaredName) declaredMods.set(imp.declaredName, imp.moduleSpecifier);
@@ -1040,7 +1049,15 @@ export async function buildCodeGraph(
       // Try to resolve to a project file
       // CSS imports from <style> blocks use CSS resolution even when the source file is Svelte/Vue
       const resolutionLanguage = imp.isCssImport ? "css" : language;
-      const resolved = resolveImport(imp.moduleSpecifier, absolutePath, resolvedPath, fileSet, resolutionLanguage, aliases, jvmSuffixMap, csNamespaceMap, goModuleInfo, phpPsr4Map, dartPackageMap, pythonRootsFor(relPath), elixirModuleMap, phpFqcnMap, rustCrates, declaredMods, imp.isModuleDeclaration === true);
+      // A bare path written inside an inline `mod { … }` block is answered
+      // without the file's declarations: they are not in that block's scope,
+      // and handing them over drew an edge rustc rejects with E0432. The map is
+      // emptied rather than omitted — omitting it turns the gate off entirely,
+      // which is the looser reading, not a stricter one. Edition 2015 is
+      // unaffected: there the path is absolute from the crate root and never
+      // consulted the map to begin with.
+      const scopedMods = imp.fromInlineBlock ? EMPTY_DECLARED_MODS : declaredMods;
+      const resolved = resolveImport(imp.moduleSpecifier, absolutePath, resolvedPath, fileSet, resolutionLanguage, aliases, jvmSuffixMap, csNamespaceMap, goModuleInfo, phpPsr4Map, dartPackageMap, pythonRootsFor(relPath), elixirModuleMap, phpFqcnMap, rustCrates, scopedMods, imp.isModuleDeclaration === true);
       if (resolved) {
         node.dependencies.push(resolved);
 
--- a/src/services/code-graph.ts
+++ b/src/services/code-graph.ts
@@ -15,7 +15,7 @@ import { ensureElixirTemplateParsers, isElixirTemplateExtension } from "./elixir
 import { detectExtensionFromSource, resolveExtensionlessExtension } from "./extensionless.js";
 import { loadPathAliases } from "./graph-aliases.js";
 import { extractImports } from "./graph-imports.js";
-import { buildCsNamespaceMap, buildDartPackageMap, buildElixirModuleMap, buildGoModuleInfo, buildJvmSuffixMap, buildPhpFqcnMap, buildPhpPsr4Map, buildPythonManifests, pythonRootsForFile, resolveImport } from "./graph-resolution.js";
+import { buildCsNamespaceMap, buildDartPackageMap, buildElixirModuleMap, buildGoModuleInfo, buildJvmSuffixMap, buildPhpFqcnMap, buildPhpPsr4Map, buildPythonManifests, buildRustCrateMap, pythonRootsForFile, resolveImport } from "./graph-resolution.js";
 import { computeUnresolvedPct, resolveCallSites } from "./graph-symbol-resolution.js";
 import { extractSymbolsAndCalls, rawCallsToUnresolvedEdges } from "./graph-symbols.js";
 import { createIgnoreFilter, shouldIgnore } from "./ignore.js";
@@ -881,6 +881,17 @@ export async function buildCodeGraph(
   const hasElixir = files.some((f) => [".ex", ".exs"].includes(path.extname(f).toLowerCase()));
   const elixirModuleMap = hasElixir ? buildElixirModuleMap(fileSet, resolvedPath) : undefined;
 
+  // Record every crate the tree declares, from each Cargo.toml (discovered by
+  // walking, like go.mod and pubspec.yaml — Cargo.toml is never in the
+  // graphable file set). A Rust path names a position in a module tree
+  // (`crate::`, `super::`) or another crate by name, neither of which the
+  // resolver could follow without knowing where each crate's root module sits:
+  // every specifier containing `::` resolved to null, so the file graph held
+  // only bare `mod` declarations. An empty list keeps `mod`, `super` and `self`
+  // resolving from the file's own position, as before.
+  const hasRust = files.some((f) => path.extname(f).toLowerCase() === ".rs");
+  const rustCrates = hasRust ? buildRustCrateMap(fileSet, resolvedPath) : undefined;
+
   for (const relPath of files) {
     let ext = path.extname(relPath).toLowerCase();
     let lang = getAstGrepLang(ext);
@@ -1002,7 +1013,7 @@ export async function buildCodeGraph(
       // Try to resolve to a project file
       // CSS imports from <style> blocks use CSS resolution even when the source file is Svelte/Vue
       const resolutionLanguage = imp.isCssImport ? "css" : language;
-      const resolved = resolveImport(imp.moduleSpecifier, absolutePath, resolvedPath, fileSet, resolutionLanguage, aliases, jvmSuffixMap, csNamespaceMap, goModuleInfo, phpPsr4Map, dartPackageMap, pythonRootsFor(relPath), elixirModuleMap, phpFqcnMap);
+      const resolved = resolveImport(imp.moduleSpecifier, absolutePath, resolvedPath, fileSet, resolutionLanguage, aliases, jvmSuffixMap, csNamespaceMap, goModuleInfo, phpPsr4Map, dartPackageMap, pythonRootsFor(relPath), elixirModuleMap, phpFqcnMap, rustCrates);
       if (resolved) {
         node.dependencies.push(resolved);
 
--- a/src/services/code-graph.ts
+++ b/src/services/code-graph.ts
@@ -1036,8 +1036,6 @@ export async function buildCodeGraph(
     // finds nothing, and falls through to a library called `foo` if the
     // workspace has one — an edge into an unrelated crate. Where the
     // declaration does not move anything, the value is the name itself.
-    // Shared and never written: one empty map for every inline-block path in
-    // the build, instead of one per import.
     const declaredMods = new Map<string, string>();
     for (const imp of importInfos) {
       if (imp.declaredName) declaredMods.set(imp.declaredName, imp.moduleSpecifier);
--- a/src/services/code-graph.ts
+++ b/src/services/code-graph.ts
@@ -1023,11 +1023,13 @@ export async function buildCodeGraph(
     // compiles with `foo` in scope, and at the file's own level a name
     // declared inside an inline block reaches the dependency instead, or
     // fails with E0432 when there is none.
+    // Only a declaration carries a declared name, so asking for the name is
+    // the whole question — pairing it with `isModuleDeclaration` would add a
+    // condition no tree can make false, and a guard nothing can fail is a
+    // guard nobody can check.
     const declaredMods = new Set<string>();
     for (const imp of importInfos) {
-      if (imp.isModuleDeclaration && imp.declaredName) {
-        declaredMods.add(imp.declaredName);
-      }
+      if (imp.declaredName) declaredMods.add(imp.declaredName);
     }
 
     for (const imp of importInfos) {
--- a/src/services/graph-imports.ts
+++ b/src/services/graph-imports.ts
@@ -240,20 +240,141 @@ function splitRustUseList(inner: string): string[] {
 }
 
 /**
+ * The module a raw identifier names. `mod r#async;` declares the module
+ * `async`, whose file rustc looks for as `async.rs` — the `r#` is how the
+ * source escapes a keyword, not part of the name. Without stripping it, the
+ * path `crate::r#async::poll` resolves to nothing and falls back to the crate
+ * root, which draws an edge at the wrong file.
+ */
+function stripRawIdent(segment: string): string {
+  return segment.startsWith("r#") ? segment.slice(2) : segment;
+}
+
+/**
  * One leaf of a `use` tree as a module path: the alias is dropped, and so are
  * trailing `self` and `*`, which name the module the path already reached
  * rather than something under it.
  */
 function rustUseLeafPath(leaf: string): string {
   const segments = leaf
-    .replace(/\s+as\s+(?:\w+)\s*$/, "")
+    .replace(/\s+as\s+(?:r#)?\w+\s*$/, "")
     .split("::")
-    .map((segment) => segment.trim())
+    .map((segment) => stripRawIdent(segment.trim()))
     .filter(Boolean);
   while (segments.length > 0 && ["self", "*"].includes(segments[segments.length - 1])) {
     segments.pop();
   }
   return segments.join("::");
+}
+
+/**
+ * Strip comments from a `use` declaration before its text is parsed as a path.
+ *
+ * A `use` tree may be spread over several lines with comments between the
+ * leaves, and the text of the AST node carries them. Left in, they become path
+ * segments: `crate::{ // note\n models::User }` yields the specifier
+ * `crate::// note\n    models::User`, which names no module and falls back to
+ * the crate root. A `use` holds no string literals, so removing comment spans
+ * from the raw text is safe.
+ */
+function stripRustComments(text: string): string {
+  return text.replace(/\/\*[\s\S]*?\*\//g, " ").replace(/\/\/[^\n]*/g, " ");
+}
+
+/**
+ * The chain of inline `mod` blocks enclosing a node, outermost first.
+ *
+ * `mod tests { … }` opens a module without opening a file, so a path written
+ * inside it is relative to a position one level below the file. The chain is
+ * what lets that position be reconstructed; it is empty for the overwhelming
+ * majority of declarations, which sit at the top of a file.
+ */
+function rustInlineModules(node: SgNodeLike): string[] {
+  const chain: string[] = [];
+  let current = node.parent();
+  while (current) {
+    if (current.kind() === "mod_item") {
+      const name = current.field("name")?.text();
+      if (name) chain.unshift(stripRawIdent(name));
+    }
+    current = current.parent();
+  }
+  return chain;
+}
+
+/** The subset of the ast-grep node API this module reads. */
+interface SgNodeLike {
+  kind(): string;
+  text(): string;
+  parent(): SgNodeLike | null;
+  prev(): SgNodeLike | null;
+  field(name: string): SgNodeLike | null;
+}
+
+/**
+ * Rewrite a path written inside inline modules so it means the same thing when
+ * read from the file's own position.
+ *
+ * `self` and `super` count module levels, and an inline `mod` is a level that
+ * the file system does not show. `use super::open_store;` inside
+ * `mod tests { … }` in `store/open.rs` names that very file — but counted from
+ * the file it reads as the parent module, and an edge is drawn at
+ * `store.rs`, which the source never imports. That edge also closes a cycle
+ * with the `mod open;` declaration pointing the other way, and `#[cfg(test)]
+ * mod tests` sits in a large share of all Rust files.
+ *
+ * Returns null when the path names the file itself, which is not an edge.
+ * Paths anchored at `crate` are unaffected, and a bare head is left alone:
+ * inside `mod tests`, `use some_crate::Thing;` is how a test reaches another
+ * crate of the project, and rebasing it would lose that edge.
+ */
+function rustPathFromInline(path: string, inline: string[]): string | null {
+  if (inline.length === 0) return path;
+  const segments = path.split("::").filter(Boolean);
+  if (segments.length === 0) return null;
+  if (segments[0] === "crate") return path;
+
+  let climbed = 0;
+  let rest = segments;
+  if (rest[0] === "self") {
+    rest = rest.slice(1);
+  } else {
+    while (rest.length > 0 && rest[0] === "super") {
+      climbed++;
+      rest = rest.slice(1);
+    }
+    if (climbed === 0) return path;
+  }
+
+  // Each `super` first consumes an inline level; only what is left of the
+  // climb reaches the file system.
+  const remainingClimb = Math.max(0, climbed - inline.length);
+  const prefix = inline.slice(0, Math.max(0, inline.length - climbed));
+  if (remainingClimb > 0) {
+    return [...Array(remainingClimb).fill("super"), ...rest].join("::");
+  }
+  const rebased = [...prefix, ...rest];
+  return rebased.length === 0 ? null : ["self", ...rebased].join("::");
+}
+
+/**
+ * The file a `#[path = "…"]` attribute points at, relative to the directory
+ * the declaring file sits in — never to the directory that file's submodules
+ * live in. Both `src/a/b.rs` and `src/a/mod.rs` resolve `#[path = "moved.rs"]`
+ * to `src/a/moved.rs`.
+ *
+ * Attributes are siblings preceding the `mod` in the tree, and several may
+ * stack (`#[cfg(test)]` above `#[path = …]`), so the walk goes back through
+ * all of them.
+ */
+function rustPathAttribute(node: SgNodeLike): string | null {
+  let previous = node.prev();
+  while (previous?.kind() === "attribute_item") {
+    const match = previous.text().match(/^#\[\s*path\s*=\s*"([^"]+)"\s*\]$/);
+    if (match) return match[1];
+    previous = previous.prev();
+  }
+  return null;
 }
 
 /**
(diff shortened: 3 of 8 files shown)


giancarloerra:
The diagnosis is right and the direction is the one this repo has taken four times now, so this lands once two wrong-edge classes are closed. Both are narrow, and both draw an edge Rust itself would not.

## 1. A leading `::` is discarded, which inverts the one unambiguous external marker

`rustUseLeafPath` splits on `::` and then `.filter(Boolean)`, which drops the empty leading segment, so `::log::info` becomes `log::info` before resolution ever sees it. That is the exact token distinguishing the external crate from a local module of the same name.

Ground truth from rustc 1.98.0 on one fixture, toggling only the leading colons. Crate `app` with `pub mod log;` in `src/log.rs`, plus a path dependency also named `log` outside the scanned tree:

```
use ::log::external_marker;   ->  Finished dev profile          (resolves to the EXTERNAL crate)
use log::external_marker;     ->  error[E0432] unresolved import (resolves to the LOCAL module)
```

On that same fixture, the `::` version, this branch emits:

```
nodes=2 edges=2
  src/lib.rs -> src/log.rs      correct, from `pub mod log;`
  src/lib.rs -> src/log.rs      wrong, from `use ::log::external_marker`
```

Fix: carry the leading empty segment through `rustUseLeafPath` and return null for it in `resolveRustImport`. Note this is not covered by the gate proposed below, since `mod log;` genuinely is declared in that file and the `::` is the only distinguishing token.

## 2. The uniform-path fallback is filesystem-based rather than scope-based

The fallback resolves any leading segment that happens to match a sibling `.rs` file, so a third-party head captures the import whenever a same-named file exists. The failure is most visible in `tests/`, where two files are separate integration-test crates and cannot import each other under any path, so the edge is unreachable in Rust regardless of naming.

Fix: gate the fallback on the importing file actually declaring `mod <head>;`, which `extractImports` already gives you, and exclude candidates that are themselves crate roots. rustc agrees that the declaration is the right gate: with `mod log;` present, `use log::external_marker;` fails E0432 rather than reaching the dependency. Worth extending the `resolved === relSourceFile` guard the `super::` branch already has to the `crate::` and uniform-path branches at the same time.

## 3. The regression test named for this cannot fail

`draws no edge into a crate for a third-party path` asserts only that every edge target ends with `.rs`, against a fixture containing no third-party import, in a tree where every graphable file is a `.rs` file. Replacing the entire fallback with "return the first `.rs` file in the set" leaves it green. It needs a fixture with a real external path such as `use serde::Serialize;` beside a same-named local module or target file, asserting those specifiers produce no edge.

## 4. Two claims in the description to correct

"0 edges from third-party paths landing on project files" is refuted by the fixture above. "Every new guard mutation-tested" does not hold either: the structural-prefix guard survives changing `relFile.startsWith(`${moduleDir}/`)` to `relFile.startsWith(moduleDir)` with the full suite green, and it is not an equivalent mutant.

Also worth marking the tree A and tree B figures as measured on a private tree. The direction reproduces at the same order of magnitude on public crates, but the specific numbers cannot be re-run by anyone else.

## Two decisions, not blockers

- **`super::` landing on a crate root resolves to nothing.** `rustModuleFile` tries `<dir>.rs` and `<dir>/mod.rs` only, so a top-level module reaching `lib.rs` via `super::` yields null, which is the most common `super::` target there is. One line if you want it here, or its own PR.
- **Self-edges.** `crate::Item` from the file that defines it is a correct resolution but a noisy file-graph edge, and at scale it makes `codebase_graph_circular` report every Rust file as a cycle. Options are a `source === target` filter at the push site, which changes behavior for TypeScript and Python too, or extending the existing guard to the remaining branches. Say which you prefer.

Two smaller ones for the same round: a comment inside a `use` group corrupts the following leaf, since the group is split on commas without stripping `//` or `/* */`, and the corrupted segments fall back to the crate root; and `self as X` inside a group drops the module edge.

The rest verified clean here: non-Rust resolution is byte-identical across the language matrix with a populated crate map passed on every call, the Python work that just merged is unaffected, omitting the map reproduces the old behavior exactly, and the manifest walk correctly skips `target/` and never touches `fileSet`.

giancarloerra:
The mutation tooling and the outside round are the right response, and most of the delta holds: the declaration gate is correct at edition 2018+ on every case I built, the manifest and edition machinery matches `cargo metadata` everywhere else, the self-edge guard sits where it should with TypeScript and Python untouched, and non-Rust resolution is unchanged across the language matrix.

Two of the new commits regress against `main`, both verified here with cargo as ground truth and an A/B against `f836e99`.

## 1. Edition 2015 turns the gate off entirely, which reopens the original blocker

`0f1f59b` relaxes the gate for 2015, and by its own comment that is every crate whose manifest omits the key. The premise is right, an unanchored 2015 path is absolute from the crate root, but the relaxation dropped the requirement that the head name a module declared *anywhere* in the crate, and it applies the crate-root rebase to `tests/`, `examples/` and `benches/` roots, whose directories hold sibling crates rather than modules.

```
package with no edition key, tests/a.rs = `use b::helper;` beside tests/b.rs

cargo check --offline --tests   ->  error[E0432]: unresolved import `b`
main f836e99                    ->  0 edges                       correct
99f28ae                         ->  tests/a.rs -> tests/b.rs      wrong
same manifest + edition = "2021" ->  0 edges                      correct
```

Two files in `tests/` are separate crates and cannot import each other under any edition, so this is the exact shape the gate exists to stop.

## 2. A declared target `path` written `./src/…` is dropped, and the root is then mis-rooted

The `fileSet` lookup uses the manifest string unnormalized, so a leading `./`, which cargo accepts, never matches. This one is a regression against base rather than a new-but-wrong edge:

```
[[bin]] name = "tool" path = "./src/tools/tool.rs",  src/tools/tool.rs = `mod part;`
src/tools/part.rs real, src/tools/tool/part.rs holding compile_error!

cargo metadata   ->  bin tool = src/tools/tool.rs
cargo build      ->  Finished          (so `mod part;` is src/tools/part.rs)
main f836e99     ->  src/tools/tool.rs -> src/tools/part.rs        correct
99f28ae          ->  src/tools/tool.rs -> src/tools/tool/part.rs   the compile_error! file
```

Same on `[lib] path = "./src/api.rs"`, where the crate is erased entirely (`name: null`, no roots), so sibling crates importing it lose their edges too. No test covers the form: every `path = "` in the suite is a bare relative path. A posix normalize dropping a leading `./` before the three lookups fixes it.

## 3. The leading `::` fix does not cover the brace-group spelling

`use ::log::info;` is fixed. `use ::{log::info};` is not: the group prefix is stripped to the empty string before the marker reaches the resolver, so the global path resolves into the local module again. Probing the tree expander: `::corelib::marker`, `::corelib::{marker}` and `::corelib::{self, marker}` keep the marker; `::{corelib::marker}` and `::{corelib::marker, log::Level}` lose it.

## 4. Two smaller wrong edges, same family

- **A file-level `mod x;` is treated as in scope inside an inline `mod` block.** rustc says otherwise: with the dependency removed, the block gets `E0432 ... help: a similar path exists: use super::corelib::marker`. This is the mirror of the case you already handle correctly, a declaration inside a block not counting at file level.
- **`#[path]` rebinds the module name, but an orphan file at the module's own name still matches by convention**, so it takes an edge rustc would send to the dependency.

## On the shape of this

Three rounds in, each fix has been correct and each has surfaced another case, which is what modelling rustc name resolution from the filesystem gets you: the tail is long, and cargo is the only oracle. It may be worth deciding the default explicitly rather than case by case, resolving only what a declaration proves, and letting anything unproven stay unresolved. Fewer edges, none of them wrong, and the remaining tail becomes documentation instead of another round? Either way, the four above need closing first.

giancarloerra:
All three findings from the last round are closed, verified here: the 2015 gate draws nothing on the `tests/` fixture, the `./`-prefixed target path resolves to `src/tools/part.rs`, and the brace-group `::` no longer reaches the local module.

Two blockers in the new round, and a request about its shape.

## 1. The environment scan emits an unanchored pattern, which deletes source in any language

`ignore.ts` writes `` ig.add(`${escapeIgnorePattern(relDir)}/`) `` for a detected environment. A directory directly under the root has no slash in its relative path, so gitignore matches that name at **every** depth. A root-level conda environment therefore deletes same-named source elsewhere in the tree:

```
fixture: root conda env at toolbox/ (conda-meta/), plus real TypeScript at packages/app/toolbox/

main   kept    packages/app/toolbox/helper.ts   kept    packages/app/toolbox/index.ts
head   IGNORED packages/app/toolbox/helper.ts   IGNORED packages/app/toolbox/index.ts
```

The same mechanism undoes this PR's own motivating case: with a root-level `env/` or `venv/` present, `crates/clap_complete/src/env/mod.rs` is excluded again, exactly as the code comment says it should no longer be. Against the installed `ignore` package: `env/` matches `crates/clap_complete/src/env/mod.rs`, `/env/` does not, and both still catch `env/lib/dep.py` at the root.

One character fixes it, `` `/${escapeIgnorePattern(relDir)}/` ``. Worth noting no test pins the current behaviour, because every new fixture sits two levels deep where the pattern is anchored by accident; a root-level fixture is what would have caught it.

Related, same file: the walk still `continue`s on a directory named `venv`, which defeats the marker check placed above it. A real virtualenv at `crates/venv/backend/env/` has its installed libraries indexed, and a nested `.gitignore` under `crates/venv/` stops being read while `RESPECT_GITIGNORE` is true.

## 2. A macro body inside an inline `mod` draws an edge rustc never compiles

`cfg_if!` in its if/else form is not valid Rust once the head and braces are blanked, so tree-sitter recovery closes the inline `mod` after the first arm and re-parents the rest to file level. On a cargo-verified fixture the branch draws `src/lib.rs -> src/arm_b.rs`, a file carrying `compile_error!` that a successful `cargo build` proves is never compiled, while the arms rustc does read get no edge. Main drew nothing there, so both edges are new. Six crates in a 551-crate registry sample carry the shape, `cfg-if` and `js-sys` among them.

A guard that accepts the unwrapped pass only when the blanked region reparses without ERROR nodes would close it.

## 3. Please split the ignore change out

`src/services/ignore.ts` is shared by every language and both walks, it is the only part of this PR with a proven cross-language regression, and it is independently useful: the anchoring bug you found is real and worth shipping on its own. As its own PR it is small enough to review in one pass and to revert alone if it misbehaves, and the Rust work then stands on its own merits rather than being gated behind it. This round alone is +1878/-90 across 8 files, and the PR is +4667/-52 over 25 commits.

## Minor

`graph-imports.ts` contains a literal NUL byte where an escape was meant (`join("<NUL>")` around line 679). It makes `grep` and `rg` treat the file as binary and skip it silently, while git still diffs it as text, so it is invisible in review. `"\0"` fixes it.

giancarloerra:
Hi let me know if you intend to continue working on this so we can merge, thanks! :-)

giancarloerra:
If I don't receive a reply within today 2nd September I'll take over and finish this as it's becoming more and more difficult to rebase it considering the amount of work going on with parallel PRs, not a problem of course. I just want to proceed quickly and swiftly with closing PRs within few hours :-)

matteodimattia:
Yes, continuing today. Both blockers from your last round are closed locally with cargo-verified fixtures; I am rebasing on current main and splitting the `ignore.ts` change into its own PR as you asked. Pushing within a few hours.

SIRTHEO:
Rebased on current `main` (`cacbac4`), and both blockers from your last round are closed. Split as you asked: the `ignore.ts` change is #136, and this branch sits on top of it.

## Your three items

**1 · The unanchored environment pattern** — `/${relDir}/`, as you wrote, plus the `venv` name removed from the walk's skip list so a nested marker is reached (it is still skipped directly under the root, where `/venv` has already excluded it — the review pointed out the walk was otherwise reading an old virtualenv's `site-packages` for nothing). A root-level fixture pins it: with a conda env at `toolbox/`, `packages/app/toolbox/helper.ts` stays. In the separate PR.

**2 · A macro body inside an inline `mod`** — reproduced on a cargo-verified crate: the branch drew `src/lib.rs -> src/arm_b.rs` twice and `src/lib.rs -> src/after.rs`, three edges into `compile_error!` files that a clean build proves are never compiled, while `mod after;` — which does belong to the block — lost its edge.

I drew the guard slightly differently from what you proposed, and measured why. Reparsing the blanked region on its own gives the *same* single ERROR (on the attribute) whether the macro sits at file level or inside a block — what differs is not the error, it is how far recovery reaches. At file level nothing moves and the arms are read where rustc reads them; inside a block the `mod` is closed after the first arm. So the rewritten source is parsed once and kept only while **every ERROR sits inside a macro this pass unwrapped**, or is one the previous pass already accepted, start and end; otherwise it is retried without the unparsable bodies written inside a block, then without any, and dropped if recovery still reaches out.

Why an early close cannot be silent is structural rather than `cfg_if`-specific: a macro body's braces are balanced, so blanking the head and the outer pair leaves the block's real `}` orphaned once recovery closes it early, and an orphan `}` is an ERROR outside every region. That is what the guard sees. It does not see everything — `m! { #[path = "x.rs"] }` followed by `mod after;` merges the attribute onto the next item with no ERROR at all. That is a synthetic shape, and I mention it because the ERROR test is a proxy, not a proof.

A first cut of this forgave, on the second pass, anything inside a macro the *first* pass had unwrapped. A pre-push review caught that this reopens your case through a side door: `outer! { mod k { cfg_if! { … } mod z; } }` is clean on pass one, and on pass two the `cfg_if!` closes `mod k` inside `outer!`'s region, so the orphan `}` was forgiven and `mod z;` drawn at file level again. What a pass hands on is now the exact ERRORs its tree had, nothing else; the case is a test. Two more from the same review are in the same commit: a `use` the second pass re-reads with more in view now replaces the first pass's reading instead of standing beside it (the first reading, a bare head in a block, could reach a workspace crate rustc refuses with E0432), and the fallbacks are computed on the first refusal rather than before the first attempt.

What the guard gives up is worth naming plainly: **every `cfg_if!` arm inside a block stays unread** — the oracle test says so itself, `src/inner/arm_a.rs` is the one file rustc compiles that the graph is not expected to reach — and a macro written *inside* an arm is not reached either, since after blanking the arm is the block of a recovered `if`. libc nests `cfg_if!` in arms; those read only their own text. Both are limits of reading a body in place.

Measured over the 153 crates of a 1,256-crate registry cache that carry the shape, distinct dependencies against the previous head:

| guard | lost | gained |
|---|---|---|
| reparse the body alone, refuse on ERROR | 78 | 3 |
| this one | **56** | 3 |

The 22 are file-level `cfg_if!` bodies in files that also hold one inside an `impl` — `ahash`, `dashmap` — which the simpler guard gives up whole. The 56 that stay lost are the arms inside blocks, plus the files blanking breaks outright (libc's `freebsd/mod.rs` parses as one ERROR from line 1), where no reading of the tree would be worth trusting. ripgrep, tokio and clap come out **identical, edge for edge**.

The blind spot you might look for — recovery that produces a MISSING node and no ERROR — I measured rather than argued: of 36,163 `.rs` files in the cache (it had grown from the 34,655 the code comment counts), 67 have a declaration outside the macro change position after blanking, and **all 67 carry an ERROR**.

**3 · Split** — done: #136 carries `ignore.ts` and its test, nothing else. This branch is rebased on top of it, so the four `fix(ignore)` commits you see at the bottom here are that PR's, and disappear from this diff once it merges — GitHub only lets me target a branch of your repo as base, so I could not point #118 at it directly. The NUL byte fix (your minor) is its own commit now; it used to share one with the anchoring.

## Cost, corrected

The +48% I reported was two unpaired measurements. Nine runs a side, three columns alternating in the same batch, tokio:

| | median |
|---|---|
| `main` | 1,304 ms |
| this branch before the guard | 2,013 ms (+54%) |
| with the guard | 2,123 ms (+62%) |

Re-measured today against `cacbac4`, same method: 1,286 ms → 2,096 ms (+63%). So the guard is ~5% of the pre-guard build; the rest was already there. The tree built to accept the rewritten source is handed to the recursive pass instead of parsing the same text twice — without that the guard alone cost 15% of extraction.

## Against `main` at `cacbac4`

Edges on these trees are what they were at `f836e99` — 76 / 167 / 77 on `main`, 612 / 3,922 / 1,196 here — and the only edges `main` draws that this branch does not are the four that pointed at a sibling where rustc reads the child. Python, TypeScript and JavaScript trees are identical edge for edge; a Go tree too.

Verified locally before pushing: `tsc`, lint, the full suite with Ollama (1,781, the cargo-oracle test included), a mutation pass over the new decisions — one of them (`insideBlock`) was held by nothing, and has a test now — and, for the review fixes above, an edge-for-edge A/B over every crate of the 271-crate registry cache on this machine: no crate moves, while the constructed case does.

One thing I did not do, and would rather name than hide: reading each macro body in isolation and attributing it the invocation's scope would make the guard unnecessary by construction — recovery could not touch the rest of the file, and both limits above go with it. A prototype recovered the 56 and ran 23% faster on tokio, but it rewrites five tests and I did not want to land that unreviewed in the same round. Happy to open it as a follow-up.

Two things at your call, since they touch how you merge: the branch is 32 commits, and I can squash it into four logical ones (crate map; path and scope rules; macro bodies and the guard; oracle and CI) if that is easier to land — say so and it is done within the hour. And the CI job relies on the runner image shipping a Rust toolchain; a `dtolnay/rust-toolchain@stable` step would make it independent of the image, if you prefer that to a comment saying it is deliberate.

giancarloerra:
Thanks for the detailed update. The previous requested split and macro-recovery corrections are now present, and CI passes on current head `f31a76a`.

Two functional blockers remain.

### 1. Undeclared workspace crates can still become dependencies

`crateNamed()` translates a declared dependency alias when one exists, but otherwise searches every workspace crate by the raw import name. This can resolve `use helper::Thing` to a sibling workspace crate named `helper` even when the importing Cargo package does not declare that dependency and Cargo rejects the import.

Current code:

https://github.com/giancarloerra/SocratiCode/blob/f31a76a55570085439c28c60f541c9f4d001e922/src/services/graph-resolution.ts#L1893-L1915

Required correction:

- Restrict cross-package lookup to dependency names or aliases declared by the importing package.
- Preserve resolution to the importing package's own library target where another target in that package imports it.
- Add a Cargo-verified workspace regression where an undeclared sibling crate shares the import head and must not become an edge.

### 2. Rust 2015 leading-`::` paths still use the wrong declaration scope

For Rust 2015, a path beginning with `::` is crate-root-relative. The current resolver can consult the importing child file's `declaredFile` mapping first, allowing a child-local `#[path] mod foo` to capture `use ::foo::Item` instead of resolving the crate root's `foo`.

Relevant code:

- https://github.com/giancarloerra/SocratiCode/blob/f31a76a55570085439c28c60f541c9f4d001e922/src/services/graph-resolution.ts#L1946-L1970
- https://github.com/giancarloerra/SocratiCode/blob/f31a76a55570085439c28c60f541c9f4d001e922/src/services/graph-resolution.ts#L2088-L2095
- https://github.com/giancarloerra/SocratiCode/pull/118#discussion_r3873268489

Required correction:

- When a Rust 2015 path starts with `::`, do not use a child source file's declaration map.
- Resolve through the applicable Cargo target root, or leave it unresolved if that root mapping is unavailable.
- Add the nested-child versus crate-root regression from the review finding.

The documented false negatives around some Rust 2015 root-relative edges and platform-selected modules inside inline macro blocks are not regressions from `main`. They can remain documented limitations for this PR rather than expanding it into a larger module-tree and isolated-macro-parser redesign.

Please fix #136 first. Once it is merged, rebase this branch so the #136 commits disappear from this diff, apply the two corrections above, and rerun CI and CodeRabbit against the final rebased head. There is no need to squash the branch or change the Rust toolchain setup before those functional items are complete.

SIRTHEO:
Status, so this does not look idle: #136 is at `bc0360e` with every item closed and CI green, and I am holding this branch until it merges, as you asked. Both corrections here are done locally and cargo-verified — the fence on declared dependencies (admitted under the name rustc knows the crate by, so a `[lib] name` dependency keeps its edge, and a package with an unparseable manifest is not fenced), and the 2015 leading `::` answered from the crate root first (module, then declared crate, then root item — a registry crate resolves to nothing). Each has a unit fixture and a crate in the cargo-oracle file; ripgrep, tokio and clap are identical edge for edge, and across a 271-crate registry cache only one edition-2015 crate moves, gaining three true edges. The moment #136 lands I rebase on `main`, push, and post the details.

giancarloerra:
> Status, so this does not look idle: #136 is at `bc0360e` with every item closed and CI green, and I am holding this branch until it merges, as you asked. Both corrections here are done locally and cargo-verified — the fence on declared dependencies (admitted under the name rustc knows the crate by, so a `[lib] name` dependency keeps its edge, and a package with an unparseable manifest is not fenced), and the 2015 leading `::` answered from the crate root first (module, then declared crate, then root item — a registry crate resolves to nothing). Each has a unit fixture and a crate in the cargo-oracle file; ripgrep, tokio and clap are identical edge for edge, and across a 271-crate registry cache only one edition-2015 crate moves, gaining three true edges. The moment #136 lands I rebase on `main`, push, and post the details.

#136 has now been merged into `main`.

Please rebase this branch onto the current `main`, include the two corrections you confirmed are already completed locally, and push the updated branch:

- Restrict cross-package Rust resolution to dependencies or aliases declared by the importing package, while preserving access to that package's own library target.
- Resolve Rust 2015 leading-`::` paths from the applicable Cargo target root without allowing a child file's declaration map to capture them.

Once the rebased branch is pushed, I will verify CI and CodeRabbit against the final head. There is no need to squash the commits or change the Rust toolchain setup.

SIRTHEO:
Rebased on `main` with #136 in, so its commits are out of this diff; both corrections are on `8bbf02a`, cargo-verified, in four commits — one each, and two more with what pre-push reviews found in them (a `[lib] name` regression the fence introduced, a fallback that made the 2015 branch's crate lookup unreachable, and the manifest-ranking defect in §3 below), each named in its message so the history says what happened rather than hiding it in a clean diff.

**1 · Undeclared workspace crates.** `crateNamed` translated a declared alias where one existed and otherwise searched every crate of the project by the raw name. The fence is now what the importing package declares, read from `[dependencies]`, `[dev-dependencies]`, `[build-dependencies]` and every `[target.*]` table, and it admits a name three ways — because the name the manifest declares and the name the code writes are not always the same: a renamed alias (what cargo hands rustc); the *library* name of a project crate declared under its own package name (`other = { path = "../other" }` with `[lib] name = "otherlib"` is imported as `otherlib`, and `use other::…` is E0432); and a plainly declared package no project crate answers to, which comes from outside. The package's own library stays reachable by its own name, which is how its binaries, tests and examples import it; a file outside every manifest has no package to ask, and a package whose manifest cannot be parsed is not fenced at all — the alternative would cost it every cross-crate edge, against the crate map's own promise.

The regression is a workspace: `app` declares `util` and `other` and not `helper`, and writes `use helper::Thing` under `#[cfg(any())]` — rustc strips it before name resolution, so the crate builds, while `helper` carries a `compile_error!` that would end the build if cargo ever compiled it for `app`. The graph reads the line all the same and now draws nothing from it; `app -> util`, `app -> other` (imported as `otherlib`) and the binary's `use app::all` still resolve, and rustc's dep-info lists exactly those. Four unit fixtures that leaned on an undeclared sibling now declare it, as rustc would have required.

One correction to the record: the PR body once listed this as a known limit, and my update of 2 September removed the line on the strength of the alias work. That was wrong — the alias path only ran where an alias existed — and this commit is what closes it. The body says so now.

**2 · Rust 2015 leading `::`.** Answered first, and from the root only, in the order a name at the root can be three things: a module of the root, found by walking from its directory with no fallback; a crate the package declares (in 2015 `extern crate foo;` at the root puts it there, and a root declaring both is E0260) — a registry crate resolves to nothing; and otherwise an item the root defines or re-exports, which is the root file, unless the name is one rustc provides on its own (`std`, `core`, `alloc`, `proc_macro`, `test`). The importing file's declaration map is not consulted at all, and a 2015 file no target root covers is left unresolved. The regression is the one from `r3873268489`, run through cargo: an edition-2015 crate whose child declares `#[path = "local.rs"] mod foo;` defining `Other` and writes `use ::foo::Item` — which compiles only because `::foo` is the root's module. The graph draws `child -> src/foo.rs` beside the declaration's own edge to `local.rs`; before, only the second existed. The unit fixture also holds the declared sibling, the registry crate, the single-segment `::foo`, an integration test reaching its library by crate name, and the file outside every target, side by side.

Both leave ripgrep, tokio and clap identical edge for edge, and over a 271-crate registry cache on this machine — where edition 2015 is common — exactly one crate moves: brotli-decompressor (no `edition` key, so 2015) gains three edges and loses none — `use ::dictionary::{…}` in `decode.rs` now reaches `dictionary/mod.rs`, and `use ::BrotliResult;` / `::HuffmanCode` in `ffi/` reach the root that re-exports them; its `use ::alloc;` and `use ::core;` draw nothing, as they should. Full suite, `tsc`, lint green; the cargo-oracle file now holds three crates (the original, the workspace, the 2015 one) and runs in the same CI job.

Re-verified on the rebased head, since it now carries your two watcher commits: `tsc`, lint, the full suite (1,818), the cargo-oracle file (14), the three trees edge for edge against the commit below these two, and the unit suite in `node:18`, `node:20` and `node:22` containers — the last CI failure here was a harness one, so Linux is checked directly rather than inferred from macOS. The four mutants behind the environment-reversal test are each red again under your `watcherActive` change.

**3 · A one-letter member directory, found reviewing the two above.** The manifest that governs a file is the innermost one containing it, and the crates were ranked by the length of their directory. The root's is `"."` — the empty prefix, one character long — so it tied with any member directory of one character, and the sort being stable over manifests walked shallowest first, the root won. Every file of `a/` was then governed by the wrong manifest, which is what makes this a defect of §1 and §2 rather than an old one: a `[workspace]`-only root declares no dependency, so the new fence turned away every crate `a` declares, and it declares no edition either — defaulting to 2015, where the new `::` branch counts from the crate root, so `use ::bb::Thing;` in a 2021 file drew an edge into `a`'s own `bb` module. Renaming the directory `aa/` made both go away; nothing about the code had changed, only the length of a name. Ranked by depth with `"."` at zero, the three trees and all 271 crates of the registry cache come out identical edge for edge. The regression is a workspace with a member in `a/` that depends on `bb` by path and declares a `bb` module of its own defining no `Thing`, so cargo's clean build is what proves `::bb` names the dependency. It also holds a second package nested at `a/sub`, because depth decides two orderings and not one — root against member, and a member against the one nested inside it — and a first cut of the test pinned only the first: answering "0 for the root, 1 for everything else" kept all of it green while `a` governed `a/sub`, where `use cc::Deep;` is E0432. Four mutants now: raw length (red on both halves, which is why the assertions sit in separate tests), flat depth, reversed order, and a correct equivalent that stays green.

One thing of the same shape I am leaving alone: `crateNamed` a few lines below ranks crates the same way and still treats `"."` as covering everything. I could not build a tree where it draws a wrong edge and cargo accepts the tree — two packages of one name collide in the lockfile — so there is nothing to prove a change against, and I would rather report it than fix it blind.

Three things the same review found that I have documented rather than changed, since each would reach past what you asked for — say the word on any of them and I will send it separately. A Rust `use` tree pushes one edge record per leaf, so `use crate::io::{A, B, C};` is three records to one file: 612 records for 261 distinct pairs on ripgrep. Deduplicating at `edges.push` would move every language's numbers, so the body now carries a distinct-pair column beside the record counts and names what reads records as connections (`graph_stats`, `avgDependencies`, `mostConnected`, `findCircularDependencies`, the HTML view). Two `#[path]` declarations of one name under opposite `cfg`s both draw their edge, but a path *through* the name resolves to whichever was written last, on every platform. And the unanchored 2015 path is now in the known limits where it belongs: the body had it listed among the orderings that follow rustc, which stopped being true when the declaration gate was turned on for every edition — a comment in `graph-resolution.ts` still said the opposite of the code twenty lines from the comment recording the decision, and that is gone too.

The limits you named stay documented as limits, and the branch is neither squashed nor changed in its toolchain setup, per your note.

SIRTHEO:
@coderabbitai review

SIRTHEO:
All three CodeRabbit comments were real, and are fixed on the head above. Each was checked before acting on it, and one of them is the good kind of finding.

**The unanchored-path pair asserted `null` on both sides.** The case named for edition 2018 held no `src/client/registry.rs`, so `null` was the answer for the same reason the 2015 case gives it — a file that is not there. It passed with the module-directory walk removed, and it would have passed with the edition dropped altogether: a test that could only ever agree with its twin. It now holds both candidates, the crate root's `registry` and the module's own, and asserts the module's; returning the file's directory instead of its module directory answers `src/registry.rs` and turns it red.

I took the name with it. `resolveRustImport` stopped telling the editions apart at this point when the declaration gate was turned on for every edition, and the 2015 case beside it is the same walk with nothing to find — which is why that one is named a declared limit. The name now says what is asserted rather than what the resolver used to do: the head is read from the file's own module directory, never from the crate root.

**Two fixtures spoke the platform separator** where everything around them speaks forward slashes. `sourcesRustcRead` compared `path.relative` output against literals like `src/lib.rs` and guarded on `target/`, so on Windows every lookup misses and the build directory is admitted instead of skipped. And the test that has to build an absolute `[lib] path` at run time interpolated it into a TOML basic string, where a backslash opens an escape — `\U` out of a drive path is one `smol-toml` rejects, so the manifest would be dropped and with it the edge that test exists to assert. Both normalised. Neither is reachable from CI, which is why they were only ever going to be found by reading.

Unchanged on this platform: `tsc`, lint, the full suite (1,818), the cargo-oracle file (14), the unit suite in `node:18` and `node:22` containers, and ripgrep, tokio and clap identical edge for edge.

giancarloerra:
@coderabbitai review

giancarloerra:
Thank you very much, merging now!

PR #124, opened 2026-08-31
Stop concurrent callers from racing to create the metadata collection

yusuke-omichi-ctc:
## Summary

`ensureMetadataCollection()` is awaited from more than ten exported functions, several of which run concurrently while a project is being opened. The guard flag is only set after creation finishes, so every concurrent caller sees "does not exist" and calls `createCollection`. All callers but the first get a 409 Conflict, which propagates as `loadProjectHashes(collName=...) failed [status 409]: Conflict` and aborts indexing. It reproduces against an external Qdrant, where the creation round trip is long enough for the callers to overlap.

## Changes

- Share the in-flight creation promise so concurrent callers join it instead of starting a second creation. The promise is cleared in a `finally` block so a transient Qdrant failure can be retried rather than cached forever. Both that clear and the readiness flag are guarded by promise identity, so a `resetMetadataCollectionCache()` landing mid-creation is neither undone by the finishing attempt nor able to have its replacement promise overwritten.
- Treat "already exists" (409, or a message saying so) as success for `createCollection`. A second MCP instance pointed at the same Qdrant can legitimately win the race, and the end state is the one we wanted. Other failures still propagate.
- Ensure the payload index through the existing idempotent `ensurePayloadIndex()` helper rather than a second inline `createPayloadIndex`, and call it on every attempt instead of only on the creation path. An attempt that created the collection and then failed before indexing used to leave the collection permanently unindexed, because every later attempt saw the collection and skipped the index along with the creation.

Note that the shared helper ignores all index errors, not only conflicts. Narrowing that is left alone here, since it applies equally to its existing caller.

<details>

### 何を直すか

メタデータ用のコレクションを複数の処理が同時に作ろうとして、インデックス構築が失敗するバグを直します。

### なぜ必要か

`ensureMetadataCollection()` は 10 個以上の関数から呼ばれ、プロジェクトを開くときに複数が同時に走ります（準備確認・ハッシュ読み込み・グラフ読み込みなど）。作成済みを示すフラグは作成が終わってから立つため、同時に走った呼び出しはすべて「まだ無い」と判断し、すべてが作成を試みます。最初の 1 つ以外は 409 Conflict を受け取り、インデックス構築が止まります。

外部の Qdrant を使うと再現します。作成の往復に時間がかかり、呼び出しが重なるためです。

### 変更の内容

作成中の Promise を共有し、同時に来た呼び出しは 2 度目の作成を始めず、進行中のものに合流します。`finally` で解放するため、一時的な失敗はやり直せます。解放と準備完了フラグの設定は、どちらも「自分が入れた Promise か」の判定の下に置いています。

「既にある」を成功として扱います。同じ Qdrant を指す別のインスタンスが競争に勝つことは正当で、結果として望んだ状態になるためです。

payload index の作成は、既存の冪等なヘルパを使い、毎回呼ぶ形にしました。以前は作成経路でしか呼ばれなかったため、コレクションを作った直後に index 作成が失敗すると、以降そのコレクションは恒久的に index を欠いていました。

</details>

## Type of change

- [ ] Documentation update

## Testing

- [x] Unit tests pass (`npm run test:unit`) — 56 files, 1191 tests
- [ ] Integration tests pass (`npm run test:integration`) — not run; requires Docker

Tests cover the concurrent case, all three shapes of "already exists", retry after a transient failure, non-conflict failures of both calls, a cache reset landing mid-creation, and the index being ensured when creation is skipped. Reverting `src/services/qdrant.ts` to `main` fails 8 of the 10 tests, so they exercise the implementation rather than a copy of it.

## Checklist

- [x] I have updated documentation (README.md / DEVELOPER.md) if needed — no public API change, so no doc update

## Related issues

None. Happy to move the discussion to an issue first if you prefer that for changes of this size.

---

This is one of five independent branches I am offering. The other four are feature additions around the embedding configuration; this one is the only bug fix, so I am opening it first. `src/services/ensureCollection()` and the private twin in `src/services/symbol-graph-store.ts` still have the same check-then-create shape — I kept this PR to the metadata collection to stay on one logical change, and I am glad to follow up on those separately if you want them fixed the same way.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when multiple operations initialize metadata storage simultaneously.
  * Prevented already-existing collection conflicts from causing failures.
  * Ensured required metadata indexes are restored after interrupted setup.
  * Improved recovery after temporary setup failures and cache resets.
  * Added build-version information to saved graph metadata.

* **Tests**
  * Added coverage for concurrent initialization, conflict handling, retries, resets, and failure scenarios.

### DIFF
--- a/src/services/qdrant.ts
+++ b/src/services/qdrant.ts
@@ -158,16 +158,45 @@ export async function ensureCollection(name: string): Promise<void> {
   }
 }
 
-/** Create a payload index on a collection (idempotent — ignores "already exists" errors) */
-export async function ensurePayloadIndex(collName: string, fieldName: string): Promise<void> {
+/** True when an error means "someone else already created it" — safe to ignore. */
+function isAlreadyExistsError(err: unknown): boolean {
+  const status =
+    (err as { status?: number })?.status ?? (err as { statusCode?: number })?.statusCode;
+  const message = err instanceof Error ? err.message : String(err);
+  return status === 409 || /already exists/i.test(message);
+}
+
+/**
+ * Create a payload index, ignoring only the "it is already there" conflict.
+ *
+ * Every other failure propagates: a 503 leaves the collection without the index,
+ * and a caller that swallows it can never tell that it still has work to do.
+ */
+async function createPayloadIndexIfMissing(collName: string, fieldName: string): Promise<void> {
   const qdrant = getClient();
   try {
     await qdrant.createPayloadIndex(collName, {
       field_name: fieldName,
       field_schema: "keyword",
     });
-  } catch {
-    // Index already exists — ignore
+  } catch (err) {
+    if (!isAlreadyExistsError(err)) throw err;
+    // Index already exists — that is the end state we wanted.
+  }
+}
+
+/** Create a payload index on a collection (best effort — never throws) */
+export async function ensurePayloadIndex(collName: string, fieldName: string): Promise<void> {
+  try {
+    await createPayloadIndexIfMissing(collName, fieldName);
+  } catch (err) {
+    // Callers of this helper index their payload opportunistically: a missing
+    // index costs filter performance but does not make their write incorrect.
+    logger.debug("ensurePayloadIndex failed (ignored)", {
+      collection: collName,
+      field: fieldName,
+      error: err instanceof Error ? err.message : String(err),
+    });
   }
 }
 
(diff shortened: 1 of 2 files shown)


giancarloerra:
One blocker remains before merge.

[`ensurePayloadIndex()`](https://github.com/giancarloerra/SocratiCode/blob/d40ea4b39ad7fbb7cf1933f7e042f139dad3bfb4/src/services/qdrant.ts#L161-L171) catches every `createPayloadIndex()` failure. [`ensureMetadataCollection()`](https://github.com/giancarloerra/SocratiCode/blob/d40ea4b39ad7fbb7cf1933f7e042f139dad3bfb4/src/services/qdrant.ts#L745-L788) therefore treats a 503 as success, marks the collection ready, and never retries the missing index in that process.

Please:

- Ignore only the already-exists conflict and propagate every other payload-index failure.
- Replace the current [503 “keeps going” test](https://github.com/giancarloerra/SocratiCode/blob/d40ea4b39ad7fbb7cf1933f7e042f139dad3bfb4/tests/unit/qdrant-metadata-collection-race.test.ts#L154-L166) with a regression proving the first call rejects and a later call retries the payload index successfully.
- Remove the hardcoded `socraticode_metadata` assumption from the [existing-collection test](https://github.com/giancarloerra/SocratiCode/blob/d40ea4b39ad7fbb7cf1933f7e042f139dad3bfb4/tests/unit/qdrant-metadata-collection-race.test.ts#L209-L224). Either isolate `QDRANT_COLLECTION_PREFIX` to an empty value before module import or derive the expected prefixed collection name.

yusuke-omichi-ctc:
All three addressed in `646c15c`, rebased onto current `main`.

**1. Ignore only the already-exists conflict.** You are right, and it undoes a trade-off I had accepted too easily: I moved to `ensurePayloadIndex()` in an earlier round for reuse and noted in the PR body that it swallows every failure. Treating a 503 as success and then marking the collection ready is exactly the hole you describe — the metadata collection is queried by `collectionName`, so an attempt that could not create that index has not made the collection usable.

Rather than change `ensurePayloadIndex()`, which `src/services/context-artifacts.ts` also calls where best-effort is the right contract, I added a private `createPayloadIndexIfMissing()`: it ignores only `isAlreadyExistsError()` and propagates everything else. `ensurePayloadIndex()` now delegates to it and keeps its own catch, so its other caller is unaffected. `createPayloadIndex` is still called from one place.

**2. Regression instead of the 503 "keeps going" test.** Replaced with `rejects when the payload index fails, then retries the index next call`: `createPayloadIndex` fails 503 on the first attempt and succeeds afterwards, and `getCollections` reports the collection as existing from the second call on. It asserts the first `loadProjectHashes` rejects, the second resolves, `createPayloadIndex` ran twice, and `createCollection` ran once.

**3. Hardcoded collection name.** Confirmed: `QDRANT_COLLECTION_PREFIX=myprefix_ npx vitest run tests/unit/qdrant-metadata-collection-race.test.ts` failed 1 of 10 before this change and passes 10 of 10 after. The tests now derive the name the way the module does — reading `QDRANT_COLLECTION_PREFIX` from `src/constants.js` after `vi.resetModules()`, the same approach `tests/integration/tools.test.ts` already uses.

Verification: `npm run lint` clean, `npx tsc --noEmit` clean, `npm run test:unit` 60 files / 1268 tests. Reverting `src/services/qdrant.ts` to `main` fails 8 of the 10 tests in that file.

The commit message is updated to describe the narrowed index handling rather than the previous "keeps its best-effort contract" wording.

yusuke-omichi-ctc:
One thing worth naming before you merge, so it is not a surprise later.

`ensureCollection()` has the same shape of bug this PR fixes, and I deliberately left it alone. It reads `getCollections()`, checks membership, and only then calls `createCollection()` followed by four bare `createPayloadIndex()` calls — check-then-create with no conflict handling, on the per-project collections rather than the metadata one.

The two helpers this PR adds, `isAlreadyExistsError()` and `createPayloadIndexIfMissing()`, are exactly what that fix would use. I kept them private and this PR scoped to `ensureMetadataCollection()` because widening it would change the failure behaviour of every indexing run, and because the reproduction and the regression tests I have are for the metadata path.

Say the word and I will open it as a separate issue, or as a follow-up PR once this one lands.

giancarloerra:
Yes, please open a follow-up PR please (merging 124 now).

Please cover both remaining collection-initialisation paths:

- `src/services/qdrant.ts::ensureCollection()`
- `src/services/symbol-graph-store.ts::ensureCollection()`

Required behaviour:

- Concurrent callers for the same collection share one in-flight operation.
- An already-exists conflict from another process is treated as success.
- Every other collection-creation failure propagates and remains retryable.
- The per-project path ensures all four required payload indexes even when the collection already exists.
- Only already-exists payload-index conflicts are ignored. Transient and unexpected failures must propagate and remain retryable.
- No existing index is deleted, changed, or made dependent on reindexing.

Please add regressions for same-process concurrency, an external 409 winner, transient failure followed by retry, and an existing collection missing one or more payload indexes.

PR #125, opened 2026-08-31
Accept OpenAI-compatible servers without /v1/models

yusuke-omichi-ctc:
## Summary

The `lmstudio` provider gates readiness and health on `models.list()`. Servers that host exactly one model — HuggingFace Text Embeddings Inference (TEI) being the common case — fix the model at startup and have no `/v1/models` route, so the listing returns 404 even though `/v1/embeddings` works. `ensureReady()` then fails with an "LM Studio is not reachable" message and the server is unusable.

## Changes

- Add `LMSTUDIO_ALLOW_MISSING_MODEL_LISTING` (default `false`, so existing behaviour is unchanged). The name carries the `LMSTUDIO_` prefix because the flag only affects that provider, alongside `LMSTUDIO_URL` and `LMSTUDIO_API_KEY`. It is resolved in `embedding-config.ts` on every `getEmbeddingConfig()` call.
- When the flag is on, **only a 404 or 405** from the listing falls back to probing `/v1/embeddings` with a single throwaway input. A refused connection, a 401, or a 5xx keeps the LM Studio diagnostics, which are what an operator who forgot to start the Local Server needs to read. `healthCheck()` applies the same rule.
- Without `/v1/models` the probe is also the only chance to validate `EMBEDDING_DIMENSIONS`, which sizes the Qdrant collection. Compare the probe's vector width against the configured value and fail fast on a mismatch, naming both numbers, rather than letting it resurface as an opaque database error.
- Accept `true` / `1` / `yes` and `false` / `0` / `no`, case-insensitively and ignoring surrounding whitespace. A non-empty value that is none of those throws, naming the variable and the value: a typo such as `ture` is a request to turn the flag on, and reading it as "off" would send the operator to debug the missing-listing failure they believed they had opted out of. Unset, empty, and whitespace-only still mean "not configured" and keep the default.
- Document the flag in the README's LM Studio section, with a TEI env example.

A model that is absent from a working listing still fails, so the flag only relaxes the missing-endpoint case. Nothing outside the `lmstudio` provider changes; `provider-litellm.ts` is untouched.

<details>
<summary>日本語の説明 (Japanese explanation)</summary>

### 何を直すか

モデル一覧の API を持たない OpenAI 互換の推論サーバを使えるようにします。

### なぜ必要か

[...]

PR #127, opened 2026-08-31
Make the embedded file path optional

yusuke-omichi-ctc:
## Summary

`prepareDocumentText()` always puts the relative path in front of the chunk content, and that string is what reaches the vector store: it feeds the dense embedding and the BM25 sparse vector built from the same text. Path tokens help path-shaped queries, but on a prose-heavy corpus — design notes, specifications, anything where the filename says little — they are noise in every chunk, and there is no way to turn them off.

## Changes

- Add `EMBEDDING_DOCUMENT_INCLUDE_PATH` (default `true`, so existing behaviour and existing indexes are unchanged).
- Accepts `true` / `1` / `yes` and `false` / `0` / `no`, case-insensitively and ignoring surrounding whitespace — close to how `indexExtensionlessEnabled()` reads `INDEX_EXTENSIONLESS`. Unset, empty and whitespace-only all mean "leave the default alone", so a compose line with nothing after the `=` does not stop the server. Unlike that helper, which treats an unrecognised value as "on", any other value here is rejected with an error naming it: a typo should announce itself as a configuration error rather than turn into search results that are quietly worse than intended.
- Resolve it in `embedding-config.ts` alongside the other `EMBEDDING_*` variables and log it with them, so a misconfiguration is visible in the host's log rather than only in the ranking. Resolving it there also means an invalid value surfaces during a tool call, where the message reaches the host, instead of during module evaluation.
- Document in the README that turning it off removes path-derived tokens from **lexical** search as well as from the dense vector, and that for context artifacts — whose `filePath` is a `context:<name>:<path>` identifier — it drops the artifact name too. Neither is obvious from the setting's name.

Changing it requires a full re-index: `codebase_remove`, then `codebase_index`, as a plain re-run skips unchanged files by content hash.

<details>

### 何を直すか

埋め込むテキストにファイルパスを含めるかどうかを、環境変数で選べるようにします。

### なぜ必要か

`prepareDocumentText()` は必ずファイルパスをチャンク本文の前に置きます。この文字列がそのままベクトルストアへ渡り、密ベクトルと、同じ文字列から作られる BM25 の疎ベクトルの両方を作ります。

パスに含まれる語は、パスを手がかりにした検索では役立ちます。しかし設計文書や仕様書のように、ファイル名がほとんど情報を持たないコーパスでは、すべてのチャンクに雑音が乗ります。現状これを止める方法がありません。

### 変更の内容

環境変数を 1 つ追加します（既定は有効なので、これまでの動作と既存のインデックスは変わりません）。

値の解釈は既存の `INDEX_EXTENSIONLESS` に近い形にしました。前後の空白を落とし、大文字小文字を区別しません。未設定・空文字・空白のみは「既定のまま」を意味します。設定ファイルで `=` の後ろが空になっていてもサーバが止まらないようにするためです。

ただし 1 点だけ既存と違います。**認識できない値は例外にします。** 既存のヘルパは未知の値をどちらかに倒しますが、打ち間違いは「検索結果が思ったより悪い」という形ではなく、設定エラーとして知らせるべきだと考えました。

無効にすると、密ベクトルだけでなく字句検索（BM25）からもパス由来の語が消えます。また context artifact（`context:<名前>:<パス>` という識別子を持つチャンク）では、その識別子ごと落ちます。どちらも設定名からは分からないので README に書きました。

</details>

## Type of change

- [ ] Documentation update

## Testing

- [x] Unit tests pass (`npm run test:unit`) — 56 files, 1204 tests
- [ ] Integration tests pass (`npm run test:integration`) — not run; requires Docker. `tests/integration/embeddings.test.ts` is updated so its assertions no longer depend on this variable being unset in the developer's shell

`EMBEDDING_DOCUMENT_INCLUDE_PATH=false npm run test:unit` also passes, so the suite does not break for the people the feature is written for. Reverting the call in `prepareDocumentText` to a fixed `true` or `false` fails the new tests, so they exercise the implementation.

## Checklist

## Related issues

None. Happy to move the discussion to an issue first if you prefer that.

---

**One thing to watch if both this and #126 are merged.** That PR makes the document prefix configurable; this one writes `search_document: ` as a literal, which is correct on its own because the prefix is not configurable on this branch. If #126 lands first and this branch is rebased mechanically, the conflict resolution can quietly drop the configurable prefix. The correct merged form is:

```ts
const head = documentIncludesPath() ? `${filePath}\n` : "";
return `${documentPrefix()}${head}${content}`;
```

Each branch's tests pass on their own either way, so this would not be caught by CI. I will rebase by hand and add a test covering both variables set together, whichever order you merge in.

The resolved value is logged but not stored on `EmbeddingConfig`; `prepareDocumentText` reads it lazily at each call. Adding a field to that type felt like a separate change from adding the variable, so I left it — happy to fold it in.

## Summary by CodeRabbit

* **New Features**
  * Added an option to include or omit document file paths from embedded text.
  * Added `EMBEDDING_DOCUMENT_INCLUDE_PATH`, enabled by default, with case-insensitive boolean values and validation for invalid settings.
  * When disabled, embedded content excludes file paths while preserving configured document prefixes.

* **Documentation**
  * Documented accepted values, default behavior, embedding impacts, and the requirement to re-index after changing this setting.


yusuke-omichi-ctc:
Thanks — changed in `6487c15`, though the failure you describe does not reproduce here, so I want to be transparent about what I verified.

On `vitest@4.1.10` (the version this lockfile resolves), with `EMBEDDING_DOCUMENT_INCLUDE_PATH` exported in the shell, `vi.stubEnv(key, undefined)` does isolate correctly:

- `EMBEDDING_DOCUMENT_INCLUDE_PATH=false npx vitest run tests/unit/embeddings.test.ts` → 13 passed
- `EMBEDDING_DOCUMENT_INCLUDE_PATH=maybe npx vitest run tests/unit/embeddings.test.ts` → 13 passed (an unisolated invalid value would throw)

That said, your underlying point stands: passing `""` does not depend on `stubEnv` being able to delete a variable at all, and for this variable `""` means exactly "leave the default alone" — so the change is free. Both call sites now pass `""`, with a comment explaining why that value and not `undefined`.

Unit tests after the change: 56 files, 1204 tests.

giancarloerra:
#126 is now merged, and this branch conflicts with current `main`. #125 also conflicts with this PR in `src/services/embedding-config.ts`.

Please wait for #125 to land, then rebase this branch onto current `main` so the shared configuration conflict is resolved once.

During the rebase, preserve the combined prefix and optional-path behavior:

```ts
const head = documentIncludesPath() ? `${filePath}\n` : "";
return `${documentPrefix()}${head}${content}`;
```

Please also add the promised combined regression test. Set a custom `EMBEDDING_DOCUMENT_PREFIX` together with `EMBEDDING_DOCUMENT_INCLUDE_PATH=false` and assert that the configured prefix remains while the path is absent.

yusuke-omichi-ctc:
Understood — I will hold this branch until #125 lands, then rebase onto `main` once so the shared `src/services/embedding-config.ts` conflict is resolved in a single pass.

Noting the two things to carry through that rebase, so they are not lost:

1. The combined form you spelled out, rather than whatever a mechanical resolution produces:

```ts
const head = documentIncludesPath() ? `${filePath}\n` : "";
return `${documentPrefix()}${head}${content}`;
```

2. The combined regression test: a custom `EMBEDDING_DOCUMENT_PREFIX` together with `EMBEDDING_DOCUMENT_INCLUDE_PATH=false`, asserting the configured prefix survives while the path is gone. That is exactly the case neither branch could cover on its own, which is why the mechanical resolution would have gone unnoticed.

I am working through your comments on #124, #125 and #128 now and will come back to this one after #125 is merged.

giancarloerra:
Merged 125, can you rebase this and let me know once done? Thanks!

yusuke-omichi-ctc:
Rebased onto `main` (`a4064c2`) in `87b677d`, still one commit. Both things you asked me to carry through are in, and GitHub now reports the branch as mergeable.

**The combined form**, verbatim as you spelled it:

```ts
const head = documentIncludesPath() ? `${filePath}\n` : "";
return `${documentPrefix()}${head}${content}`;
```

**The combined regression test.** A custom `EMBEDDING_DOCUMENT_PREFIX="passage: "` with `EMBEDDING_DOCUMENT_INCLUDE_PATH=false` asserts `"passage: const a = 1;"` — configured prefix kept, path gone, in one assertion. Two more cover the ends of the range: the empty prefix bge-m3 wants (content alone), and a prefix configured without a trailing space (`検索文書:`) running straight into the content, since with the path off nothing separates them but the prefix itself.

To check those three actually guard what you were worried about, I reverted `prepareDocumentText` to the hardcoded `search_document: ` this branch used to have: exactly those three fail, the other 23 in the file pass. So a future mechanical resolution that undoes #126 gets caught here.

**Folded into `parseBooleanEnv()`.** Now that #125 is in, `documentIncludesPath()` is a call to the shared helper rather than its own parser — the accepted spellings, the whitespace handling and the unset/empty/whitespace-only-means-default rule were already identical, so there was no reason for a second one. The error text is now the helper's, and the test that asserts it exactly was updated to match.

**One doc line this PR had to catch up with.** The `MAX_CHUNK_CHARS` docblock #128 landed says `prepareDocumentText` prepends "the path and a newline" and puts the 2048-char crossover at a path longer than 30 characters. Both stop holding once the path is optional, so that paragraph now scopes them to the path being on.

**Env-pollution fix outside the new file.** `tests/unit/embedding-prefix.test.ts` asserts `prepareDocumentText` output with the path in it but does not pin the new variable, so `EMBEDDING_DOCUMENT_INCLUDE_PATH=false` in a developer's shell fails 5 of its cases. Added the same `vi.stubEnv(..., undefined)` line the file already uses for the two prefix variables. Same reason for the additions in `tests/unit/embeddings.test.ts`, `tests/integration/embeddings.test.ts` and `tests/unit/embedding-config.test.ts`; the first two also had a duplicate `beforeEach` after the rebase, now merged into one.

Verification: `npm run lint` clean (121 files), `npx tsc --noEmit` clean, `npm run build` clean, `npm run test:unit` 63 files / 1364 tests. Also 1364 with `EMBEDDING_DOCUMENT_INCLUDE_PATH=false` and again with `EMBEDDING_DOCUMENT_PREFIX='検索文書:'`, so the suite does not depend on either being unset. `git diff main..HEAD` deletes 22 lines, all of them lines this PR rewrites — none of the five merged branches is rolled back.

**Unrelated, repeating it here since it is easy to hit:** `npm test` fails for me on `tests/e2e/full-workflow.test.ts > step 5 — build code graph (fire-and-forget)` — the fire-and-forget build from step 4 is still running when step 5 asserts. It fails on a clean `main` worktree too (2 of 3 runs there), so it is not from this branch.

PR #128, opened 2026-08-31
Make the per-chunk character cap configurable

yusuke-omichi-ctc:
## Summary

`MAX_CHUNK_CHARS` caps every chunk at 2000 characters, and content past the cap is dropped by `applyCharCap` before the chunk is stored — it reaches neither the vector, nor the payload, nor the BM25 text, so no search retrieves it. The cap has to match the embedding model's context window, but it is a hardcoded constant.

At the default it silently loses content on short-context models, in two different ways. Measured on a 54-chunk corpus of TypeScript with Japanese comments, tokenised with the model's own tokeniser: 14 chunks (26%) hit the 2000-char cap, and 17 chunks (31%) exceed 512 tokens.

- The 26% is the lossy case: `applyCharCap` drops those tails before `upsertChunks` builds either the payload or the BM25 text, and nothing later in the pipeline puts them back.
- The 31% is milder: those chunks overflow a 512-token window, so the provider pre-truncates the text it embeds, and their tails are absent from the vector while the full chunk remains in the payload and the BM25 text, where keyword search still finds it.

Neither loss is reported.

## Changes

- Read `MAX_CHUNK_CHARS` from the environment, defaulting to 2000 so existing behaviour and existing indexes are unchanged. A non-integer or non-positive value throws at module load, naming the offending value, rather than failing part-way through an index.
- Document in the README, `src/constants.ts`, `DEVELOPER.md` and the codebase-management skill reference what the cap actually does: it **truncates, it does not re-split**. For a language with an ast-grep grammar, chunks are cut at top-level declaration boundaries and adjacent chunks do not overlap, so a truncated tail is absent from the index rather than carried into the next chunk. `chunkByCharacters` — the minified/bundled path gated by `MAX_AVG_LINE_LENGTH` — is the only place where this cap decides chunk boundaries.
- Note that raising it beyond the model's context length × the provider's chars-per-token estimate does not put more content into the embedding: the provider pre-truncates, so the extra characters reach the payload and the BM25 text but are not represented in the vector.
- Note that a character cap cannot guarantee a token budget. The same corpus measured 1.41–3.39 chars per token, and emoji or rare CJK glyphs drop below 1.0 (8 emoji → 23 tokens).

Changing the cap changes the stored chunks, and re-running indexing alone does not apply it, because unchanged files are skipped by content hash. The index has to be rebuilt: `codebase_remove`, then `codebase_index`.

<details>

### 何を直すか

1 チャンクあたりの文字数上限を環境変数で設定できるようにします。

### なぜ必要か

`MAX_CHUNK_CHARS` が全チャンクを 2,000 文字で切り、超えた部分を捨てています。捨てられた部分はベクトルにも payload にも BM25 のテキストにも入らないため、**どの検索でも見つかりません。**

この上限は埋め込みモデルのコンテキスト長に合わせる必要があります。ところが定数でハードコードされており、モデルを替える設定はあるのに上限を合わせる方法がありません。

日本語コメント付き TypeScript の 54 チャンクをモデル自身のトークナイザで測ったところ、14 チャンク（26%）が 2,000 文字の上限に達し、17 チャンク（31%）が 512 トークンを超えていました。**2 つの損失は性質が違います。** 前者は索引に入らないので取り出せません。後者はベクトルに表れないだけで、payload と BM25 には残るため字句検索では見つかります。どちらも通知されません。

### 変更の内容

環境変数から読むようにし、既定は 2,000 のままにしました。整数でない値や 0 以下の値は、その値を挙げて起動時に例外にします。インデックス構築の途中で失敗するより早く気づけるためです。

あわせて、上限が実際に何をするのかを文書化しました。**これは切り詰めであって、再分割ではありません。** ast-grep の文法を持つ言語では、チャンクはトップレベルの宣言の境界で切られ、隣接するチャンクに重複がありません。したがって切り詰められた末尾は、次のチャンクへ持ち越されるのではなく、索引から消えます。

また、文字数の上限ではトークン数を保証できません。同じコーパスで 1 文字あたりのトークン比は 1.41〜3.39 の幅があり、絵文字や稀な CJK 文字では 1.0 を下回ります（絵文字 8 個で 23 トークン）。

</details>

## Type of change

- [x] Documentation update

## Testing

- [x] Unit tests pass (`npm run test:unit`) — 56 files, 1197 tests
- [ ] Integration tests pass (`npm run test:integration`) — not run; requires Docker

The new tests include a case that drives `chunkByCharacters` with a non-default cap (four 3000-character lines, so the average line length crosses `MAX_AVG_LINE_LENGTH`) and asserts the chunk count rises as the cap falls. Replacing the `MAX_CHUNK_CHARS` reference inside `chunkByCharacters` with the literal 2000 fails that test, so it exercises the implementation rather than the constant.

## Checklist

## Related issues

None. Happy to move the discussion to an issue first if you prefer that.

---

One thing I deliberately did **not** add: validation against an upper bound. The docblock says the cap must stay below `CHARS_PER_TOKEN_ESTIMATE × model_context_length`, but `src/constants.ts` is the lowest layer — it imports only `node:module` — and `MAX_CHUNK_CHARS` is resolved at module load, while the embedding config resolves at call time. Comparing them at load would risk throwing during import and would invert the layering, so I documented the limit instead. If you would rather have the check, the natural home is wherever the embedding config is read; I am happy to add it there.

The 26% / 31% / 1.41–3.39 figures come from a corpus I measured outside this repository, so they are not reproducible from the tree. Reviewing this repository's own `src/**/*.ts` at the default cap gives 122 of 474 chunks (25.7%) at the cap, which lines up with the 26%.

This is the last of five independent branches I am offering.

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring the maximum indexed chunk size with the `MAX_CHUNK_CHARS` environment variable.
  * Chunks default to 2,000 characters and are truncated when they exceed the configured limit.
  * Invalid values are rejected with a clear error.

* **Documentation**
  * Updated configuration and chunking guidance, including rebuild requirements after changing the limit.


giancarloerra:
Four corrections remain before merge:

1. Scope “truncates, it does not re-split” to the small-file, AST, and line-based paths. The minified/bundled `chunkByCharacters()` path uses `MAX_CHUNK_CHARS` as a split boundary and can produce additional chunks. Please correct the absolute wording in the [README](https://github.com/giancarloerra/SocratiCode/blob/7598c5db7b98a46e2c03d74080ff0fddf2100fbd/README.md#L1231), constants documentation, and [skill reference](https://github.com/giancarloerra/SocratiCode/blob/7598c5db7b98a46e2c03d74080ff0fddf2100fbd/skills/codebase-management/references/tool-reference.md#L252).

2. The [`MAX_CHUNK_CHARS` documentation](https://github.com/giancarloerra/SocratiCode/blob/7598c5db7b98a46e2c03d74080ff0fddf2100fbd/src/constants.ts#L141-L159) says staying below the model budget makes provider pre-truncation “always fire”. It should say that staying below the budget prevents provider truncation from being needed, while retaining it as the final safety net.

3. Replace the hardcoded [`2000 chars`](https://github.com/giancarloerra/SocratiCode/blob/7598c5db7b98a46e2c03d74080ff0fddf2100fbd/DEVELOPER.md#L370) in the indexing data flow with the configured cap, defaulting to 2000.

4. Strengthen the [configured truncation test](https://github.com/giancarloerra/SocratiCode/blob/7598c5db7b98a46e2c03d74080ff0fddf2100fbd/tests/unit/max-chunk-chars.test.ts#L101-L115) to assert exactly one chunk containing exactly the first 120 characters. Its current `length > 0` and `length <= 120` assertions would also pass if the implementation re-split the input into several chunks.

yusuke-omichi-ctc:
All four addressed in `e186eb4`, rebased onto current `main`.

**1. Scope of "truncates, it does not re-split".** You are right, and I had the paths backwards in the absolute wording. Reading `chunkFileContent` again, the cap plays two different roles:

| Path | Gate | Cap's role |
|---|---|---|
| `chunkByCharacters` | `avgLineLength > MAX_AVG_LINE_LENGTH` | **split boundary** — `Math.min(offset + MAX_CHUNK_CHARS, …)` per iteration, so a lower cap produces more chunks |
| whole-file single chunk | `lines.length <= CHUNK_SIZE` | truncation only |
| `chunkByAstRegions` | AST regions found | truncation only |
| `chunkByLines` | fallback | truncation only |

Corrected in the README row, the `src/constants.ts` docblock and the skill reference. I found the same absolute wording in the header comment of `tests/unit/max-chunk-chars.test.ts` and scoped that too.

While re-reading I also narrowed a second claim: the "the dropped tail is not recovered from the next chunk" statement only holds on the AST path, since `chunkByLines` carries `CHUNK_OVERLAP`. That is now scoped as well.

**2. Pre-truncation wording.** Changed from "always fires as a second line of defence" to staying below the budget meaning provider truncation *is not needed*, leaving it as the last-resort defence rather than a routine one.

**3. `DEVELOPER.md` data flow.** `Hard character cap (2000 chars)` → `Hard character cap (\`MAX_CHUNK_CHARS\`, default 2000 chars)`.

**4. Truncation test.** Now `expect(chunks).toHaveLength(1)` and `expect(chunks[0].content).toBe(long.slice(0, 120))`.

To check the strengthening actually closes the hole you named, I mutated `applyCharCap` from truncating to re-splitting: the new assertions fail with `expected [...] to have a length of 1 but got 5`, while the old `length > 0` / `<= 120` pair would have passed on those same five chunks.

Verification: `npm run lint` clean, `npx tsc --noEmit` clean, `npm run test:unit` 60 files / 1274 tests. Reverting the `MAX_CHUNK_CHARS` reference inside `chunkByCharacters` to the literal 2000 still fails the split-boundary test. `src/services/indexer.ts` is untouched by this PR.

The commit message carried the same absolute wording and is corrected as well.

yusuke-omichi-ctc:
Follow-up in `a7c8ed6` (amended, still one commit on `main`). CodeRabbit was right about the arithmetic in the `MAX_CHUNK_CHARS` docblock, and correcting it turned up two more errors in the same paragraph.

**Your wording from point 2 is untouched.** `git diff` shows the "pre-truncation is not needed, leaving it as the last-resort defence rather than a routine one" sentence as a context line, byte-identical. What I rewrote is the sentence *after* it, which came from before this PR.

**The old claim.** "The default 2000 leaves headroom for the `prepareDocumentText` path prefix (~50 chars) prepended at embed time." With the tightest budget being 2048 chars, 2000 + 50 = 2050 does not leave headroom — it crosses. Two further problems:

- **The prepended length is not a constant.** `prepareDocumentText` prepends `documentPrefix()` + the path + a newline. The prefix is configurable, and the path length varies per chunk (and `context-artifacts.ts` passes `context:<name>:<path>`, which is longer than a relative path). "~50 chars" cannot be right for all of them.
- **`CHARS_PER_TOKEN_ESTIMATE` is not one number.** It is 1.0 in `provider-ollama.ts` and 3.0 in the other four, so ollama is the tightest case and the one 2000 is sized against.

The docblock now says the cap bounds the chunk body only, that the prepended header pushes the embedded text past it by a non-constant amount, and gives the exact boundary: under the 17-char default prefix a 2000-char chunk crosses a 2048-char budget once the path exceeds 30 characters. `pretruncateTexts` cuts on `t.length > maxChars`, so a 30-char path (exactly 2048) survives and 31 is the first that does not.

**One thing I got wrong and caught before pushing.** I had written that pre-truncation has a gap — that a model outside `MODEL_CONTEXT_LENGTHS` gets `guessContextLength() === 0`, hits `if (contextLength <= 0) return texts;` and loses the defence entirely. That is wrong. All five call sites substitute a positive default first (`provider-ollama.ts` even says so in a comment), so 0 never reaches the function. The docblock now states what is actually true: pre-truncation always runs, but the substituted default (2048 tokens everywhere except openai's 8191) need not match the real model, so `EMBEDDING_CONTEXT_LENGTH` is what sizes it correctly.

I also fixed the "~50 characters" comment that had survived in `tests/unit/max-chunk-chars.test.ts`, and two type errors in that file: `await import(...).catch(e => e as Error)` widens to `Error | typeof import(...)`, so `.message` did not type-check. It is now `.then(() => null, e => e as Error)` plus `expect(err).toBeInstanceOf(Error)`, which also asserts directly that the import still throws instead of inferring it from an `undefined` message. The repo's `tsconfig.json` only includes `src/**/*`, so `npx tsc --noEmit` never saw these — I checked the tests with a temporary config that adds `tests/**/*`.

Verification: `npm run lint` clean, `npx tsc --noEmit` clean, `npm run build` clean, `npm run test:unit` 60 files / 1274 tests.

**Unrelated, but you may want to know.** `npm test` (the full run) fails for me on `tests/e2e/full-workflow.test.ts > step 5 — build code graph (fire-and-forget)`: `expected '⚠ Graph build already in progress for…' to contain 'Graph build started in the background'`. The fire-and-forget build from step 4 is still running when step 5 asserts. It is not from this PR — on a clean `c32cdde` worktree it failed 2 of 3 runs, and on this branch 3 of 3. Happy to open a separate issue or PR for it if useful.

giancarloerra:
Please open a separate follow-up PR (merging 128 now).

The defect is in `tests/e2e/full-workflow.test.ts::waitForIndexingComplete()`, not in the `MAX_CHUNK_CHARS` change.

The current predicate accepts any status containing `chunks`, but `codebase_status` prints `Indexed chunks` while indexing is still active. Step 3b therefore returns early and step 5 can collide with indexing's automatic graph build.

Please:

- Return only when a status line starts with `Last operation: Full index` and ends with `completed`.
- Do not use generic `chunks`, `files`, or `indexed` text as completion evidence.
- Fail immediately if the last full-index operation reports failure.
- Add a regression proving that an in-progress status containing `Indexed chunks` does not satisfy the completion predicate, while the completed status does.
- Run the full workflow repeatedly to confirm step 5 no longer encounters an already-running graph build.

Keep this as a test synchronisation fix.

Issue #31, opened 2026-04-20
[Bug]: using external ollama and qdrant runs docker containers locally


hertznsk:
### Describe the bug

When using external ollama and qdrant, docker images are pulled anyway and containers are launched.
It doesn't seem to make sense.

### Steps to reproduce

opencode.json:
```json
  "mcp": {
    "socraticode": {
      "type": "local",
      "command": ["npx", "-y", "socraticode"],
      "env": {
        "OLLAMA_MODE": "external",
        "OLLAMA_URL": "http://REDACTED",
        "EMBEDDING_MODEL": "nomic-embed-text-v2-moe",
        "EMBEDDING_DIMENSIONS": "768",
        "QDRANT_MODE": "external",
        "QDRANT_URL": "http://REDACTED",
        "QDRANT_API_KEY": ""
      }
    }
  }
```

```shell
$ docker ps
CONTAINER ID   IMAGE                                COMMAND                  CREATED          STATUS                       PORTS                                                                                          NAMES
be9e82e69729   ollama/ollama:latest                 "/bin/ollama serve"      2 minutes ago    Up 2 minutes                 0.0.0.0:11435->11434/tcp, [::]:11435->11434/tcp                                                socraticode-ollama
587074fc1553   qdrant/qdrant:v1.17.0                "./entrypoint.sh"        59 minutes ago   Up 59 minutes                0.0.0.0:16333->6333/tcp, [::]:16333->6333/tcp, 0.0.0.0:16334->6334/tcp, [::]:16334->6334/tcp   socraticode-qdrant
```

### Expected behavior

ollama and qdrant images do not pull docker and containers do not start.

### Actual behavior

ollama and qdrant, docker images are pulled anyway and containers are launched

### Embedding provider

Ollama (native / external)

### Qdrant mode

External (self-hosted / cloud)

### MCP host

Other (specify below)

### Operating system

Fedora 43

### Node.js version

v22.22.2

### Relevant log output

```shell

```

### Additional context

Thanks for this tool! Although I'm still in the process of indexing for the first time with this tool, it looks promising.

### Checklist

- [x] I have searched existing issues and this hasn't been reported before
- [x] I am using the latest version of SocratiCode

giancarloerra:
Hi, there's no code path that creates containers in external mode:

[ensureQdrantReady](docker.ts:154): early-returns to [ensureExternalQdrantReady] when [QDRANT_MODE=external]: it does only HTTP healthcheck, no docker calls

[...]

PR #130, opened 2026-08-31
feat(gdscript): native loader, per-file Godot roots, relative .tscn paths

JiminyJoo:
## Summary

Add first-class GDScript and Godot resource support without requiring existing users to rebuild their indexes or changing the supported Node.js floor.

## Changes

- Index `.gd`, `.tscn`, and `.tres` files in new and existing projects.
- Use native GDScript AST parsing when `tree-sitter-gdscript` is available.
- Fall back to syntax-aware GDScript chunking and import extraction when the optional native parser cannot load.
- Extract and resolve `extends`, `preload()`, `load()`, `res://`, relative Godot resource paths, `class_name`, and Godot UID references.
- Resolve each file against its nearest `project.godot`, including repositories containing multiple Godot projects.
- Keep runtime-relative `load()` paths rooted at the Godot project while source-relative `extends` and `preload()` paths resolve from the importing file.
- Prevent `.uid` sidecars from becoming graph nodes or resolving to deleted resources.
- Package and preflight the optional native parser portably, including hoisted and linked npm installations.
- Document full and fallback support and the optional native dependency.

## Backward compatibility

- Existing indexes remain usable and gain the new Godot file types through an ordinary incremental update. No re-index is required.
- Existing non-Godot extension profiles and unchanged files are preserved.
- The optional parser cannot make installation fail; unsupported native environments retain fallback GDScript support and all other dynamic languages.
- Node.js `>=18.17.0` remains supported.
- Qdrant remains on the existing compatible `1.18.x` line.

## Validation

- `npm run lint`
- `npm run build`
- Full unit suite: 1,832 tests passed
- Focused GDScript, index-profile, indexer, and watcher suite: 246 tests passed
- Packed-package smoke tests on the current Node.js runtime and Node.js 18.17
- Packed-package fallback smoke test with the optional GDScript parser absent
- Linked npm installation and GDScript test suite
- Local CodeRabbit review: no findings

## Summary by CodeRabbit

* **New Features**
  * Added indexing support for GDScript files and Godot resources (`.tscn`, `.tres`).
  * Added resolution for Godot paths, class names, UIDs, and project references.
  * Added profile-aware indexing for file discovery, chunking, embeddings, and updates.
  * GDScript analysis uses enhanced parsing when available, with a fallback for broader compatibility.

* **Documentation**
  * Updated supported file types, language counts, configuration guidance, engine update instructions, and Godot support details.
  * Documented optional GDScript parser requirements and fallback behavior.


giancarloerra:
Thanks for this! The GDScript direction makes sense, but the current head is not merge-ready.

## Blocking

1. Please rebase onto current `main`. The branch currently conflicts in seven files, including `package.json`, `code-graph.ts`, `graph-resolution.ts`, and `indexer.ts`. Preserve the Node 26/Qdrant compatibility guard and the graph changes already on `main`.

2. Commit an updated `package-lock.json`. The current locked tree is inconsistent with `package.json`: `npm ci` rejects the Qdrant version and reports `tree-sitter-gdscript` and its dependencies as missing.

3. Remove filename-based resolution for bare GDScript class references. Godot resolves `extends Foo` through the globally registered `class_name Foo`; an unnamed script is referenced by path. A file named `Foo.gd` must not win over, or substitute for, the file that actually declares `class_name Foo`.

   Please add regressions covering:
   - `Foo.gd` declaring a different class while another file declares `class_name Foo`
   - a file named after a Godot built-in type without a matching `class_name`

4. Resolve the Linux ARM64 behavior. `tree-sitter-gdscript@6.1.0` has no Linux ARM64 prebuild, and the current loader does not use a source-built addon. Either support the package's built artifact or explicitly constrain the supported platform claim and ensure unsupported systems do not advertise AST and graph support.

5. Remove commit `067dfcd2` from this PR. SocratiCode declares Node `>=18.17.0` and tests Node 18, 20, and 22. `@qdrant/js-client-rest@1.19.0` requires Node `>=22.0.0` and depends on `undici@7.29.0`, which requires Node `>=20.18.1`. The existing `^1.17.0` range already permits compatible installations to resolve Qdrant 1.19. Raising the minimum to 1.19 is unnecessary for Node 26 and removes the Node 18-compatible 1.18 resolution. Keep the Qdrant change out of this PR (happy to revisit and solve it in a different PR!).

6. After the rebase and fixes, run CI and a full CodeRabbit review against the new head.

## Additional corrections

- Add `export_variable_statement` and `onready_variable_statement` to the GDScript top-level chunk kinds.
- Add `tree-sitter-gdscript` to `THIRD-PARTY-LICENSES`.

The Qdrant upgrade may remain as a separate commit in this PR. It does not need to be split solely for formatting or title reasons.

giancarloerra:
The enabled CI confirms that current head `7542518` still has two code blockers.

1. Restore the Node 18-compatible Qdrant locked tree.

   `package.json` correctly remains at `@qdrant/js-client-rest: ^1.17.0` with Node `>=18.17.0`, but `package-lock.json` now resolves:

   - `@qdrant/js-client-rest@1.19.0`, requiring Node `>=22.0.0`
   - its nested `undici@7.29.0`, requiring Node `>=20.18.1`

   The [Node 18 CI job](https://github.com/giancarloerra/SocratiCode/actions/runs/33473996127/job/99799090205) consequently fails with `ReferenceError: File is not defined`. Node 22 passed; Node 20 was cancelled after the matrix failure.

   Preserve the new `tree-sitter-gdscript` entries, but restore the dependency result used by current `main`:

   - `package.json`: `@qdrant/js-client-rest: ^1.17.0`
   - `package-lock.json`: `@qdrant/js-client-rest@1.18.0`
   - Qdrant's Undici 6 dependency rather than nested Undici 7

2. Fix explicit `res://` resource resolution in both the `gdscript` and `godot-resource` branches.

   Both paths currently pass `[".gd", ".tscn", ".tres"]` to `resolveRelativePath()`. This can convert an unmatched explicit path such as `res://assets/player.png` into a false edge to `assets/player.tscn`, as described in the [current CodeRabbit finding](https://github.com/giancarloerra/SocratiCode/pull/130#discussion_r3900943180).

   Use direct file-set matching without fallback extensions in both branches:

   ```ts
   return resolveRelativePath(resPath, basePath, projectPath, fileSet, []);
   ```

   Add the requested regression proving that `res://assets/player.png` does not resolve to `assets/player.tscn`.

The README link finding is non-blocking. If correcting it in this update, point it to `#language-support`.

After the other ready PRs have landed, rebase once onto the then-current `main`, rerun the complete Node 18/20/22 CI matrix, and request a full CodeRabbit review against the final head.

giancarloerra:
Thanks for this!

Commit `4a82206` fixes the two blockers from the previous review:

- The lockfile now resolves `@qdrant/js-client-rest@1.18.0` with Undici 6, preserving Node 18 support.
- Both `gdscript` and `godot-resource` now use direct matching for explicit `res://` paths, including the requested regression for `player.png` versus `player.tscn`.
- The README anchor is corrected.

The PR is not merge-ready yet.

### Blocking

#### 1. The native GDScript loader ignores source-built artifacts and can silently produce an approximate graph

[`code-graph.ts`](https://github.com/giancarloerra/SocratiCode/blob/4a8220674f088bdf2afd1bc4fa36eca48466628c/src/services/code-graph.ts#L589-L637) hardcodes four packaged prebuild locations. However, `tree-sitter-gdscript` uses [`node-gyp-build`](https://github.com/prestonknopp/tree-sitter-gdscript/blob/v6.1.0/bindings/node/index.js#L1-L7), which selects either a matching prebuild or an artifact compiled into `build/Release`.

Consequences:

- On Linux ARM64, installation may successfully compile the addon, but SocratiCode ignores the resulting artifact and declares the parser unavailable.
- If the native binary cannot install, `tree-sitter-gdscript` is a normal dependency, so installation can fail before the runtime fallback is reached.
- The GDScript binary is added to the same registration batch as every other dynamic grammar without validating that the binary loads.
- `loadedDynamicLanguages` records GDScript before registration succeeds. A registration failure can therefore leave diagnostic status incorrect and can affect the registration batch for unrelated languages.
- The fallback in [`graph-imports.ts`](https://github.com/giancarloerra/SocratiCode/blob/4a8220674f088bdf2afd1bc4fa36eca48466628c/src/services/graph-imports.ts#L193-L217) scans raw source. It therefore treats `preload(...)` or `extends Foo` inside comments and string literals as real dependency edges.

Required correction:

- Resolve the actual installed native artifact through `node-gyp-build`, including source-built output.
- Validate the artifact before adding it to the shared registration batch.
- Mark GDScript as loaded only after registration succeeds.
- Preserve installation for users who do not use GDScript.
- If GDScript parsing is unavailable, report that state explicitly rather than advertising regex and line-based fallback as Full Support.
- Add explicit loader tests for a packaged prebuild, source-built artifact, missing or invalid artifact, accurate status, and isolation of the other dynamic grammars.

#### 2. Godot project roots and `class_name` registries are incorrectly global to the indexed repository

[`buildCodeGraph`](https://github.com/giancarloerra/SocratiCode/blob/4a8220674f088bdf2afd1bc4fa36eca48466628c/src/services/code-graph.ts#L831-L840) computes one Godot root by walking upward from the indexed path and builds one `class_name` index across every `.gd` file.

For an indexed repository such as:

```text
repo/
  game-a/project.godot
  game-a/scripts/A.gd
  game-b/project.godot
  game-b/scripts/A.gd
```

the current implementation:

- finds no `project.godot` by walking upward from `repo/`;
- falls back to resolving `res://` against `repo/`, so valid paths inside either game do not resolve;
- combines both projects into one `class_name` map, allowing one project to overwrite or resolve classes from the other.

Godot defines the directory containing `project.godot` as the project root, and `res://` always points to that root: [Godot file-system documentation](https://docs.godotengine.org/en/stable/tutorials/scripting/filesystem.html).

Required correction:

- Determine the nearest `project.godot` ancestor for each Godot source file.
- Maintain `class_name` indexes scoped to each Godot project root.
- Do not fall back from a missing `project.godot` to the arbitrary SocratiCode indexing root.
- Add `buildCodeGraph` regressions for a nested Godot project and two sibling Godot projects containing duplicate class names.

#### 3. Valid relative `.tscn` and `.tres` resource paths are discarded

The extractor accepts every `[ext_resource path="..."]`, but [`resolveImport`](https://github.com/giancarloerra/SocratiCode/blob/4a8220674f088bdf2afd1bc4fa36eca48466628c/src/services/graph-resolution.ts#L1619-L1630) only resolves paths beginning with `res://`. A valid path such as:

```text
[ext_resource type="Material" path="material.tres" id="1"]
```

is classified as external and receives no edge.

Godot documents both `res://` paths and paths relative to the `.tscn` file as valid: [TSCN external-resource documentation](https://docs.godotengine.org/en/stable/engine_details/file_formats/tscn.html#external-resources).

Resolve relative resource paths from the directory containing the `.tscn` or `.tres` file and add regressions for both forms.

The same documentation lists valid section headings as `ext_resource`, `sub_resource`, `node`, and `connection`. `instance` is an attribute of a `node`, not an `[instance path="..."]` section. The new `[instance]` extractor, test, and README claim should be removed unless a documented Godot format version using that section can be identified. Scene-to-scene dependencies are already represented by `PackedScene` entries in `ext_resource`.

### Final validation

After these corrections:

1. Rebase onto current `main`. The branch is currently 10 commits behind.
2. Run the complete Node 18, 20, and 22 CI matrix against the rebased head.
3. Run platform-specific native-loader checks for every claimed supported path.
4. Request a full CodeRabbit review. The current green CodeRabbit status is not full coverage: its report says reviews are paused, `package-lock.json` was excluded, and all selected files in the latest delta were skipped as similar.
5. Update the title and PR description. They still describe a Qdrant 1.19/Node 26 change, while the current diff preserves the existing `^1.17.0` declaration and 1.18 lock resolution.

giancarloerra:
Current head `6565f6a` completes these previously requested items:

- The Node 18-compatible Qdrant dependency tree is restored: `package.json` remains on `^1.17.0`, while the lockfile resolves `1.18.0` with Undici 6.
- Explicit `res://` paths use direct matching, including the `player.png` versus `player.tscn` regression.
- `tree-sitter-gdscript` is optional, and `node-gyp-build.path()` now discovers both packaged prebuilds and `build/Release` output.
- `gdscriptParserAvailable` is set only after the shared registration call succeeds.
- Normal nested and sibling projects now receive separate `class_name` maps and nearest-root handling.
- Relative `.tscn` and `.tres` paths are resolved from the resource file directory, and the invalid `[instance]` section extractor is removed.
- The requested top-level GDScript kinds and third-party licence entry are present.
- The branch is current with `main`, mergeable, and all four current CI checks pass for Node 18, 20, and 22. The title is also corrected.

Those completed points do not need further work. The PR is not merge-ready because the following current-head issues remain.

## Previously requested, still incomplete

### 1. Native-loader validation, failure isolation, status, and tests

The current loader obtains a path and calls `accessSync()` before adding GDScript to the shared registration batch. File accessibility does not prove that the addon can load, exports `tree_sitter_gdscript`, or has a compatible Tree-sitter ABI: [`code-graph.ts`](https://github.com/giancarloerra/SocratiCode/blob/6565f6a76b045922930429cffe7907e97c8d88e5/src/services/code-graph.ts#L606-L655). `node-gyp-build.path()` only selects a candidate path; its actual load path calls `require()` on that candidate: [`node-gyp-build.js`](https://github.com/prebuild/node-gyp-build/blob/v4.8.4/node-gyp-build.js#L19-L25).

This matters because ast-grep constructs the whole dynamic-language batch locally and replaces its global registry only after every candidate succeeds. Open-library, missing-symbol, and incompatible-ABI failures abort the batch: [`ast-grep dynamic registration`](https://github.com/ast-grep/ast-grep/blob/0.40.5/crates/dynamic/src/lib.rs#L103-L169). Current code nevertheless adds all non-GDScript candidates to `loadedDynamicLanguages` before that batch succeeds, does not record GDScript as failed when the shared call fails, catches the batch failure only in the outer guard, and leaves `dynamicLangsRegistered = true`, preventing a retry. The resulting status can claim unrelated grammars are loaded when none of the batch was registered.

The new tests cover only the successful native binary in the current environment and unconditionally require `gdscriptParserAvailable === true`: [`graph-gdscript.test.ts`](https://github.com/giancarloerra/SocratiCode/blob/6565f6a76b045922930429cffe7907e97c8d88e5/tests/unit/graph-gdscript.test.ts#L1122-L1179). They do not cover a source-built artifact, an omitted optional dependency, a corrupt or wrong-ABI addon, a missing symbol, batch-status reconciliation, retry behaviour, or isolation of the other grammars.

Required correction:

- Preflight the candidate with the same ast-grep registration and parse mechanism in an isolated child process before it enters the process-wide batch. Loading it only through Node validates the N-API addon but does not validate ast-grep's required exported symbol and Tree-sitter ABI.
- Add all successful candidates to `loadedDynamicLanguages` only after the parent batch succeeds.
- On batch failure, record accurate failure state for every affected candidate and keep retry state accurate.
- If GDScript preflight fails, omit only GDScript and register the validated unrelated grammars.
- Add deterministic packaged-prebuild, source-built, omitted, corrupt, wrong-ABI or missing-symbol, status, retry, and unrelated-grammar isolation tests. No test may assume that the optional parser exists on every supported platform.

### 2. Per-file Godot roots and the requested graph-level regressions

`buildGodotProjectIndexes()` discovers roots only while iterating `.gd` files: [`graph-resolution.ts`](https://github.com/giancarloerra/SocratiCode/blob/6565f6a76b045922930429cffe7907e97c8d88e5/src/services/graph-resolution.ts#L1175-L1205). `findGodotRootForFile()` then returns a known ancestor before checking the filesystem for a nearer `project.godot`: [`graph-resolution.ts`](https://github.com/giancarloerra/SocratiCode/blob/6565f6a76b045922930429cffe7907e97c8d88e5/src/services/graph-resolution.ts#L1218-L1238).

That fails for a resource-only project nested under a project containing GDScript:

```text
repo/project.godot
repo/outer.gd
repo/game/project.godot
repo/game/scene.tscn
repo/game/material.tres
```

The inner project is absent from the map. `scene.tscn` is assigned the outer known root without checking its nearer `game/project.godot`, so `res://material.tres` is resolved against the wrong project.

The requested `buildCodeGraph` regressions were also not added. The test file imports helper functions but not `buildCodeGraph`, and the root tests exercise those helpers directly: [`test imports`](https://github.com/giancarloerra/SocratiCode/blob/6565f6a76b045922930429cffe7907e97c8d88e5/tests/unit/graph-gdscript.test.ts#L4-L16), [`root tests`](https://github.com/giancarloerra/SocratiCode/blob/6565f6a76b045922930429cffe7907e97c8d88e5/tests/unit/graph-gdscript.test.ts#L1181-L1343).

Required correction:

- Compute and cache the nearest `project.godot` for every `.gd`, `.tscn`, and `.tres` file, not only for projects discovered through `.gd` files.
- Build each project's class map from its `.gd` files after the complete per-file root map exists.
- Add actual `buildCodeGraph` tests for nested projects, sibling projects with duplicate class names, files without `project.godot`, and a resource-only nested project.

### 3. Support status, PR description, documentation, and final review

The README still places GDScript under `Full Support` while the same paragraph states that parser absence changes it to regex extraction and line chunking with known false matches: [`README.md`](https://github.com/giancarloerra/SocratiCode/blob/6565f6a76b045922930429cffe7907e97c8d88e5/README.md#L993-L1003). The broader claim that there are no per-language parsers or unsupported-language limitations is also incompatible with this optional native parser: [`README.md`](https://github.com/giancarloerra/SocratiCode/blob/6565f6a76b045922930429cffe7907e97c8d88e5/README.md#L231-L240). The fallback test does not validate either documentation or unavailable-parser status; it only asserts that the parser is available in the current environment.

The PR description still says Qdrant was upgraded to `^1.19.0` for Node 26. Current `package.json` remains at `^1.17.0`, and the lockfile resolves `1.18.0`. The description must state the current diff.

The TSCN link in the README and source comment is a 404. The current documentation is [TSCN file format](https://docs.godotengine.org/en/stable/engine_details/file_formats/tscn.html).

The current CodeRabbit status is not a review of head `6565f6a`. Its report remains paused and identifies its reviewed delta as ending at `4a82206`; it also excluded `package-lock.json` and skipped every selected file in that delta: [CodeRabbit report](https://github.com/giancarloerra/SocratiCode/pull/130#issuecomment-5485068446).

Required correction:

- Classify native GDScript AST support as conditional, and make parser availability and failure status accurate.
- Correct the language-agnostic claim, the TSCN link, and the stale PR description.
- Request a full CodeRabbit review after the final commit and confirm that its reviewed commit is the final head.

## Newly identified in the current-head review

These were not part of the earlier requests. They are separated here so they are not presented as previously missed instructions.

### 1. `node-gyp-build` is resolved as an undeclared root dependency

The latest loader uses the module-level `esmRequire("node-gyp-build")`: [`code-graph.ts`](https://github.com/giancarloerra/SocratiCode/blob/6565f6a76b045922930429cffe7907e97c8d88e5/src/services/code-graph.ts#L606-L616). SocratiCode does not declare `node-gyp-build`; it is a dependency of `tree-sitter-gdscript`: [`package.json`](https://github.com/giancarloerra/SocratiCode/blob/6565f6a76b045922930429cffe7907e97c8d88e5/package.json#L65-L102), [`tree-sitter-gdscript package.json`](https://github.com/prestonknopp/tree-sitter-gdscript/blob/v6.1.0/package.json#L29-L32).

This works with npm's default hoisted layout only because the transitive package happens to be installed at the root. Node resolves a bare `require()` from the requiring module's directory and then its ancestors: [Node module resolution](https://nodejs.org/api/modules.html#loading-from-node_modules-folders). npm's supported `nested` and `linked` strategies do not hoist that package, and npm specifically recommends `linked` to detect undeclared phantom dependencies: [npm `install-strategy`](https://docs.npmjs.com/cli/v11/using-npm/config#install-strategy).

Required correction:

- Create a `require` rooted at the resolved `tree-sitter-gdscript/package.json`, then resolve its own `node-gyp-build` dependency from that package context. Do not rely on root hoisting.
- Add an install-and-loader regression using `npm ci --install-strategy=linked` or an equivalent isolated layout.

### 2. Valid relative and UID-based Godot paths are lost

The GDScript resolver handles only `res://` and `class:` specifiers: [`graph-resolution.ts`](https://github.com/giancarloerra/SocratiCode/blob/6565f6a76b045922930429cffe7907e97c8d88e5/src/services/graph-resolution.ts#L1684-L1713). Current extraction and resolution therefore miss or discard valid forms with different bases:

- `extends "base.gd"` is relative to the current script directory.
- `preload("helper.gd")` is relative to the current script directory.
- A relative `load("helper.gd")` is normalized by the current engine as `res://helper.gd`, relative to the project root.
- `uid://...` is accepted by current Godot resource loading, including scripts.

The first two forms are shown in the stable [GDScript reference](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_basics.html#inheritance). The exact engine handling is visible in [`extends`](https://github.com/godotengine/godot/blob/74e99170ee7292c759bbdf65027381fd166494a8/modules/gdscript/gdscript_analyzer.cpp#L438-L442), [`preload`](https://github.com/godotengine/godot/blob/74e99170ee7292c759bbdf65027381fd166494a8/modules/gdscript/gdscript_analyzer.cpp#L4795-L4815), and [`ResourceLoader` path normalization](https://github.com/godotengine/godot/blob/74e99170ee7292c759bbdf65027381fd166494a8/core/io/resource_loader.cpp#L689-L697).

TSCN external resources also carry both `uid` and `path`. Godot prefers a valid UID and falls back to the text path only when the UID cannot be resolved: [`resource_format_text.cpp`](https://github.com/godotengine/godot/blob/74e99170ee7292c759bbdf65027381fd166494a8/scene/resources/resource_format_text.cpp#L474-L499). The current extractor ignores `uid`, so it can follow a stale fallback path instead of the resource Godot loads.

Required correction:

- Preserve whether an import came from `extends`, `preload`, or `load`, then apply the correct base for each form.
- Build a project-scoped UID index from `.uid` sidecars and UID-bearing `.tscn` and `.tres` headers.
- Resolve `uid://` first and use the text path only as the documented TSCN fallback.
- Add tests for source-relative `extends` and `preload`, project-root-relative `load`, `../` normalization, script and resource UIDs, a valid UID with a stale path, and an invalid UID with a valid fallback path.

### 3. AST import extraction can invent or corrupt paths

For `load()` and `preload()`, the AST branch accepts any call whose text starts with the function name, then recursively finds any descendant string: [`graph-imports.ts`](https://github.com/giancarloerra/SocratiCode/blob/6565f6a76b045922930429cffe7907e97c8d88e5/src/services/graph-imports.ts#L373-L383). For example, `load(resolve_path("res://fake.gd"))` is not a literal load path, but the nested string is recorded as a dependency.

The same code obtains values with `.replace(/['"]/g, "")`. GDScript supports raw strings such as `r"res://x.gd"`, so the result becomes `rres://x.gd`; escape handling and triple-quoted strings are not decoded as GDScript string values. Raw strings are part of the stable [GDScript literal syntax](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_basics.html#literals).

Required correction:

- Inspect the direct call argument rather than an arbitrary descendant.
- Evaluate only statically supported string expressions and decode their actual GDScript value. Dynamic expressions must remain unresolved, not produce an edge from a nested string.
- Use the same syntax-aware literal handling for `extends`, `preload`, `load`, and the parser-unavailable path. The fallback must not scan comments and strings as executable code.
- Add raw, escaped, triple-quoted, nested-call, dynamic-expression, comment, and string-literal regressions.

### 4. `class_name` indexing and fallback extraction do not follow GDScript syntax

Both class indexes scan raw text with `^class_name\s+(\w+)`: [`graph-resolution.ts`](https://github.com/giancarloerra/SocratiCode/blob/6565f6a76b045922930429cffe7907e97c8d88e5/src/services/graph-resolution.ts#L1130-L1148), [`per-project index`](https://github.com/giancarloerra/SocratiCode/blob/6565f6a76b045922930429cffe7907e97c8d88e5/src/services/graph-resolution.ts#L1175-L1201). This produces both misses and false registrations:

- A valid same-line annotation such as `@abstract class_name AbstractClass` is missed.
- A `class_name Phantom` line inside a triple-quoted string is registered as a real class.
- Valid Unicode identifiers are missed because JavaScript `\w` is ASCII-only for this use.
- The regex fallback accepts only uppercase `extends` identifiers, although GDScript identifiers are case-sensitive and may begin with lowercase letters: [`graph-imports.ts`](https://github.com/giancarloerra/SocratiCode/blob/6565f6a76b045922930429cffe7907e97c8d88e5/src/services/graph-imports.ts#L193-L217).

Godot permits [same-line annotations](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_basics.html#annotations) and [most UAX #31 Unicode identifiers](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_basics.html#identifiers). The installed `tree-sitter-gdscript@6.1.0` grammar also limits identifiers to ASCII, so switching this index directly to that grammar does not solve Unicode correctness: [`grammar.js`](https://github.com/prestonknopp/tree-sitter-gdscript/blob/v6.1.0/grammar.js#L78-L83).

Required correction:

- Replace raw-source declaration regexes with a GDScript-aware tokenizer or parser that excludes comments and string bodies, handles annotations, and implements current identifier rules.
- Correct the fallback's identifier handling or disable unsafe class edges when exact parsing is unavailable.
- Address the upstream grammar's Unicode limitation for AST chunking, or remove the unconditional Full Support classification until the parser accepts Godot-valid identifiers.
- Add annotated declaration, lowercase class, Unicode class, and multiline-string false-declaration regressions.

### 5. The TSCN extractor rejects valid whitespace

The current regex requires `[` in column 1 and exactly `path="..."`: [`graph-imports.ts`](https://github.com/giancarloerra/SocratiCode/blob/6565f6a76b045922930429cffe7907e97c8d88e5/src/services/graph-imports.ts#L221-L242). Godot states that TSCN whitespace is insignificant outside strings and defines headings as `[<resource_type> key1=value1 ...]`: [TSCN entries](https://docs.godotengine.org/en/stable/engine_details/file_formats/tscn.html#entries-inside-the-file). Valid leading whitespace or `path = "material.tres"` is therefore missed.

Required correction:

- Replace the section-header regex with a small tokenizer that respects insignificant whitespace, quoted strings, escapes, comments, and arbitrary attribute order.
- Parse `uid` and `path` together for the UID-first behaviour above.
- Add leading-whitespace, spaces-around-`=`, attribute-order, quoted-value, escape, and comment regressions.

## Additional cleanup identified in the same diff

- Remove the unreachable second `case "gdscript"` in [`graph-imports.ts`](https://github.com/giancarloerra/SocratiCode/blob/6565f6a76b045922930429cffe7907e97c8d88e5/src/services/graph-imports.ts#L649-L691); the earlier GDScript branch always returns.
- Remove the stale `[instance]` reference in [`code-graph.ts`](https://github.com/giancarloerra/SocratiCode/blob/6565f6a76b045922930429cffe7907e97c8d88e5/src/services/code-graph.ts#L998-L1004).
- Compute each file's Godot root once before its import loop instead of once per import: [`code-graph.ts`](https://github.com/giancarloerra/SocratiCode/blob/6565f6a76b045922930429cffe7907e97c8d88e5/src/services/code-graph.ts#L1095-L1131).
- Remove or justify the legacy global `classNameIndex` path, which is not used by normal graph builds when Godot files exist: [`code-graph.ts`](https://github.com/giancarloerra/SocratiCode/blob/6565f6a76b045922930429cffe7907e97c8d88e5/src/services/code-graph.ts#L850-L862).
- Update the `graph-resolution.ts` service table, which still documents only the old global helpers: [`DEVELOPER.md`](https://github.com/giancarloerra/SocratiCode/blob/6565f6a76b045922930429cffe7907e97c8d88e5/DEVELOPER.md#L1000-L1018).

## Final validation required

1. Add the regressions listed above at the lowest useful unit level and through `buildCodeGraph` where integration behaviour is the requirement.
2. Run the complete Node 18, 20, and 22 suite.
3. Add native-loader coverage for packaged, source-built, missing, and invalid artifacts, plus at least one isolated non-hoisted install.
4. Run platform-specific checks for every native path still claimed as supported.
5. Update the PR description and documentation to match the final behaviour.
6. Request a full CodeRabbit review against the final commit and confirm the reviewed SHA.

JiminyJoo:
@coderabbitai resume

giancarloerra:
Hi @JiminyJoo. The other pending base changes are now merged, including #139, and current `main` is `a66be4d`.

PR #130 currently conflicts with `main`. Please rebase onto current `main` and complete the still-applicable items from the previous review. The latest CI rerun is green on Node 18, 20, and 22, including the Node 18.17 consumer-install check, but green CI does not clear the outstanding functional review findings.

The exact-head CodeRabbit review for `334508f` still has seven unresolved threads. The current code and behaviour items are:

- resolve `node-gyp-build` from the installed `tree-sitter-gdscript` package context rather than relying on root hoisting;
- use the nearest on-disk `project.godot` for nested resource-only projects;
- make native-parser tests conditional when the optional parser is unavailable and add fallback-path coverage;
- remove the unused global `classNameIndex` construction and duplicate `.gd` file read;
- make the GDScript support description accurately conditional on native-parser availability.

The supported-extension count and broken TSCN documentation link are minor documentation cleanups, not merge blockers.

After the rebase and fixes, please run the complete suite and request a full CodeRabbit review against the final head.

I'm ready to take this over and bring it over the line so we can close it, if easier. Let me know!

giancarloerra:
@coderabbitai review

giancarloerra:
Hi thanks, taking over this one now to fix the last few things and merge. Thanks for this and all the work you've done on it!

giancarloerra:
@coderabbitai review

giancarloerra:
@coderabbitai review

PR #133, opened 2026-09-01
fix(indexing): preserve effective index profiles

### LINKED ISSUE #129: Preserve existing indexes when index-shaping settings change
### Problem

Settings and code changes that alter embedded text, vector generation, chunk output, or stored payload identity are not recorded with the index. Incremental indexing decides whether to skip a file from a hash of its raw source content only. A setting can therefore change the data that should be stored while every unchanged file is skipped.

`EMBEDDING_DOCUMENT_PREFIX` from #126 is one concrete case. `prepareDocumentText` includes the configured prefix in the text sent to both dense embedding and BM25, while `hashContent` hashes only the raw file contents. Project metadata stores file hashes but no encoding or index-format profile. Context artifacts make a separate staleness decision from `contentHash` alone.

The failure mode is a mixed collection: unchanged files retain the old representation, while edited or newly added files use the new representation. Search then uses the current query configuration against data produced by more than one configuration.

Context artifacts also embed their configured path and store their path and description in payloads. Changing an artifact path or description without changing its content currently leaves the old vectors or payloads in place because only `contentHash` is compared.

### Backward-compatibility invariant

An existing index must remain usable after an upgrade without a mandatory re-index.

- A new index-shaping feature must either work safely with the existing index or remain inactive for that index until its next explicit fresh index.
- An upgrade must not silently activate new encoding, chunking, provider, model, or path-composition behavior for only part of an existing index.
- An incremental update, file-watcher update, or context-artifact refresh must continue using the existing index's effective behavior.
- Search must use the query-side behavior paired with the effective document and embedding behavior of the collection being searched.
- A changed setting may be reported as pending, but it must not make the existing index unusable and must not trigger an automatic full rebuild.

### Required design

Persist a versioned effective index profile with each code collection and context-artifact collection.

The profile must cover every value or feature gate that changes stored chunk text, vector generation, chunk boundaries, stored language, or the compatible query representation. The first implementation must include:

- `EMBEDDING_QUERY_PREFIX` and `EMBEDDING_DOCUMENT_PREFIX` from #126
- document-path inclusion from #127
- `MAX_CHUNK_CHARS` from #128
- effective embedding identity for fresh indexes: provider, model, dimensions, effective context length, and output-shaping options such as `LITELLM_SEND_DIMENSIONS`
- `EXTENSION_LANGUAGE_MAP`, because it changes language selection, AST chunking, and stored language without changing the raw source hash
- `MAX_FILE_SIZE_MB`, because lowering it currently skips scanning an oversized file while leaving that path in the current-file set, so its previously stored chunks are not deleted
- an index-format version so future index-shaping behavior can be added without changing existing collections

Requested runtime configuration and effective collection configuration are different concepts:

1. A fresh index records the current requested settings and latest index-format version as its effective profile.
2. An existing index continues using its stored effective profile for search, full/incremental skip decisions, changed-file embedding, watcher updates, and context-artifact refreshes.
3. If requested settings differ from the effective profile, status surfaces the difference as pending.
4. The pending profile becomes effective only when the user explicitly creates a fresh index.
5. Every future setting or feature that changes stored representation must be enrolled in this profile before release.

Linked-collection search must not reuse one query vector across collections with different effective query profiles. It must generate and reuse one vector per compatible effective query profile, then merge the results as it does today. Collections whose legacy embedding identity is unverified must not be assumed compatible with one another.

Each context-artifact state must also persist a configuration signature. The signature must include at least the configured path and description in addition to the content hash. A change to any of those values must replace that artifact's chunks and payloads using the context collection's effective profile.

### Legacy metadata migration

The latest published release is v1.12.0. It predates #126, #127, and #128. A collection with no effective profile must therefore be treated as the released legacy format for these new fields:

- query prefix: `search_query: `
- document prefix: `search_document: `
- document path included: `true`
- maximum chunk characters: `2000`

This migration must not re-embed or rewrite existing vector points. The inferred legacy profile may be persisted as metadata.

`EXTENSION_LANGUAGE_MAP`, `MAX_FILE_SIZE_MB`, and embedding configuration already existed in v1.12.0, but their historical values were not stored. The first profile-aware version may adopt the currently running values once and mark them as legacy-unverified. This preserves the adopted behavior from that point forward without claiming those values produced the existing points.

Do not represent an adopted legacy value as historically verified. In particular, a same-dimensional provider or model cannot be reconstructed from the current metadata or from the collection vector size alone.

### Both indexing subsystems are in scope

The code index and context artifacts make independent staleness decisions and persist independent metadata. The compatibility contract and tests must cover both. Fixing project `fileHashes` alone is insufficient.

### Acceptance tests

- Upgrade a legacy-profile code index, set all newly introduced options to non-default values, run an incremental update with no changed files, and verify that no stored point changes.
- Edit one file in that legacy index and verify that its replacement chunks still use the legacy effective profile rather than mixing in the pending profile.
- Search the legacy index after changing the requested query prefix and verify that the effective legacy query prefix is used.
- Persist an adopted legacy embedding identity as unverified, change the requested provider or model, and verify that later updates and searches continue using the adopted effective identity.
- Create a fresh index with the requested settings and verify that the new profile becomes effective.
- Repeat the unchanged, changed-content, and fresh-index cases for context artifacts.
- Change only a context artifact's path or description and verify that its vectors or payloads are replaced.
- Search linked collections with different effective query profiles and verify that each compatible profile receives its own query vector. Verify that unverified legacy identities are not grouped together.
- Verify that status reports requested settings as pending and identifies legacy-unverified fields without rejecting search or incremental indexing.
- Verify that removing and freshly indexing a collection activates the pending profile.
- Verify that changing `EXTENSION_LANGUAGE_MAP` or lowering `MAX_FILE_SIZE_MB` does not partially mutate a legacy collection before a fresh index.
- Verify that a file which grows beyond the effective `MAX_FILE_SIZE_MB` limit has its previously stored chunks and hash removed.

### Merge and release ordering

#127 and #128 are merged. Implement this issue against the combined shape of #126, #127, and #128.

The implementation of this issue must land before any release that exposes those index-shaping settings. The dependency is on release, not on the earlier feature pull requests.

giancarloerra:
## Summary

- persist versioned effective profiles for code and context collections
- resolve indexing, watcher updates, context refreshes, and search against each collection's effective profile
- retain the released prefix, path, and chunk behavior for legacy collections while marking unavailable historical fields as unverified
- isolate linked query embeddings by compatible verified profiles
- detect context artifact path and description drift
- remove stale chunks and hashes when files exceed the effective size limit
- preserve Node 18.17 compatibility in the production dependency graph

## Validation

- `npm run lint`
- `npx tsc --noEmit`
- `npm run build`
- `npm test`
- clean unit suites on Node 18.17.0, 20.19.5, and 22.22.0
- Node 18.17.0 packed consumer install with strict engine enforcement
- `coderabbit review --agent --light --include-untracked`
- `snyk test`

Closes #129

## Summary by CodeRabbit

* **New Features**
  * Added persisted indexing profiles for embedding settings, formatting, file limits, chunk sizes, and extension mappings.
  * Added profile compatibility for code, context, and linked searches, including legacy collections.
  * Added status reporting for pending profile changes and unverified legacy settings.
  * Improved context-artifact updates by replacing changed content and removing stale entries.
* **Bug Fixes**
  * Improved oversized-file handling, vector compatibility, provider readiness, and embedding dimensions.
  * Search and status checks no longer modify collections lacking profile metadata.
* **Documentation**
  * Clarified profile persistence and collection recreation requirements.


giancarloerra:
@coderabbitai full review

PR #155, opened 2026-09-07
fix(qdrant): support Node 26 without dropping Node 18

### LINKED ISSUE #154: [Bug]: 1.13.0 refuses to start on Node 26+ — pinned @qdrant/js-client-rest ~1.18.0 is incompatible
## Summary

Fresh install of the latest release (1.13.0) via `npx -y socraticode@latest` on Node 26+ refuses to start. It exits immediately with:

```
socraticode: Node 26.8.1 needs @qdrant/js-client-rest 1.19 or newer (installed: 1.18.0).
  Clients before 1.19 bundle an undici incompatible with Node 26+.
  Reinstall socraticode (or update its lockfile) to pick up a current client,
  or use Node 22.x (via nvm: `nvm install 22 && nvm use 22`).
  See https://github.com/qdrant/qdrant-js/issues/134.
```

## Environment

- Node: 26.8.1 (reproduces on any Node 26+)
- Install: `npx -y socraticode@latest` (resolves to 1.13.0)
- Qdrant: external (any — the failure is at startup, before the first Qdrant call)

## Root cause

`package.json` pins `@qdrant/js-client-rest: ~1.18.0`. That range only resolves to 1.18.x.

The upstream incompatibility is qdrant/qdrant-js#134: clients before 1.19 bundle undici v6 and hand its `Agent` to Node's built-in `fetch()`. On Node 26 the built-in `fetch` uses undici v8, which strictly validates the `onError` hook and rejects the v6 Agent on the first dispatch (`UND_ERR_INVALID_ARG: invalid onError method`).

A fix was released in `@qdrant/js-client-rest` **1.19.0** (undici bumped v6 → v7). See the closing comments on qdrant/qdrant-js#134.

## Impact

1. **1.13.0 is unusable on Node 26+.** The new startup check (correct in principle) reads the installed qdrant client version, sees 1.18.x, and aborts. Because the dependency range `~1.18.0` can never resolve to 1.19.x, this fires on *every* Node 26+ startup — the shipped release cannot work on Node 26+.
2. **1.12.0 is not a working fallback either.** It has no startup check, so it appears to start, but it pins the same broken client (`^1.17.0`, no 1.19) and fails silently at the first Qdrant call. Pinning to 1.12.0 masks the crash rather than fixing it.

## Suggested fix

Raise the dependency range so it can resolve to 1.19.x:

```
"@qdrant/js-client-rest": "~1.19.0"   // or "^1.19.0"
```

and refresh the lockfile. That's a one-line change plus a lockfile bump.

## Note on the startup check

The version check itself is a good idea — it converts a confusing first-request crash into a clear message. The problem is only that the pinned range can never satisfy it. With the range bumped, the check becomes genuinely advisory.

giancarloerra:
## Summary

- keep `@qdrant/js-client-rest` 1.18 so Node 18.17 remains supported
- on Node 26+, pair Qdrant's per-request dispatcher with fetch from the same undici line
- leave unrelated fetch traffic and unaffected runtime/client pairs on native fetch
- add clean packaged-consumer Qdrant smoke coverage for Node 18.17 and Node 26
- add Node 26 to the unit-test matrix

## Compatibility

Qdrant client 1.19 requires Node 22 or newer, so upgrading it cannot fix Node 26 without dropping existing Node 18 support. The bridge applies only to Node 26 or newer with a Qdrant client older than 1.19. It does not change index data or metadata and requires no reindex.

The consumer smoke tests install the generated tarball with engine checks and assert that its installed Qdrant client is 1.18 before performing a real Qdrant request. This prevents a substituted dependency from hiding a failure in the package that users actually install.

Fixes #154

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Qdrant connectivity across supported Node.js and client-version combinations.
  * Qdrant requests now select the appropriate transport when compatibility adjustments are needed.
  * Improved handling of invalid, unknown, and prerelease Qdrant client versions.
  * Startup no longer exits solely because a Qdrant client version is incompatible or unrecognized.

* **Tests**
  * Added Node.js 26 compatibility coverage.
  * Expanded packaged-installation, CLI, readiness, timeout, and Qdrant integration testing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [18, 20, 22]
+        node-version: [18, 20, 22, 26]
     steps:
       - uses: actions/checkout@v7
       - uses: actions/setup-node@v7
@@ -41,6 +41,11 @@ jobs:
   node-floor-consumer-install:
     name: Node 18.17 Consumer Install
     runs-on: ubuntu-latest
+    services:
+      qdrant:
+        image: qdrant/qdrant:v1.17.0
+        ports:
+          - 6333:6333
     steps:
       - uses: actions/checkout@v7
       - uses: actions/setup-node@v7
@@ -53,6 +58,16 @@ jobs:
         shell: bash
         run: |
           set -euo pipefail
+          qdrant_ready=false
+          for attempt in {1..30}; do
+            if curl -fsS http://127.0.0.1:6333/healthz >/dev/null; then
+              qdrant_ready=true
+              break
+            fi
+            sleep 1
+          done
+          test "$qdrant_ready" = true
+          repo_dir="$PWD"
           consumer_dir="$(mktemp -d)"
           npm pack --pack-destination "$consumer_dir"
           package_file="$(find "$consumer_dir" -maxdepth 1 -type f -name 'socraticode-*.tgz' -print -quit)"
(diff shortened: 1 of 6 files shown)


giancarloerra:
@coderabbitai full review

giancarloerra:
@coderabbitai full review

giancarloerra:
@coderabbitai full review

giancarloerra:
@coderabbitai full review

PR #161, opened 2026-09-09
fix(graph): stop the Python sibling-flat fallback fabricating edges inside packages

### LINKED ISSUE #157: [Bug]: Python sibling-flat fallback fabricates edges inside packages — self-edges and false import cycles
## Describe the bug

The Python sibling-flat fallback in `resolveImport` (`graph-resolution.ts`, `case "python"` — shipped at `dist/services/graph-resolution.js:2310-2325`) is applied to every absolute import that fails the earlier probes, including imports made from inside a package. Where a first-party module shares a name with a third-party distribution, the import resolves to the sibling file instead of the third-party package, fabricating an edge CPython would never create.

A file importing the third-party package it is named after resolves **to itself**, producing a self-edge. `codebase_graph_circular` reports these as import cycles.

The fallback's comment gives the justification — `sys.path[0]` is the script's own directory — and that is correct for a file run as a script (`python main.py`, #46). It does not hold for a module inside a package: CPython never puts a package module's directory on `sys.path`, so `import requests` from within a package resolves through `sys.path` to site-packages.

Opposite direction from #107: that fix addressed missing edges; this fabricates them.

## Steps to reproduce

Against the shipped resolver, no index required:

```js
import { resolveImport } from 'socraticode/dist/services/graph-resolution.js';

const PROJECT = '/repo';
const call = (spec, from, files, roots = []) =>
  resolveImport(spec, `${PROJECT}/${from}`, PROJECT, files, 'python',
    undefined, undefined, undefined, undefined, undefined, undefined,
    roots, undefined, undefined, undefined, undefined, false,
    undefined, undefined, undefined, undefined, undefined);

// A. regular package
const pkg = new Set(['src/pkg/__init__.py', 'src/pkg/client.py', 'src/pkg/requests.py']);
call('requests', 'src/pkg/client.py',   pkg);  // -> src/pkg/requests.py
call('requests', 'src/pkg/requests.py', pkg);  // -> src/pkg/requests.py   (self-edge)
call('numpy',    'src/pkg/client.py',   pkg);  // -> null                  (control)

// B. PEP 420 namespace package, no __init__.py anywhere, 'src' a declared root
const ns = new Set(['src/ns/sub/mod.py', 'src/ns/sub/requests.py']);
call('requests', 'src/ns/sub/mod.py',      ns, ['src']);  // -> src/ns/sub/requests.py
call('requests', 'src/ns/sub/requests.py', ns, ['src']);  // -> itself

// C. #46's own layout — the fallback is CORRECT here, yet:
const svc = new Set(['service-a/main.py', 'service-a/config.py', 'service-a/requests.py']);
call('config',   'service-a/main.py',     svc);  // -> service-a/config.py  (correct, keep)
call('requests', 'service-a/requests.py', svc);  // -> itself               (self-edge)
```

## Expected behavior

Every `requests` call returns `null`; `config` in C still returns `service-a/config.py`. The `numpy` control shows the correct answer where no colliding sibling exists.

## Actual behavior

Each `requests` call returns the sibling, twice resolving a file as its own dependency.

In a real repo (753 files, 2711 edges, graph built by 1.13.2) this produces four of six reported "circular dependencies", all false: two self-loops, and one pair reported twice because two import statements yield the same fabricated edge. Two first-party modules are named after declared dependencies (`whisperx`, `faster_whisper`), each wrapping the package it is named after. The other two chains are legitimate `TYPE_CHECKING` back-edges.

## Suggested direction

Two changes, orthogonal rather than ranked:

**1. Reject self-resolution.** If the resolved path equals the source file, return `null`. No Python module imports itself, so this is unconditionally safe. It is not a lesser version of (2): case C above shows a self-edge arising in *#46's own layout*, where the sibling fallback is legitimate and must stay. That case is unreachable by any package-detection gate, so (1) is the only change that covers every layout.

**2. Skip the sibling fallback when `<sourceDir>/__init__.py` is in the `fileSet`.** A directory with `__init__.py` is a regular package, so its modules are never `sys.path[0]`. This matches the CPython semantics the fallback's comment already appeals to, and preserves #46, whose layout is a runnable service directory with no `__init__.py`:

| fixture | `__init__.py`? | `import config` from `main.py` | under the gate |
|---|---|---|---|
| `service-a/{main,config}.py` (#46) | no | `service-a/config.py` | unchanged |
| `src/pkg/{client,requests}.py` | yes | `src/pkg/requests.py` | `null` |

**On PEP 420 (case B):** the `__init__.py` gate does not cover namespace packages, and I do not think a purely structural test can. The obvious candidate — skip the fallback when `sourceDir` sits below a declared import root — regresses #46. `buildPythonManifests` returns `"."` among the roots for any project with a root `pyproject.toml` (verified: a src-layout project yields `["src", ".", …]`), so `service-a/` is "below a declared root" too, the sibling step is skipped, and `import config` then finds nothing. Requiring `sourceDir` to *be* a declared root fails the same way whenever the services carry no per-service manifest.

The honest position is that a directory with neither `__init__.py` nor a manifest is genuinely indistinguishable from a script directory, and defaulting to today's behaviour there is defensible. That is precisely why (1) matters independently: it removes the self-edges in namespace packages even though (2) cannot reach them. Matching module names against declared dependencies would cover the rest, but distribution and import names are not reliably related — the Pillow/PIL point you made on #107 — so I have not proposed it.

**Side effect:** a fabricated edge counts as *resolved* in the 1.13.0 import-resolution advisory, so this slightly flatters the resolved-over-captured ratio.

## Environment

- socraticode **1.13.2**
- Node **v25.9.0**, Ubuntu 24.04.4
- Embedding provider: Ollama (native / external), `nomic-embed-text`
- Qdrant mode: Managed (Docker)
- MCP host: Other — Claude Code

## Checklist

- [x] I have searched existing issues and this hasn't been reported before — #46 introduced the fallback, #107 addressed under-resolution; neither covers this.
- [x] I am using the latest version of SocratiCode (1.13.2).

gregoryfoster:
## Summary

The Python sibling-flat fallback (#46) was applied to every absolute import that fell past the earlier probes, including imports made from inside a package. Where a first-party module shares a name with a third-party distribution, the import resolved to the sibling instead of the installed package, and a module named after the package it wraps resolved to itself. `codebase_graph_circular` reported those as import cycles.

Scope agreed in #157: remove the unambiguous regular-package collision and the self-edge cases, preserve #46 and #112, and pin the non-self PEP 420 case as an explicit structural ambiguity.

## Changes

- **No edge from a file to itself**, on all five probes — relative, project-root, `src`/`lib`, sibling, and manifest roots. The relative branch matters most: `from . import x` in a package's own `__init__.py` resolves the bare `.` to that same file, the commonest self-edge in any Python tree.
- **A self match at a probe standing for a real `sys.path` entry ends resolution** rather than falling through. Falling through traded a self-edge for a wrong one — `src/mod.py` → `lib/mod.py`, and `packages/pkg-a/src/requests.py` → `packages/pkg-b/src/requests.py`, a fabricated cross-package edge. Only the sibling probe still falls through, because it alone is a guess at a directory rather than a path entry.
- **The sibling probe is skipped when any ancestor up to the project root holds `__init__.py`.** A file inside a package tree is reached as a dotted module, so the flat guess is not what CPython does — verified directly against CPython for the namespace-subdirectory case (`src/pkg/vendor/mod.py` under `src/pkg/__init__.py` is reached as `pkg.vendor.mod`, and `import requests` there loads the installed distribution).
- **PEP 420 namespace packages are deliberately unchanged**, pinned by a test with the reasoning: no `__init__.py` to test, and the alternative discriminator ("sourceDir below a declared import root") also fires on `packages/pkg-a/src/pkg_a` and would invert `prefers the sibling-flat guess over a manifest root`. Such a directory is indistinguishable from the runnable script directory #46 serves.

### Known cost, accepted deliberately

`sys.path[0]` is the *script's* directory whether or not it holds `__init__.py`, so `python service-a/main.py` genuinely imports `service-a/config.py` even when `service-a/__init__.py` exists — and that edge is now dropped. Pinned by its own test rather than left in a comment. A name collision with a dependency cannot be falsified from the tree, while a package directory at least says the file is normally reached as `pkg.client`; the false edge is judged the costlier of the two. Discussed in #157.

## Type of change

## Testing

- [x] Unit tests pass (`npm run test:unit`) — 2053 across 82 files
- [x] Integration tests pass (`npm run test:integration`) — 172 across 12 files
- [x] New tests added for new/changed functionality — 11 in a `sibling-flat fallback scope (#157)` describe, each verified failing before the fix

No pre-existing test expectation was altered; the only deletion is a duplicated `pyResolve` helper hoisted to the outer describe. Behavior also checked against a 753-file Python repo: exactly 4 edges removed, all of them the target defect, zero legitimate edges lost.

One integration file (`manual-indexing-mode`) failed at suite level on an earlier run with zero failed assertions, and passed both in isolation and on a clean re-run — order-dependent, and untouched by this change. Flagging in case you see it too.

## Checklist

- [ ] I have updated documentation (README.md / DEVELOPER.md) if needed — no public API change
- [ ] I have addressed all [CodeRabbit](https://coderabbit.ai) review comments — pending first review

## Related issues

Fixes #157

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented Python imports from resolving back to the file that issued the import.
  * Preserved valid sibling-module links, including in package directories and namespace subdirectories.
  * Improved resolution consistency across project roots, source directories, and manifest-defined roots.
  * Ensured discarded self-matches do not incorrectly fall through to unrelated resolution paths.

### DIFF
--- a/src/services/graph-resolution.ts
+++ b/src/services/graph-resolution.ts
@@ -2600,9 +2600,17 @@ export function resolveImport(
       // Absolute: foo.bar.baz → foo/bar/baz.py or foo/bar/baz/__init__.py
       const modulePath = moduleSpecifier.replace(/\./g, "/");
 
-      const direct = notSelf(
-        resolveRelativePath(modulePath, projectPath, projectPath, fileSet, [".py"]),
-      );
+      // A self match HERE ends resolution rather than falling through. The
+      // project root is ahead of `src/`, `lib/` and every manifest root for a
+      // file that sits at the root — it is that file's own `sys.path[0]` — so
+      // nothing a later probe finds is a module CPython could reach past the
+      // source file itself. Root `mod.py` doing `import mod` beside a
+      // `src/mod.py` imports itself; answering with `src/mod.py` trades the
+      // self-edge for an edge to an unrelated module, which is worse. The
+      // later probes still fall through, because there the discarded match is
+      // a guess (the sibling probe) or one root among several.
+      const direct = resolveRelativePath(modulePath, projectPath, projectPath, fileSet, [".py"]);
+      if (direct === relSourceFile) return null;
       if (direct) return direct;
 
       // Try common Python source directories (src layout)
@@ -2630,14 +2638,25 @@ export function resolveImport(
       // root both offer the module, the sibling is what actually gets
       // imported.
       //
-      // Only for a file that could BE `sys.path[0]` (#157). A directory
-      // holding `__init__.py` is a regular package, and CPython never puts a
-      // package's own directory on `sys.path`: `import requests` from
-      // `src/pkg/client.py` is resolved through `sys.path` to the installed
-      // distribution, so a sibling `src/pkg/requests.py` is not what runs.
-      // Without this gate any first-party module sharing a name with a
-      // dependency invents an edge — the graph-wide version of the self-edge
-      // the guard above removes.
+      // Only for a directory that is not itself a regular package (#157). A
+      // directory holding `__init__.py` is imported as a package, and every
+      // documented way of importing one — installed distribution, `-m`, a
+      // path entry above it — resolves `import requests` from
+      // `src/pkg/client.py` through `sys.path` to the installed distribution,
+      // so a sibling `src/pkg/requests.py` is not what runs. Without this gate
+      // any first-party module sharing a name with a dependency invents an
+      // edge — the graph-wide version of the self-edge the guard above
+      // removes.
+      //
+      // The gate is a trade, not a law: `sys.path[0]` is the *script's*
+      // directory whether or not it holds `__init__.py`, so `python
+      // service-a/main.py` really does import a sibling `service-a/config.py`
+      // even when `service-a/__init__.py` exists, and that edge is dropped
+      // here. A name collision with a dependency is unfalsifiable from the
+      // tree, while a package directory at least says the file is normally
+      // reached as `pkg.client`; the false edge is judged the costlier of the
+      // two. The #46 layout the fallback serves — a runnable directory with no
+      // `__init__.py` — is unaffected.
       //
       // Deliberately NOT extended to PEP 420 namespace packages, which have
       // no `__init__.py` to test: `src/ns/sub/mod.py` importing `requests`
--- a/src/services/graph-resolution.ts
+++ b/src/services/graph-resolution.ts
@@ -2648,68 +2648,26 @@ export function resolveImport(
       // root both offer the module, the sibling is what actually gets
       // imported.
       //
-      // Only for a directory that is not itself a regular package (#157). A
-      // directory holding `__init__.py` is imported as a package, and every
-      // documented way of importing one — installed distribution, `-m`, a
-      // path entry above it — resolves `import requests` from
-      // `src/pkg/client.py` through `sys.path` to the installed distribution,
-      // so a sibling `src/pkg/requests.py` is not what runs. Without this gate
-      // any first-party module sharing a name with a dependency invents an
-      // edge — the graph-wide version of the self-edge the guard above
-      // removes.
+      // A NON-SELF collision here is left alone, deliberately (#157). Where a
+      // first-party module shares a name with a third-party distribution —
+      // `src/pkg/requests.py` beside `src/pkg/client.py` doing `import
+      // requests` — which one CPython loads depends on how the file is run,
+      // and the tree does not say. Imported as `pkg.client`, `sys.path` finds
+      // the installed distribution; executed as `python src/pkg/client.py`,
+      // `sys.path[0]` is `src/pkg` and the sibling wins. An `__init__.py` does
+      // not settle it: `sys.path[0]` is the SCRIPT's directory whether or not
+      // the directory is also a package, so gating on one would drop the
+      // legitimate #46 edge from a directly executed service directory that
+      // happens to contain `__init__.py`. The same ambiguity covers PEP 420
+      // namespace packages, which have no `__init__.py` to test at all.
       //
-      // The gate is a trade, not a law: `sys.path[0]` is the *script's*
-      // directory whether or not it holds `__init__.py`, so `python
-      // service-a/main.py` really does import a sibling `service-a/config.py`
-      // even when `service-a/__init__.py` exists, and that edge is dropped
-      // here. A name collision with a dependency is unfalsifiable from the
-      // tree, while a package directory at least says the file is normally
-      // reached as `pkg.client`; the false edge is judged the costlier of the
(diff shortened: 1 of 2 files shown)


giancarloerra:
Thanks for explicitly verifying the direct-script counterexample. That evidence changes the conclusion: the `inPackage` gate cannot be merged because it deliberately drops a legitimate #46-style sibling import whenever a directly executed service directory also contains `__init__.py`. The tree does not reveal whether the file is run directly or imported as a package module, so treating that loss as an accepted default would be a backward-compatibility regression.

Please revise #161 to:

- remove the `inPackage` gate and the regression that records its lost direct-script edge as an accepted cost;
- retain the unambiguous self-resolution guards on all probes, including preventing a self match at a real import root from falling through to an unrelated root;
- preserve every existing non-self sibling-fallback case from #46 and #112, including a directly executed service directory that contains `__init__.py`;
- leave the non-self regular-package and PEP 420 collisions documented as execution-mode ambiguities for now, rather than changing default behavior on an inference the current resolver cannot prove;
- keep the separate graph-wide self-loop suggestion out of this PR. An explicit self-import such as `src/foo.ts` importing `./foo` is a real graph cycle, not the same defect as a resolver fabricating a self-edge. Filtering every self-loop in `findCircularDependencies` would hide genuine source errors across all languages and change existing behavior. Any wider fix should instead begin with a concrete resolver-generated false edge in another language.

That leaves #161 as a focused, backward-compatible correction of the false self-edges without replacing them with fabricated cross-root edges or removing previously valid edges.

gregoryfoster:
Done in `19b94ac`.

- `inPackage` gate removed, along with the test recording its lost direct-script edge and the two asserting a non-self collision resolves to null. The `#107` describe is back to its original contents.
- Self guards kept on all five probes, including no fall-through from a self match at a real import root to an unrelated one.
- The regular-package collision, the same collision in a namespace subdirectory, and the PEP 420 case are now pinned as execution-mode ambiguities — each resolves to the sibling, as before this branch.
- Added a test for the layout the gate would have broken: a service directory containing `__init__.py` still resolves its sibling import.
- Self-loop suggestion left out, and your framing is right — an explicit `import "./foo"` is a genuine cycle, not a resolver artifact. I'll leave a wider fix until there's a concrete resolver-generated false edge in another language to anchor it.

No pre-existing expectation is altered; the only deletion left in the test diff is a duplicated helper hoisted to the outer describe. `tsc`, `biome` and 2051 unit tests green.

One review catch of my own while revising: the comment justifying the sibling probe's fall-through used an unreachable example (`src/pkg/requests.py` → `src/requests.py`, which the `src`/`lib` probe would have returned first). It now names the manifest-root case the pinning test actually exercises.

Issue #120, opened 2026-08-28
[Bug]: PHP: resolve DIR/bare requires and mapless-autoloader use imports; stamp the persisted graph with its builder version


gregoryfoster:
### Describe the bug

On a PHP 8.4 project the dependency graph resolves **zero PHP edges** across 1124 PHP files. Every edge in the graph (119) comes from the 113 JavaScript/CSS files in the same index.

PHP **parsing and symbol extraction are fine** — `codebase_symbols` returns complete, correctly-numbered symbol lists for the exact files the graph cannot connect. Only import resolution produces nothing.

The consequence is worse than an empty result, because two tools do not report ignorance — they emit a confident negative. For a class referenced by 41 `use` statements, `codebase_impact` answers *"Total impacted files: 0 — No callers found — nothing else depends on this."* Its own description says to run it before refactoring, renaming or deleting, so the failure mode is indistinguishable from a genuine "safe to delete".

**Not a duplicate of #43.** That fix is present. #43's signature was *zero symbols* (0 → 2,301 after fix) from an atomic `registerDynamicLanguage` failure; here symbol extraction works fully, on v1.12.0 versus #43's v1.8.3. This looks like the same family as #45 (Go module paths), #106 (Dart `package:`) and #107 (Python src layout): per-language import resolution not implemented for PHP.

### Steps to reproduce

Any PHP project with a standard composer PSR-4 map reproduces this. Minimal shape:

1. `composer.json` declares a conventional map:
   ```json
   { "autoload": { "psr-4": { "App\\": "app/" } } }
   ```
2. `app/View/Composers/FrontPage.php` contains:
   ```php
   namespace App\View\Composers;

   use App\Support\WeekRange;   // defined at app/Support/WeekRange.php
   ```
3. Run `codebase_index`, then `codebase_graph_build`, and wait for both to complete.
4. Confirm the file parsed: `codebase_symbols file=app/View/Composers/FrontPage.php`
5. Ask for its dependencies: `codebase_graph_query filePath=app/View/Composers/FrontPage.php`
6. Ask for the target's callers: `codebase_symbol name=WeekRange`

### Expected behavior

Step 5 lists an import edge to `app/Support/WeekRange.php`, and step 6 lists `FrontPage.php` as a caller of `WeekRange` — matching how the same tools behave for JavaScript in the same index.

### Actual behavior

Step 4 succeeds — the parser has no trouble with the file:

```
[...]

PR #169, opened 2026-09-12
perf(graph): rebuild only when a graph input changed

### LINKED ISSUE #163: [Feature]: Incremental update rebuilds the whole code graph for changes that cannot affect it
**Area:** Indexing, Code graph

## Problem or motivation

Every incremental update rebuilds the entire code graph whenever any indexed file
changed. The gate in `updateProjectIndex` is:

```ts
if (added > 0 || updated > 0 || removed > 0) {
  const graph = await rebuildGraph(resolvedPath, { skipSymbolGraph: false });
}
```

Measured by calling `rebuildGraph` directly against a throwaway Qdrant v1.17.0
container on a spare port, timing only that call (embeddings not involved):

| repo | files | edges | `rebuildGraph` |
|---|---|---|---|
| small service | 132 | 291 | 1.5s |
| mid-size app | 2,713 | 6,297 | **15.9s** |
| large app | 3,872 | 12,822 | **26.8s** |

Roughly 7ms/file, near-linear.

A git hook fires an update per commit, so a commit touching only files the graph
is not built from — a README, a JSON fixture, a Flyway migration, a k8s
manifest — pays that in full and produces the same graph it already had.

This is negligible below ~1k files. It is worth raising only because the cost
scales with repository size and is paid per commit.

## Why this is not a one-line change

I tried the obvious predicate in #160 and it was correctly rejected. The code
graph admits inputs that `getAstGrepLang` does not answer for: Elixir templates,
Godot `.tscn`/`.tres` resources, configured extra extensions, and
`EXTENSION_LANGUAGE_MAP` entries. Gating on it would have silently stopped
rebuilding for real source files — a stale graph, which is worse than a
redundant rebuild.

The general problem is that **there is no single authority for "does this file
contribute to the code graph?"** Any predicate written outside `code-graph.ts`
will drift from the set of inputs `rebuildGraph` actually consumes, and the
drift is silent.

So this is an ownership question before it is a performance one, which is why I
am opening an issue rather than a PR.

## Proposed solution

Three approaches, no strong preference — the choice is really yours:

1. **Derive it from the build.** Have `rebuildGraph` report the files it
   consumed — it already walks them — and persist that set with the graph. The
   next update intersects changed + removed against it. No second predicate
   exists, so nothing can drift.

2. **Own the predicate in `code-graph.ts`.** Export something like
   `graphAcceptsFile(relativePath, profile)` answering for every input class,
   and have the indexer ask it. Cheaper than (1), but predicate and builder then
   have to be kept in step by review rather than by construction.

3. **Finish incremental symbol-graph updates.** `updateChangedFilesSymbolGraph`
   currently returns `fullRebuildRequired: true` for any change, with a comment
   deferring cross-file caller re-resolution. Completing that makes gating
   largely moot, since the work becomes proportional to the change. Much larger
   piece of work.

(1) is the only one that cannot drift. (3) is the better end state.

## Alternatives considered

- **Gate on `getAstGrepLang` alone** — rejected in #160, see above.
- **Debounce or coalesce rebuilds across rapid commits** — reduces frequency,
  not cost, and delays graph freshness.
- **`SOCRATICODE_WATCHER=off`** — sidesteps the automation rather than fixing
  the cost.

## Additional context

This was reverted out of #160 at your request; CodeRabbit has persisted the
constraint about graph inputs as a repo learning, so the trap is at least
recorded.

Happy to implement whichever approach you prefer, or to leave it if the cost
does not justify the change for your users.

mgd43b:
## Summary

Implements approach 1 from #163: `rebuildGraph` reports the inputs it consumed, they are persisted beside the graph, and the next incremental update skips the rebuild when the change is proven to touch none of them.

Every incremental update currently rebuilds the whole graph whenever any indexed file changed, so a commit touching only a README, a fixture, a migration or a manifest the resolver never reads pays a full rebuild and produces the graph it already had — ~7ms/file, so 15.9s on a mid-size app and 26.8s on a large one, per commit.

There is no predicate describing which files a graph is built from, and none is added here. The build records each input at the point that opens it, so a file stops being recorded in the same commit that stops it being read.

> Updated after the first review round — the record gained an `unreadable` bucket, environment roots are recorded, and the decision is no longer gated on the index's counters. See [this comment](https://github.com/giancarloerra/SocratiCode/pull/169#issuecomment-5647015595) for what changed and why.

## The input inventory

Not a list to keep in step — the set of recording call sites *is* the inventory. The buckets mirror the kinds of read the build makes, because each kind can only be watched the way it was made:

| the build | recorded as | watched by |
|---|---|---|
| read a file whole | content hash | re-hashing it |
| head-read one to classify it | head hash | re-reading that head |
| used it without opening it | its size | a stat |
| tried to open it and failed | its path | trying again |
| listed a directory | its entry names | listing it again |

**Files read whole** — every source the build loop parses; the ignore files `createIgnoreFilter` read (root `.gitignore`, every nested one, `.socraticodeignore`, carried out of the filter with the bytes it was built from); `tsconfig.json`/`jsconfig.json` and the whole `extends` chain, including a base config outside the project root; every `go.mod`, `pubspec.yaml`, `pyproject.toml`, `Cargo.toml`, `composer.json`; `project.godot` and every `.uid` sidecar; the sources read for the namespace / FQCN / module / `class_name` maps. Most of those are never graph nodes, which is why a record of nodes would not do.

**Head-read and turned away** — an extensionless file detection rejected. Only the head was read, so only the head is watched: a whole-file hash would describe bytes the decision never saw, and a size would miss a same-length edit that flips the answer (`# hello!!` becoming `#!/bin/sh`).

**Used without being read** — a leaf node made from its path alone, a file too large for the graph. Recorded as the size the build saw, since size is what carries a file across the limit in either direction.

**Tried and failed** — a manifest the walk found and could not read, a `tsconfig.json` that exists but is unreadable, a source that vanished mid-walk, an extensionless file whose head-read threw. There is no size the build ever saw and no content to hash; the only question that matters is whether it can be read *now*, because a file that starts being readable starts contributing.

**Directory listings** — what makes an addition detectable at all. The other buckets describe files that existed when the build ran, so none can speak for a path that did not, and the index cannot speak for one it does not hold: `go.mod` (`.mod` is in neither `SUPPORTED_EXTENSIONS` nor `SPECIAL_FILES`), `project.godot`, `.uid` sidecars, and every ignore file unless `INCLUDE_DOT_FILES` is set. Adding a nested `go.mod` to split a Go module changes no indexed file and every Go import under it.

Environment roots are listed too, though the walk never enters them. A directory is an environment because of a `pyvenv.cfg` or `conda-meta` *inside* it, so deleting that marker re-admits a whole subtree while the parent's listing does not move — `createIgnoreFilter` now reports the roots it found for exactly this reason.

**Not files at all**, folded into one `settings` hash: the effective extra extensions, `EXTENSION_LANGUAGE_MAP`, `INDEX_EXTENSIONLESS`, `INCLUDE_DOT_FILES`, `RESPECT_GITIGNORE`, `MAX_GRAPH_FILE_BYTES`.

## The metadata shape

One new payload field on the graph's existing metadata point, stored as JSON beside `graphData`:

```json
{
  "version": 1,
  "builtByVersion": "1.13.3",
  "files":       { "src/index.ts": "…", "go.mod": "…", ".gitignore": "…" },
  "heads":       { "NOTES": "…" },
  "presence":    { "src/huge.ts": 1234567 },
  "unreadable":  [ "jsconfig.json" ],
  "directories": { ".": { "all": "…", "kept": "…" }, "src": { "all": "…", "kept": "…" } },
  "settings":    "…"
}
```

Backward compatible the same way `importCount` and `builtByVersion` already are: graphs persisted before it simply omit it. Absent, unparsable, or carrying a `version` this build does not recognise are read identically — **one full rebuild, which writes the current shape**. No migration, no re-index, no reset, nothing to run.

It is read back projected to that one field. The point also holds the serialized graph — 5.6MB on a 3,941-file repository — and every incremental update asks this question, so pulling the graph across to answer it would cost more than the rebuild it avoids.

## Each listing is hashed twice, and the cheap one is asked first

Comparing every entry name needs nothing but a `readdir`. Comparing only the entries the ignore filter kept needs the filter, and building one costs a synchronous walk plus an `ignore` test per entry — measured at **2.1s of the 2.2s** a check took on a 3,941-file repository, against **5ms** for the listings themselves.

So `all` settles the common case and `kept` decides, computed only for a directory whose contents actually moved. Without `kept` a `.DS_Store` appearing, or a coverage directory a test run writes and cleans, would rebuild the graph and leave the skip theoretical.

## When the decision is asked

Whether or not the index itself moved. The two sets are not the same — a commit touching only `go.mod` or `.gitignore` has every index counter at zero and still leaves the graph wrong — so gating on those counters would leave the whole mechanism unreachable for exactly the inputs it goes to the trouble of recording.

It is gated instead on a record having been found. Without one the answer is "rebuild" on no evidence at all, so a project that has never been built under this scheme keeps the old trigger: the first update that changes anything is what rebuilds and writes the record.

## What still rebuilds

- **any addition the index saw**, and any file that appeared or vanished where the walk looked
- any settings change, and a graph built by a **different SocratiCode version** — a persisted graph is served unchanged across upgrades, so one cut before a resolver shipped keeps answering as if that resolver did not exist (#120). Compared on its own rather than folded into the settings hash, so the rebuild is legible in the log and is one rule to remove if you would rather not have it
- any recorded input removed, changed, gone from disk, changed size, that became readable, or that the index **tried to read this run and could not** — the index keeps the hash it already had for those, so the gate would otherwise read that as proof they had not changed
- a graph with no usable record, when the index also saw a change
- a decision that cannot be reached at all — if anything throws, the caller does exactly what it did before this gate existed

The skip is reached only after checking every recorded input and finding all of them unchanged.

## Cost

The index has already hashed everything it holds, so a recorded input the index covers is answered from that work rather than read again — `hashContent` moves to `constants.ts` so the two are the same hash.

| repository | graph nodes | rebuild | decision, using the index's hashes | decision, reading every input from disk |
|---|---|---|---|---|
| this one | 179 | 1.5s | <1ms | 12ms |
| a mid-size app | 2,836 | 12.0s | 7ms | 82ms |
| a large app | 3,941 | 18.9s | **8ms** | 173ms |

The record costs 5–10% of the graph payload it sits beside (12 KB against 237 KB here; 435 KB against 5.6 MB on the large app). The build itself does not regress: the walk now applies the ignore filter once, through the same function the check re-reads a directory with, rather than filtering its own entries a second time.

## Changes

- `graph-inputs.ts` (new) — the record, the recorder, and the decision
- the build records at each point that opens a file: `code-graph.ts`, `graph-resolution.ts`, `graph-aliases.ts`
- `createIgnoreFilter` reports the ignore files it read, with the bytes it was built from, and the environment roots it found
- `resolveExtensionlessDetection` reports the head its answer was scored on, and whether the read failed
- `saveGraphData` persists the record; `loadGraphInputs` reads it back projected
- the incremental update asks `shouldRebuildGraph` before rebuilding, and reports which files it could not read

## Type of change

## Testing

- [x] Unit tests pass (`npm run test:unit`) — 2,156
- [x] Integration tests pass (`npm run test:integration`) — 194

`tests/unit/graph-input-invalidation.test.ts` builds a real graph over a fixture and drives the decision: source, manifest (with nothing from the index to report it, so the decision has to read the file), ignore file, extensionless head, a virtualenv marker deleted, configuration including `INDEX_EXTENSIONLESS`, extra-extension set, addition, a file the index never holds appearing, removal, an input gone from disk, one the index could not read, one that became readable, a too-large file that stopped being too large, and every legacy-metadata case — absent, unknown version, another builder version, malformed, and the persisted string form. Two skips are asserted explicitly, including that an *ignored* file appearing does not rebuild, so the feature cannot quietly disappear while the rest stays green. Four tests in `index-profile-indexer.test.ts` pin the wiring, including both halves of the trigger condition, and one in `qdrant.test.ts` pins the round-trip through a real Qdrant.

Every one of those was checked by mutation — inverting the decision, dropping the `go.mod` recording, dropping the directory recording, defanging either tier of the directory check, dropping the environment-root recording, dropping the unreadable check, restoring the counter guard, and ignoring `fromRecord` each fail exactly the tests that name them.

## Checklist

- [ ] I have addressed all [CodeRabbit](https://coderabbit.ai) review comments — CodeRabbit skipped this PR (`review-ready` label not set, which I cannot add); Copilot's review is addressed in full

## Related issues

Closes #163

No incremental symbol-graph work, and no `getAstGrepLang` predicate — there is no predicate at all.

## Summary by CodeRabbit

* **New Features**
  * Code graph updates can skip rebuilding when tracked inputs remain unchanged.
  * Rebuilds detect changes to files, directories, configuration, parser capabilities, and previously unreadable inputs.
  * Graph input metadata is stored alongside graph data for reliable incremental updates.
  * Extensionless file detection reports unreadable files more accurately.
  * Graph builds retry when inputs change during processing and preserve the previous graph if consistency cannot be achieved.

* **Documentation**
  * Updated documentation to explain when graph rebuilds are skipped or triggered.

* **Tests**
  * Added coverage for incremental rebuild decisions, persistence, and build consistency.


mgd43b:
Thanks — that review found a real hole, and one of its findings turned out to be sharper than it looked. All eight substantive points are addressed; two I've pushed back on with reasoning rather than code.

**The critical one was right, and it was mine.** The decision sat inside `if (added > 0 || updated > 0 || removed > 0)`, so a commit touching only `go.mod`, `.gitignore` or `project.godot` — every index counter at zero — never reached it. That is exactly the class of input this goes to the trouble of recording, so the mechanism was unreachable for its own headline case. The gate is now asked whether or not the index moved, and gated instead on a record having been found: without one there is nothing to argue from, so a project that has never been built under this scheme keeps the old trigger and the first update that changes anything rebuilds and writes the record. Both halves of that condition are pinned by their own tests.

**Five findings collapsed into one mechanism.** Failed manifest reads (7 sites), failed `.uid` sidecar reads, failed config reads, failed extensionless detection, and `-1` conflating "unreadable" with "path-only leaf" were all the same gap: a file the build *tried* to open and could not was either unrecorded or filed next to files it deliberately never opened. There is now a separate `unreadable` bucket, watched by the only question that matters for it — whether it can be read now, since a file that starts being readable starts contributing. `presence` goes back to meaning just "used without being opened", watched by size. The recorder's precedence does the rest: `read` supersedes `unreadable`, so a catch that fires after a successful read is a no-op and each call site is one unconditional line.

**`INDEX_EXTENSIONLESS`** is in the settings hash now. Same reasoning as the others already there — it moves the node set with nothing on disk changing.

**The environment-marker finding was the sharpest, and my first fix for it didn't work.** I wrote the regression expecting the directory listings to cover it, and it failed: `backend/env/` is excluded by a `pyvenv.cfg` *inside* it, the walk never enters it, and the parent's listing doesn't move when the marker goes — so the cheap tier sees nothing and the filtered tier is never reached. `createIgnoreFilter` now reports the environment roots it found alongside the ignore files it read, and the walk lists each root, so a deleted marker is an entry-name change like any other. Mutation-checked: drop the environment recording and that test is the one that fails.

Two I have not changed:

**Ignored directory-entry changes not persisted after a skip** — correct, and the cost is worth naming: once an ignored file appears in a walked directory, its unfiltered hash stays stale, so every later decision builds the ignore filter (~0.6s on a 3,941-file repo) instead of stopping at the cheap tier. Fixing it means writing the record on a skip, which turns a read-only decision into a Qdrant write on every no-op update. The staleness clears itself on the next real rebuild. Happy to change it if you'd rather pay the write.

**`extraExtensions` not passed to `rebuildGraph`** — pre-existing on both call sites (`indexer.ts:1445` and the update path), unchanged here, and the decision deliberately uses the same default the rebuild uses, so the two cannot disagree and no spurious rebuild comes of it. The real divergence is between the extras the *index* admits and the extras the *graph* is built with, which predates this and would change graph content for any caller passing them — worth its own issue rather than a rider on this one, unless you'd like it here.

Re-measured after all of the above, 3,941-file repository: rebuild 18.9s, decision 8ms against the index's hashes, 173ms reading every recorded input from disk. Suite is 2,156 unit + 194 integration, green. Each new behaviour is mutation-checked — restoring the counter guard, ignoring `fromRecord`, and dropping the environment recording each fail exactly the test that names them.

giancarloerra:
Thanks. I checked the current `b21c193` head. The general design is the right one, and CI is green, but five bounded points still need addressing before this is ready for the hosted review:

1. **Legacy or unusable graph-input metadata is still skipped on a no-op update.** `decideGraphRebuild()` returns `rebuild: true, fromRecord: false` for an absent, malformed or unknown-version record, but `updateProjectIndex()` then suppresses that rebuild when the code-index counters are zero. That contradicts the agreed “rebuild once and populate” compatibility rule and the README. An existing graph with no usable record must rebuild once even when no indexable file changed. Please distinguish “no graph exists” if necessary, and do not collapse non-not-found Qdrant read failures into a silent missing-record result. Add zero-counter regressions for legacy, malformed/unknown and metadata-read-failure paths.

2. **Please include `extraExtensions` in this PR.** It is directly part of the new invalidation contract, not a separate optimization. Pass the same effective set into the initial/full graph build, `shouldRebuildGraph()`, and every automatic incremental rebuild. Otherwise a graph built with per-call extensions can be compared or rebuilt under the defaults and lose those leaf nodes. Add a regression proving a custom-extension leaf survives an update without a false configuration mismatch.

3. **A failed recursive directory listing is still not recorded.** If `walk()` cannot `readdir()` a kept subtree, it returns before recording any input for that directory. If the directory later becomes listable while its parent entry is unchanged, the graph can remain stale. Record unreadable directories as a distinct input, re-check them with `readdir()` rather than a file-readability test, and add the recovery regression.

4. **Refresh the cheap directory-listing cache after an ignored-only change.** When `all` changes but `kept` does not, the decision can safely skip, but it leaves the old `all` hash persisted. Every later no-op then rebuilds the expensive ignore filter. This requires one metadata refresh when an ignored-only directory change is actually observed, not a Qdrant write on every no-op. Persist the refreshed listing under the existing update lock and add a test showing the next decision remains on the cheap tier.

5. **Persist the effective parser capability used by the build.** Optional/native grammar availability can differ across supported runtimes or hosts at the same SocratiCode version, and it changes import/symbol extraction without changing the current settings hash. Include a stable capability fingerprint, such as the successfully loaded grammar names and relevant parser modes rather than error text, and rebuild conservatively when it changes. Add a regression for a grammar becoming available at the same version.

Please keep the changes limited to these invalidation and compatibility paths. Once the updated head is stable and CI is terminal, I will run the full CodeRabbit review.

mgd43b:
All five done at `0553b4f`. Two of them I had wrong in ways worth naming.

**1 — legacy metadata skipped on a no-op update.** You're right, and it contradicted the rule I'd quoted back at you. The gate is no longer keyed on "was a record found" but on **whether a graph exists at all**, which is the only case that should stay suppressed: with nothing persisted there is no graph to refresh, and building a first one is not an update's decision to make. Everything else rebuilds however quiet the update was.

`loadGraphInputs` now returns three outcomes instead of collapsing them:

| | means | on a zero-counter update |
|---|---|---|
| `absent` | no metadata point — no graph | no rebuild |
| `stored` | a graph is there; the record may be missing, malformed or of an unknown version | **rebuild once and populate** |
| `failed` | the read itself failed | **rebuild** — nothing is known, least of all that there is no record |

A not-found collection still reads as `absent`; every other Qdrant error is `failed` and carries its message into the reason so it cannot be mistaken for the legacy path. Zero-counter regressions for all of legacy, malformed, unknown-version and read-failure, at both the decision layer (`graph-rebuild-load-outcomes.test.ts`, which mocks the read) and the indexer's gate.

**2 — `extraExtensions`.** In, and you were right that it belongs here: the set was already in the settings hash, so passing it in one place and not the other was the invalidation contract disagreeing with itself. It now reaches the full build (`indexer.ts:1445`, which never passed it), `shouldRebuildGraph`, and the incremental rebuild. Regression builds a graph under `{".md"}`, asserts the leaf is recorded, and asserts that deciding under the same set is a skip rather than a configuration mismatch.

**3 — failed recursive listing.** Recorded as `unreadableDirectories`, distinct from the file-level `unreadable`, and re-checked with `readdir` rather than a readability test, as you asked. The parent's listing holds the directory's name either way, so nothing else would have moved. Regression chmods a subtree to `000`, builds, restores it, and expects `graph input directory became listable`.

**4 — refresh the cheap listing cache.** The decision now hands back a refreshed record when — and only when — a directory's `all` moved while its `kept` did not. The indexer writes it once at that point, under the lock the update already holds, via `setPayload` so the graph half of the same point is untouched. No write on an ordinary no-op. Two tests: the decision hands one back and the next decision against it stays on the cheap tier (observable as no second refresh); and the indexer calls the writer exactly once. The mocked lock in the indexer suite was missing `holdsProjectLock` entirely, so the ownership guard was throwing into the non-fatal catch rather than answering — fixed, which is what made that path testable at all.

**5 — parser capability.** Its own `capabilities` field, hashed from the loaded grammar names and the GDScript mode. Not the error text, for the reason you gave: those carry host-specific paths and would move for reasons unrelated to what the build could parse. Compared before the version check so the reason reads `the parsers available to the build changed` rather than being mistaken for an upgrade. Regression asserts a grammar set differing at the same version rebuilds.

Re-measured, 3,941-file repository: rebuild 17.1s, decision **9ms** against the index's hashes, 128ms reading every recorded input from disk. 2,166 unit and 195 integration, green. The four new behaviours are mutation-checked — defanging the capability check, emptying `unreadableDirectories`, restoring the counter guard, and suppressing the refresh write each fail exactly the tests that name them.

One consequence worth flagging before the hosted review: `capabilities` is a required field, so every record written by the previous head fails to parse and is read as legacy. That is one rebuild per project on first run, which is the behaviour you specified for missing metadata — but it is a rebuild that happens on upgrade to this head, not just on a change.

giancarloerra:
Thanks. The five requested areas are substantially addressed at `0553b4f`, but the verification exposed three remaining holes in those same fixes:

1. **The parser-capability fingerprint still omits the HEEx/EEx parser mode.** `buildCodeGraph()` awaits `ensureElixirTemplateParsers()`, and both import and symbol extraction return different results when that WASM parser is unavailable, but `currentGraphCapabilities()` records only the ast-grep grammar names and GDScript mode. Capture the boolean result used by the build and by the update decision, include it in `graphCapabilitiesHash()`, and extend the same-version capability regression. Otherwise a graph can still be reused after its effective parser capability changes.

2. **The special environment-root listing still uses the file bucket on failure.** In `getGraphableFiles()`, the `ig.environments` loop catches a failed `readdir()` with `recorder.unreadable(absolute)`, while only the recursive `walk()` catch uses `unreadableDirectory()`. Route both failed directory listings through `unreadableDirectory()` and cover the environment-root path in the recovery regression, so it is re-tested with `readdir()` as required.

3. **The ignored-entry cache refresh can overwrite the canonical input record with a stale one.** The decision returns a copy of the old `graphInputs`, and `refreshGraphInputs()` later replaces that field. A concurrent explicit graph rebuild does not use the index lock, and lock ownership can also be lost after `holdsProjectLock()` but before the asynchronous `setPayload()`. If the newer graph is saved first and this refresh lands second, its input record is replaced by the older graph's record, which can make later updates skip incorrectly. The refresh also suppresses every Qdrant failure and reports success.

   Keep the canonical `graphInputs` immutable except when the graph itself is saved. If the persistent cheap-tier optimization is retained, store it as a separate advisory cache keyed to an immutable graph-record generation or fingerprint, and apply it conditionally so a stale generation cannot modify a newer one. A mismatched cache must be ignored. If that conditional cache is not worth the added mechanism, remove the persistence refresh and accept the repeated ignore-filter check; that is slower but correct. In either case, do not silently swallow a failed requested cache write.

These are not additional feature scope. They complete the parser, unreadable-directory, and cache-refresh fixes already requested. Please hold further additions after these so the next stable green head can go directly to the full CodeRabbit review.

mgd43b:
All three at `b718263`. I took the second option you offered on the cache, and one of my own regressions turned out to be vacuous.

**1 — HEEx/EEx in the capability fingerprint.** In. `currentGraphCapabilities()` is now async and awaits `ensureElixirTemplateParsers()`, so the build and the decision resolve it the same way by construction rather than by both remembering to. I resolve it unconditionally instead of only where the project has templates: making it conditional means the two sides have to agree on *when* to ask, which is another thing that can drift, and the loader memoises its own promise — measured at 10ms on first load and 0ms after, once per process. The regression pins both halves: that the hash moves with the flag, and that the live fingerprint is built from the flag rather than from the grammar names alone.

**2 — environment-root listing failure.** Routed through `unreadableDirectory()`, so both failed listings are re-checked with `readdir` as you asked.

The regression for it was wrong on the first attempt, and it is worth saying how. I chmod'd the environment root to `0o000`, and the test passed — then passed again with the fix mutated out. At `0o000` the `pyvenv.cfg` inside is unstattable, so the filter stops calling the directory an environment, the *walk* tries to enter it, and the walk's catch records it. Right bucket, wrong path, no coverage of the line under test. It is `0o111` now — executable but not readable, so the marker still stats and the directory is still an environment, while the `readdir` that loop makes fails. Mutating the fix now fails that test.

**3 — the ignored-entry cache refresh.** Removed, rather than made conditional. Your analysis of the race is right and it is worse than a lost optimisation: `refreshGraphInputs()` replaced a field the graph save owns, so a refresh landing after a concurrent explicit rebuild would put an older graph's inputs on a newer graph, and the next update would skip against them. The ownership check could not close it either — the lock can be lost between the check and the `setPayload`, and an explicit rebuild does not take that lock at all.

I went with the second option you offered. A generation-keyed advisory cache would work, but it means a second persisted field, a fingerprint to key it by, merge logic, and a conditional write — new mechanism in the exact area that has now produced three defects, on a PR you want to freeze for review. The record is immutable except when the graph is saved, `refreshGraphInputs()` is gone, and with it the swallowed write failure.

The cost is stated in the code where the decision gives it up: an ignored entry appearing after the last build leaves the cheap hash stale, so each later decision pays one ignore-filter build — ~0.6s on a 3,941-file repository — until the next real rebuild. Still an order of magnitude under the 17.3s rebuild it replaces, and self-healing. If you would rather have the conditional cache, it is a clean follow-up against a record shape that will not need to change for it.

Re-measured, 3,941-file repository: rebuild 17.3s, decision **10ms** against the index's hashes, 129ms reading every recorded input from disk. 2,167 unit and 194 integration, green. The two new behaviours are mutation-checked: dropping the template mode from the fingerprint, and routing the environment-root failure back into the file bucket, each fail exactly the test that names them.

Holding here. No further additions.

giancarloerra:
The full CodeRabbit review has now completed against `b718263`. The previous three requested fixes are correct, but the review found one real consistency blocker in the same graph-input mechanism:

1. **A graph build can persist a mixed snapshot when the same source is read more than once.** Resolver setup reads some source files before the main graph loop reads them again. `GraphInputRecorder.read()` currently replaces the earlier hash with the later one, and `unreadable()` ignores a later failure when an earlier read or head exists. If a file changes or becomes unreadable between those reads, resolver output can reflect the first observation while nodes, imports or symbols reflect the second, and the persisted record describes only the later observation. A later update can then treat that mixed graph as current and skip rebuilding it.

   Please make repeated observations consistency-checked rather than last-write-wins. Keep the first observation, mark incompatible later observations as a conflict, and make `finish()` throw a typed transient-change error. The graph rebuild should retry from a fresh scan a bounded number of times and must not replace either the cached or persisted graph unless one complete attempt is internally consistent. If files keep moving after the bounded retry, surface the non-fatal graph-build failure and retain the previous graph. The expected extensionless `head` to full-file transition must remain valid when the full content has the same exact `DETECT_HEAD_BYTES` byte window; identical repeated reads must also remain valid. Please add regressions for different whole-file reads, read then unreadable, identical repeated reads, the unchanged extensionless head-to-full transition, and the bounded retry/no-persist behaviour.

2. **Complete the `extraExtensions` wiring regression.** Pass a non-default extension set to `updateProjectIndex()` and assert that the exact same set reaches `shouldRebuildGraph()`. This pins the propagation bug already fixed in this PR.

3. **Correct the README sentence.** It should say that any **graph-relevant** addition triggers a rebuild, since ignored additions intentionally do not.

I am treating CodeRabbit's newline-containing filename serialization note as non-blocking and do not want that theoretical edge added to this pass.

Please keep the next revision limited to these three points and stop there so CI and the final review can run against a stable head.

mgd43b:
All three at `f98cb8b`. One place where I have followed your intent rather than your wording, flagged below so you can overrule it.

**1 — mixed snapshots from repeated reads.** Consistency-checked now, not last-write-wins. The first observation stands; a later one that contradicts it becomes a conflict; `finish()` throws `GraphInputsMovedDuringBuild`, carrying the path and what disagreed. `unreadable()` after a read or head is a conflict too, for the reason you gave — it is the same mixed snapshot the other way round.

The rebuild retries from a fresh scan up to three times, and neither the cache nor the persisted record is touched until an attempt comes back consistent. That meant removing the `graphCache.delete()` that ran *before* the build: clearing it up front is what would have cost a project its graph on giving up, which is the opposite of retaining the previous one. On exhaustion it rethrows, which the incremental update already reports as a non-fatal graph-build failure. A failure that is not files moving comes straight out on the first attempt — a real defect should not get three goes and then a shrug.

**Where I have diverged.** You asked me to keep the first observation. For the head→full transition I keep the *full-content* hash instead, and treat the transition as a refinement rather than a contradiction. Keeping the head would leave the record watching an 8KB window of a file the build parsed whole, so any change past byte 8192 would go unnoticed by the very mechanism this PR exists to provide — the full hash subsumes the window, since a change inside the window changes it too. The validity check is exactly as you specified: the transition is accepted only when the full content's `DETECT_HEAD_BYTES` window is byte-identical to the head that was scored, tested at the exact boundary with content continuing past it. Say the word and I will keep the head instead.

I also extracted `detectionHeadWindow()` out of `detectExtensionFromSource()` rather than computing that window a second time, so the recorder scores the same bytes detection does instead of holding a second opinion about where the window ends.

Regressions for all five you listed: different whole-file reads, read then unreadable, identical repeated reads, the unchanged head-to-full transition, and the bounded retry with no persist — the last in `graph-build-consistency.test.ts`, which also pins that the previous graph is still served after a give-up, and that a non-transient failure neither retries nor is swallowed.

**2 — `extraExtensions` wiring regression.** In. A non-default set goes into `updateProjectIndex()` and the test asserts the same object — identity, not equality — reaches `shouldRebuildGraph()`, and that it reaches `rebuildGraph()` too.

**3 — README.** Corrected to "any graph-relevant addition", naming what is excluded: a file appearing where the build looked, but not one the ignore rules exclude.

Nothing else touched. The newline-in-filename serialization stays as it is, per your note.

Re-measured, 3,941-file repository: rebuild 19.5s, decision **9ms** against the index's hashes, 145ms reading every recorded input from disk. The consistency check costs one extra head-window hash per file read whole, which is inside the noise on the build. 2,178 unit and 194 integration, green.

Five mutations, each failing exactly the test that names it: reverting `read()` to last-write-wins, dropping the head-to-full validation, ignoring `unreadable()` after a read, cutting the retry to a single attempt, restoring the pre-build cache clear, and forwarding `undefined` extensions instead of the caller's.

Stopping here.

giancarloerra:
@coderabbitai review

giancarloerra:
The exact-head CodeRabbit review completed against `f98cb8b`. The requested mixed-snapshot, `extraExtensions`, and README fixes are correct, but one new finding is valid in the changed consistency path:

- `readFileHead()` records the UTF-8 decoding of the first raw `DETECT_HEAD_BYTES`, while the later whole-file observation is compared through `detectionHeadWindow()`, which decodes and re-encodes a different representation when the input contains an invalid UTF-8 byte. An unchanged extensionless file can therefore be reported as moving during every attempt and exhaust all three graph-build retries. Make both observations use the same canonical representation, preferably by normalising the discovery head through `detectionHeadWindow()` before it is returned or recorded, and add an unchanged invalid-UTF-8 regression that would currently exhaust the retries.

The repeated newline-filename finding remains deliberately non-blocking, and the JSDoc placement note is not a merge blocker. Please keep the correction limited to the extensionless-head mismatch. Once the push is finished and CI is terminal, trigger `@coderabbitai review` for that final head and stop until it completes.

mgd43b:
Fixed at `cb480de`. The finding is real, and verifying it turned up something about the existing suite worth knowing.

**The mismatch, measured.** A file whose head holds one invalid UTF-8 byte, padded past the window:

| window taken from | characters |
|---|---|
| raw bytes on disk (`readFileHead`) | 8192 |
| already-decoded content (`detectionHeadWindow`) | 8190 |

U+FFFD stands in for the invalid byte and costs three bytes on the way back out, so the decoded side's window ends earlier. Hashes differed, so an unchanged file conflicted on every attempt.

`readFileHead()` now returns the canonical window — your preferred shape, normalised before it is returned, so there is one representation rather than two call sites that must remember to agree. Detection is indifferent to it: the window is idempotent under itself, which the regression asserts directly rather than leaving to inspection. `detectionHeadWindow()` takes `maxBytes` now so the disk reader normalises with its own budget rather than assuming the default, since it accepts one.

**What that broke, and why it is worth mentioning.** `extensionless.test.ts` has a test whose anti-vacuity line was `expect(detectExtensionlessExtension(await readFileHead(p))).toBe(".sh")` — it proved the two windows genuinely diverge for its fixture by borrowing `readFileHead`'s unnormalised output. Normalising made that line false, which is the correct outcome: there is no longer an unnormalised window in production to borrow. It now derives the raw window from the buffer itself and keeps the same guarantee, plus an assertion that the disk reader agrees with the canonical window rather than diverging from it. The test's intent is unchanged; only its source of raw bytes moved.

**Regressions.** `readFileHead` and `detectionHeadWindow` agree on invalid-UTF-8 content at the window boundary; the normalisation does not move detection's answer; and at the recorder, a head-then-whole-file observation of that content seals without conflict — the case that would have exhausted the retries.

One honest qualification on "would currently exhaust the retries": with the current call sites it could not yet. `recorder.head()` is reached from exactly one place, the discovery branch for extensionless files it *rejects*, and a rejected file never enters the set the build loop reads — so head-then-read does not happen in a real build today. The comparison was unsound regardless, the contract in its own docblock claimed otherwise, and the first call site that records a head for an accepted file would have hit it as a permanent build failure. I have pinned it at the recorder rather than writing a build-level test that pretends the path exists.

2,181 unit and 194 integration, green; CI terminal on all nine checks. Mutation-checked: reverting the normalisation fails all three of the tests that name it, including the pre-existing one.

Triggering the review below, and stopping there.

mgd43b:
@coderabbitai review

PR #170, opened 2026-09-12
feat(index): reclaim stored identities explicitly

### LINKED ISSUE #167: [Feature]: No way to reclaim a project index after its directory is gone
## Summary

There is no supported way to reclaim a project's index once its directory is
gone. `removeProjectIndex` is complete and correct — I verified that below —
but nothing ever calls it on its own, and `projectPath`, which is already
stored for every entry, is never re-read to ask whether the directory still
exists.

On one developer machine this had accumulated to **2,660 collections for 882
project identities, of which only 16 corresponded to a directory that still
existed**. Removing the rest by hand took storage from 2.1 GB to 1.4 GB,
resident Qdrant memory from 1.14 GiB to 601 MiB, and container CPU from ~16%
to 1.7%. The shared `socraticode_metadata` collection holds 1,188 points in
50 MB, of which 1,164 name a `projectPath` that no longer exists.

## What already works — please don't fix this part

I first suspected a leak in the removal path and could not reproduce one.
A lifecycle probe against **1.13.3**, driving the MCP server over stdio:

```
index  /Users/.../probe   →  +4 collections  (<id>_symgraph_{meta,file,index},
                                              codebase_<id>)
                          →  +2 socraticode_metadata points
remove /Users/.../probe   →  all 4 collections gone, both points gone
                          →  residue: 0
```

`removeProjectIndex` calls both `deleteCollection` and `deleteProjectMetadata`,
and the pairing with `saveProjectMetadata` (indexer.ts, 7 call sites) is
symmetric. The surviving integration test cleans up in `afterAll`. Every stale
row I hold was written by 1.12.0 by test fixtures whose files no longer exist
upstream. **I have no evidence that 1.13.3 leaks anything.**

## The gap

The accumulation comes from directories that disappear *outside* the tool —
`git worktree remove`, a deleted clone, a renamed folder — after which
`codebase_remove` can no longer be invoked for them in the normal way. Because
`coreProjectId` is the SHA-256 of the absolute path, a machine that uses
worktrees mints a new identity per directory and retires none.

`.socraticode.json` with a committed `projectId` prevents the duplication going
forward and is well documented. It does not reclaim what is already stored.

## Proposed shape

A `codebase_prune` tool that reads what the store already knows:

- **Reports candidates by default; deletes only on an explicit apply.**
- Lists each stored identity with its `projectPath`, `lastBuiltAt`,
  `builtByVersion`, and its associated collections.
- **Distinguishes a missing path from an inaccessible one.** A permission
  error, an unmounted volume or a disconnected network share must never
  authorize deletion. Only a path that resolves and is absent becomes a
  candidate — and a candidate is not proof of abandonment.
- Targets the identity as inventoried, rather than recomputing it from a
  directory that is gone: recomputation would lose a pinned `projectId`.
- Reuses `removeProjectIndex` / `deleteProjectMetadata` for the actual
  deletion, so the retirement semantics stay in one place.
- Is idempotent: a second apply does nothing.

## Deliberately not proposed here

- **A startup sweep behind an env flag.** It inherits every deletion ambiguity
  and runs before an operator has seen the candidates. Worth considering only
  after explicit pruning has proven itself.
- **Deriving the default identity from the repository** rather than the
  absolute path. That addresses duplication, not retirement, and needs defined
  behaviour for divergent worktrees, clones, migrations and pinned ids. It
  belongs in its own proposal.
- **Making `deleteSymbolGraphData` call `deleteProjectMetadata` for symmetry.**
  The absent call alone does not establish a defect, and I have not measured
  which metadata that function owns.

## Open questions for you

1. Does reclamation belong in the product, and is an MCP tool the right
   interface for it — or would you rather it were a CLI/maintenance path?
2. What deletion semantics do you want for a shared store with more than one
   client indexing into it? I have not measured that topology.

Happy to implement it from SIRTHEO/SocratiCode once you've settled the shape.
I'd rather agree on the scope first than send a PR you have to redirect.

## Environment

socraticode 1.13.3 (mechanism identical in 1.12.0), Qdrant in Docker,
macOS 26.6 / Apple Silicon, Node 22.

SIRTHEO:
## Summary

- Add `codebase_prune`, a report-only inventory of stored identities, their collections and every metadata record.
- Deletion takes the exact identity, a confirmation token from a fresh inventory, and `acknowledgeNoRemoteWriters: true`; it is revalidated under an identity-scoped barrier immediately before the delete.
- Path states are neutral (`present-on-this-host`, `absent-on-this-host`, `unknown/inaccessible`); no candidate set is derived from them and there is no blanket apply.

## Safety

- Only resources under the configured `QDRANT_COLLECTION_PREFIX` are inventoried or deleted; an unprefixed lookalike is neither.
- The confirmation token is a sorted per-point fingerprint (point id, collectionName, projectPath, indexingStatus, timestamps, builder version) over every record of the identity; deletion is refused while any record reports `in-progress`.
- A reclamation barrier (`prune`, `index` and `watch` locks by identity, plus an in-process set) is held from final validation through deletion. Indexing, updates, the watcher, graph builds and context indexing check it before they start, and treat a lock that cannot be inspected as held. A lock inspection error refuses the delete.
- Locks live on this host: `acknowledgeNoRemoteWriters` states, on the operator's behalf, that no other host sharing the store is writing. Documented as a limitation.
- Not-found during deletion counts as deleted; every other failure is reported per resource. Metadata points are deleted with `wait: true`, and success is reported only when the re-read inventory shows nothing left for the identity; otherwise the leftovers are listed.
- Graph cache, loaded file hashes and the symbol-graph cache are invalidated by identity, so a pinned identity loaded through another checkout is not served from cache after its collections are gone.
- Metadata points that cannot be attributed to an identity are returned as a list with point id, safe fields and the reason, and are never deleted. Same-canonical-path identities are advisory only. Existing indexes need no migration or re-index.

## Tests

- `npx vitest run tests/unit/remove-tools.test.ts tests/unit/list-collections-metadata-scroll.test.ts tests/unit/reclamation-barrier.test.ts tests/unit/reclamation-writers.test.ts tests/unit/graph-cache-identity.test.ts`
- `npx vitest run tests/integration/prune-reclamation.test.ts` against a real Qdrant under an isolated prefix: creates code, graph, symbol, context and metadata resources for two identities plus a foreign-prefix installation and an unprefixed lookalike; checks the report, refuses a stale token and an unacknowledged apply, deletes exactly one identity, verifies the others untouched, and repeats the apply.
- `npx tsc --noEmit`, `npm run lint`, `npm run build`
- `npm test`: 2370 passed, 78 skipped, 3 failed: the three GDScript tests in `graph-gdscript.test.ts` that need the `tree-sitter-gdscript` native package, which is not installed here, as with the previous head. Docker, Qdrant and Ollama were up, so the Docker-gated integration suites ran rather than skipped.

## Docs

README (“Reclaiming stored identities”), the management tool reference, the codebase-management skill and DEVELOPER.md describe the report/apply flow, path-state meaning, confirmation requirements, shared-store limitation, partial-failure behaviour and idempotent retry.

Closes #167

## Summary by CodeRabbit

- **New Features**
  - Added the `codebase_prune` tool to inventory stored project identities and reclaim selected resources.
  - Supports report-only previews and confirmed cleanup using an exact identity, fresh confirmation token, and shared-store acknowledgement.
  - Reports per-resource outcomes and withholds ambiguous resources for manual inspection.
  - Prevents indexing, watching, graph building, and artifact updates during reclamation.
  - Clears related cached data after cleanup.

- **Documentation**
  - Added `codebase_prune` guidance to the README and codebase-management references.

- **Tests**
  - Added comprehensive unit and integration coverage for reclamation workflows.


giancarloerra:
@coderabbitai review

giancarloerra:
CI is green and the full CodeRabbit review has completed against `ddafbbe`. The overall product shape is right, but this destructive path is not ready to merge. Please make one bounded correction pass covering the following points:

1. **Exclude foreign collections before parsing identities.** CodeRabbit is correct that the symbol-graph branch slices `QDRANT_COLLECTION_PREFIX` without first checking the prefix. Reject every resource name that does not begin with the configured prefix, then parse the prefix-stripped name. Add a non-empty-prefix regression proving that an unprefixed lookalike is neither inventoried nor deleted.

2. **Make the confirmation token cover every relevant metadata record.** The inventory currently collapses multiple metadata points into one `lastIndexedAt`, `lastBuiltAt` and `builtByVersion`, with the last scrolled value winning. A change to another record can therefore leave the token unchanged. Retain a deterministic, sorted per-point fingerprint containing the point ID and relevant projected fields, including `collectionName`, `projectPath`, `indexingStatus`, timestamps and builder version. Include every record in the token, and refuse deletion when any record reports `in-progress`. Add a regression with code and context metadata proving that changing either record invalidates the token.

3. **Serialize pruning with every writer and fail closed.** The current lock checks have a check-then-delete race, omit graph and standalone context-artifact writes, and `isProjectIdentityLocked()` treats a lock inspection failure as unlocked. Use an identity-scoped exclusion barrier held from final inventory validation through deletion, and make every indexing, watcher, graph-build and context-indexing writer respect it. A lock-status error must refuse deletion. Local filesystem locks cannot prove that another host sharing Qdrant is idle, so require an explicit acknowledgement that no remote writer is active and document that limitation unless a real cross-host lease is implemented. Add regressions for a lock-check error, a writer starting after the former final check, and graph/context activity.

4. **Make deletion genuinely idempotent and wait for metadata removal.** CodeRabbit is correct that not-found during either deletion means the resource is already gone and should count as `deleted`; all other failures must remain explicit. Pass `wait: true` to the Qdrant metadata-point deletion before immediately re-reading the inventory, then verify the exact resources are absent. Add not-found and delayed-delete regressions.

5. **Invalidate in-memory state by identity, not only by the stored path.** A pinned identity can be loaded through another local checkout path. Invalidating only `entry.projectPath` can leave that graph cached after its backing collections are deleted, and the identity-keyed index hash cache is not cleared by this path. Add identity-based invalidation for the graph and index hash caches, retain the symbol cache reset, and cover a pinned identity used through a different local path.

6. **Return actionable legacy/manual-inspection entries and document the tool.** `unrecognisedMetadataEntries` currently reports only a count, so the operator cannot identify what needs inspection. Return a structured list containing the stable metadata point ID, safe available fields and the reason it cannot be classified, without inferring an identity. Document the report/apply flow, path-state meaning, confirmation requirements, shared-store limitation, partial-failure behavior and idempotent retry in the README and management-tool reference.

7. **Exercise the real Qdrant boundary.** The new cleanup tests currently mock Qdrant. Add an isolated-prefix integration regression that creates real code, graph, symbol, context and metadata resources, verifies report-only behavior, applies exact-identity deletion, confirms foreign resources remain untouched, and repeats the apply safely.

These points complete the agreed identity selection, stale-preview, active-operation, complete cleanup, legacy-reporting and compatibility requirements. Please do not add unrelated reclamation or migration behavior. Once this correction pass is finished and CI is terminal, trigger `@coderabbitai review` for the final head and stop pushing until that review completes.

SIRTHEO:
Correction pass pushed as 0eb2b6f, rebased on `main` after #169. One commit, scope limited to the seven points; nothing else in reclamation or migration changed.

1. **Foreign collections.** `identityFromResourceName` now rejects any name that does not start with `QDRANT_COLLECTION_PREFIX` before it looks at families or the symbol-graph suffix. Regression with `team_` set: an unprefixed `codebase_alpha` and `alpha_symgraph_meta` are neither inventoried nor deleted, and their metadata point is listed as unattributable (`list-collections-metadata-scroll.test.ts`). The real-store test adds a foreign-prefix installation and an unprefixed lookalike that survive the apply.

2. **Token.** Every metadata point of an identity is kept as a record (point id, `collectionName`, `projectPath`, `indexingStatus`, `lastIndexedAt`, `lastBuiltAt`, `builtByVersion`); the token is a sha256 over the identity, path state, the sorted collection set and the sorted records. Regression with a code and a context record: changing either one, or dropping one, changes the token; unchanged input does not. Any record at `in-progress` refuses the delete.

3. **Barrier.** New `reclamation-barrier.ts`: an in-process set plus the `prune` lock and every writer lock (`index`, `watch`, `graph`, `context`) taken by identity through `lock.ts`, held from the re-read under the barrier through deletion and the post-delete verification. `indexProject`, `updateProjectIndex`, `startWatching`, `rebuildGraph`, `indexAllArtifacts` and `ensureArtifactsIndexed` call `assertNoReclamationBarrier` before they start; a `LockInspectionError` there stops the writer too. Graph builds and context indexing held no cross-process lock before, so a build in another local process was invisible to prune and could recreate collections after the delete; they now hold `graph`/`context` locks by identity while they run (`withWriterLock`), which also means a second process asking to build the same identity's graph while one is running is refused rather than run alongside it — the one behaviour change outside prune, stated in the README and the tool reference. `isProjectIdentityLocked` throws on an inspection failure (including an unreadable lock file: only `ENOENT` means absent) instead of answering false; `isProjectLocked` keeps its previous answer for its existing non-destructive callers. A barrier lock reported compromised by proper-lockfile refuses the delete. `acknowledgeNoRemoteWriters: true` is required on apply, and the README and tool reference state what it does and does not cover. Regressions: lock-check error (check throwing, and stat failing), writer starting after the former final check (barrier released), graph and context activity in this process and their locks held by another process, barrier refusing to stand on a lock this process holds, another process holding the prune lock, writer lock held for the duration of a write (`remove-tools.test.ts`, `reclamation-barrier.test.ts`, `reclamation-writers.test.ts`).

4. **Idempotence.** Not-found from either delete counts as `deleted`; other errors stay `failed` with the message. Metadata outcome lines carry the point id (`metadata codebase_x [point …]`), so two legacy records for one collection are told apart. The metadata delete passes `wait: true`; the inventory is re-read afterwards and success is reported only when nothing of the identity remains, otherwise the leftovers are listed. Regressions for not-found on both paths, for `wait: true`, and for an identity still stored after every delete reported success.

5. **Invalidation by identity.** `invalidateGraphCacheForIdentity` drops every cached graph whose path resolves to the identity; `invalidateProjectHashesForIdentity` clears the identity-keyed hash cache; `dropSymbolGraphCache(identity)` plus the collection cache reset. Regression: a pinned identity loaded through a second checkout is still served after a path-keyed invalidation of the recorded path and is not after the identity-keyed one (`graph-cache-identity.test.ts`).

6. **Legacy entries and docs.** Identity parsing checks the symbol-graph suffix before the family prefix, so a pinned id that itself begins with `context_` or `codebase_` stays one identity. `unrecognisedMetadata` is a list of `{ pointId, collectionName, projectPath, reason }`; the report prints each. README section “Reclaiming stored identities” and `tool-reference.md` cover the report/apply flow, path states, confirmation requirements, the shared-store limitation, partial failure and idempotent retry.

7. **Real store.** `tests/integration/prune-reclamation.test.ts` runs under a random prefix against the configured Qdrant: two identities with code, graph, context, symbol-graph collections and their metadata points, one directory removed before the report; verifies the report, refuses stale token and unacknowledged apply without deleting, deletes exactly the selected identity, checks the live identity and the foreign resources are untouched, and repeats the apply. It gates on Qdrant being reachable rather than on Docker, since a shared store need not be local.

Verification on this machine: `npx tsc --noEmit`, `npm run lint`, `npm run build` clean; `npm test` 2370 passed, 78 skipped, 3 failed: the three GDScript tests in `graph-gdscript.test.ts` that need the `tree-sitter-gdscript` native package, which is not installed here, as with the previous head. Docker, Qdrant and Ollama were up, so the Docker-gated integration suites ran rather than skipped..

I will trigger `@coderabbitai review` once CI is terminal and not push again until it completes.

SIRTHEO:
@coderabbitai review

SIRTHEO:
Follow-up pushed as 71e0a86 for the five CodeRabbit findings on `0eb2b6f`, plus the one outside the diff. Same scope, nothing new in reclamation.

- **Ambiguous names (outside diff, `qdrant.ts`).** Confirmed: `codebase_` + `context_docs_symgraph_meta` and the symbol-graph meta of `codebase_context_docs` are one string, so no parsing order can be right. The inventory now reads every interpretation a name has and settles it with what the store shows: a metadata point names a family collection, a complete `_symgraph_{meta,file,index}` triple names a symbol graph. Both claims, or neither, and the name goes to a new `unattributedCollections` list with the reason, printed by the report and never deleted. Regressions for all three suffixes in both directions and for the conflicting case (`list-collections-metadata-scroll.test.ts`); the real-store test now includes a pinned `context_docs` identity whose triple is attributed to it. Changing the naming contract itself is yours to decide; I did not touch `config.ts`.
- **Barrier reentrancy (`reclamation-barrier.ts:77`).** Confirmed. The barrier now takes every lock non-reentrantly (a lock this process holds belongs to a writer and answers false) and flags itself before the first lock, so a writer's check fails from the start of the acquisition, not the end. Regression reproduces the interleaving: a writer takes `index` while the barrier waits on `prune`; the barrier returns null and the writer keeps its lock.
- **Nullable build promise (`code-graph.ts:329`).** Confirmed. The promise cached in `graphBuildPromises` now rejects on a refused lock before it is shared, so a joining caller never sees null; the cast is gone.
- **Graph cache by stored identity (`code-graph.ts:166`).** Confirmed. Each cache entry keeps the identity resolved when it was filled; identity invalidation matches that first and the current resolution second. Regression: a checkout that lost its `.socraticode.json` after caching is still dropped.
- **Suffix-ending identities in the integration test.** Covered by the unit regressions above for all three suffixes; the real-store run adds the family-prefixed pinned identity.
- **README count.** Removed the number.

Verification: `npx tsc --noEmit`, `npm run lint`, `npm run build` clean; `npm test` 2380 passed, 78 skipped, 3 failed (the same three GDScript tests needing `tree-sitter-gdscript`), with Docker, Qdrant and Ollama up..

SIRTHEO:
@coderabbitai review

giancarloerra:
CI and the exact-head CodeRabbit review are clean on `71e0a86`, and the previous requested corrections are present. Two destructive-safety gaps remain in the current implementation:

1. **An unattributed collection can become deletable after its metadata evidence is removed.** In the existing conflict regression, `context_docs_symgraph_meta` is left unattributed because both a family metadata point and a complete symbol-graph triple claim it. However, the `docs_symgraph_meta` entry still contains that metadata point and is not marked `requiresManualInspection`. Applying that entry deletes the metadata point, after which the next inventory reclassifies the same collection as `context_docs` symbol-graph data. The first apply can report success without deleting the ambiguous collection, and a later apply can delete it under the other identity. Keep the candidate identities for every unattributed collection, mark any matching inventory entry as requiring manual inspection, and refuse apply for either candidate without deleting collections or metadata. Add a regression that attempts the conflict-case apply, proves no delete runs, and proves a repeated inventory remains ambiguous.

2. **A barrier lost during cleanup does not stop the remaining destructive writes.** `barrier.isCompromised()` is checked once before `removeProjectReclamationEntry()`. If proper-lockfile reports the barrier lost while one of the awaited Qdrant deletions is running, the function continues deleting all remaining collections and metadata and can still report success. Re-check barrier ownership before every destructive write and after the deletion pass; on loss, stop immediately and report incomplete cleanup. Add a regression that flips the barrier to compromised during the first deletion and proves no subsequent resource is deleted and success is not reported.

Please keep this final pass limited to those two paths. Once the changes are finished and CI is terminal, trigger `@coderabbitai review` against the new head and wait for it to complete before pushing again.

SIRTHEO:
Final pass pushed as d8bfc17, limited to the two paths.

1. **Unattributed collections hold back every candidate.** `UnattributedCollection` now carries `candidateIdentities`; every inventory entry among them gets `requiresManualInspection` with the reason (`collection X fits this identity and another; the store does not settle which`), and apply refuses it before the barrier is taken, so neither collections nor metadata are deleted. Regressions: in `list-collections-metadata-scroll.test.ts` the conflict case (family metadata point plus a complete symbol-graph triple) marks both `context_docs` and `docs_symgraph_meta`, and a second inventory read is still ambiguous; in `remove-tools.test.ts` an apply for each candidate is refused with no barrier and no delete; in `prune-reclamation.test.ts` the same conflict is created in a real store, apply for each candidate is refused, the collection list and the contested metadata point are unchanged, and the next report still lists the name as unattributed. `manualInspectionReasons` is also on every entry now, so the report says why an identity is held back rather than only that it is.

2. **A lost barrier stops the remaining writes.** `removeProjectReclamationEntry` takes a `mayContinue` callback asked before every destructive write; once it answers false, the write and every later one are recorded as `skipped` with the reason and nothing more is attempted. `codebase_prune` passes `() => !barrier.isCompromised()`, re-checks ownership after the pass, and reports incomplete cleanup with the skipped resources whenever the barrier was lost; success is only reported when every outcome is `deleted`, the barrier held throughout, and the re-read inventory shows nothing left. Regressions: the real deletion path stops after the first delete when consent is withdrawn during it (no further collection or metadata delete reaches the client); the tool flow flips the barrier to compromised during the first deletion and reports incomplete, never success.

Verification: `npx tsc --noEmit`, `npm run lint`, `npm run build` clean; `npm test` 2384 passed, 78 skipped, 3 failed (the same three GDScript tests needing `tree-sitter-gdscript`), with Docker, Qdrant and Ollama up. I will trigger `@coderabbitai review` once CI is terminal and wait for it before any further push.

SIRTHEO:
@coderabbitai review

SIRTHEO:
Pushed 9677af3 for the one finding on `d8bfc17`. Confirmed: a hold that lifts on its own, because the unattributed collection is gone, left the identity's records and resources unchanged, so a token minted while it was held back would have deleted it once free. `inventoryToken` now covers the sorted `manualInspectionReasons` as well. Regression: with the metadata point and resources fixed, a lone `context_docs_symgraph_file` appearing holds `context_docs` back and changes its token; its disappearing lifts the hold and restores the original token, so the token minted under the hold is refused by `pruneRefusal` either way.

Verification: `npx tsc --noEmit`, `npm run lint`, `npm run build` clean; `npm test` 2385 passed, 78 skipped, 3 failed (the same GDScript tests needing `tree-sitter-gdscript`). I will trigger `@coderabbitai review` once CI is terminal and wait for it before any further push.

SIRTHEO:
@coderabbitai review

giancarloerra:
`9677af3` correctly fixes the confirmation-token finding. Of the two new CodeRabbit findings, one is valid and one conflicts with the agreed scope:

1. **Valid blocker: indexing can re-enter the barrier-held `index` lock.** Both `indexProject()` and `updateProjectIndex()` check the reclamation barrier before calling `acquireProjectLock()`. If reclamation acquires its barrier between those operations, `acquireProjectLock()` currently uses the re-entrant default and returns success for the lock already held by this process on behalf of the barrier. Indexing can then write while deletion is active and can release the barrier's lock. Make these indexing lock acquisitions non-re-entrant, preserving normal acquisition when no lock is held. Add a regression where the writer's barrier check passes, reclamation acquires the barrier before the writer takes its lock, and the writer is then refused before any Qdrant write. Cover both full and incremental indexing.

2. **Do not implement the `pathState === absent-on-this-host` restriction.** The accepted design intentionally makes path state advisory. A live path can still have an explicitly selected superseded identity after pinning, while an absent path on a reader host says nothing about a shared store. Exact identity selection, a current confirmation token, active-writer checks, the reclamation barrier, and `acknowledgeNoRemoteWriters` are the deletion gates.

Please keep this pass limited to the non-re-entrant indexing race. Once the correction is complete and CI is terminal, trigger `@coderabbitai review` on the new head and wait for it to finish before any further push.

SIRTHEO:
On the two outside-diff comments from the review of `9677af3`:

- **Index lock re-entrancy during reclamation (`indexer.ts`).** Confirmed, and fixed in 4866280. Between the writer's barrier check and its lock acquisition the barrier can take the identity's locks, and the re-entrant acquire then hands the writer the barrier's own `index` (or `watch`) lock. `indexProject`, `updateProjectIndex` and `startWatching` now record whether they held the lock before acquiring and, if they did not and the barrier is held afterwards, stand down without releasing a lock that is not theirs. The re-entrant acquire itself stays as it was, since `updateProjectIndex` delegating to `indexProject` under its own lock depends on it. Graph builds and context indexing were already non-reentrant through `withWriterLock`. Regressions in `reclamation-writers.test.ts`: each of the three writers stands down and never calls release.

- **Refusing apply unless `pathState` is `absent-on-this-host` (`index-tools.ts`).** Not applied, deliberately. The corrected design on #167 makes path state an observation the operator reads, not a gate: in a shared store the path exists only on the host that wrote the index, so it is absent precisely where deleting is most dangerous, and the case this issue was opened for — a path-hash identity superseded by a pinned `projectId` for the same checkout — has a path that is present. `unknown/inaccessible` also covers legacy points with no recorded path, which must stay reclaimable without migration. Refusing on those states would block the intended uses and not protect the shared-store one; the exact identity, the fresh token, `acknowledgeNoRemoteWriters` and the barrier are the gate. If you would rather have an explicit acknowledgement for a present or unknown path, that is a one-line addition and I will make it.

Verification: `npx tsc --noEmit`, `npm run lint`, `npm run build` clean; `npm test` 2388 passed, 78 skipped, 3 failed (the same GDScript tests needing `tree-sitter-gdscript`). I will trigger `@coderabbitai review` once CI is terminal and wait for it before any further push.

SIRTHEO:
@coderabbitai review

giancarloerra:
`4866280` does not fully close the same race. `heldBefore` is sampled after `assertNoReclamationBarrier()`, so it cannot establish who owns the lock:

1. If reclamation acquires the `index` or `watch` lock before `heldBefore` is read, `heldBefore` is `true`; the writer then treats the barrier's lock as its own and proceeds.
2. If reclamation sets its in-process barrier flag but the writer acquires the operation lock first, the new check makes the writer return without releasing the lock it actually acquired. Reclamation then fails its acquisition and clears the barrier, leaving the writer lock leaked.

The regressions currently force `holdsProjectLock()` to return `false`, so neither ordering is covered.

Please replace the before/after ownership inference with an atomic ownership result. The simplest shape is non-reentrant acquisition for the public `indexProject`, `updateProjectIndex`, and `startWatching` entry paths, while preserving the intentional update-to-full-index delegation through a private already-locked implementation path rather than a re-entrant outer acquisition. Add regressions for both orderings above, proving that no Qdrant write begins and that neither the barrier's lock nor a writer-acquired lock is incorrectly released or leaked.

This is still the same indexing/reclamation race, not new scope. Once corrected and CI is terminal, trigger `@coderabbitai review` on the new head and wait for its final result.

SIRTHEO:
Pushed fdf60d7. Both orderings were as you describe; the inference is gone.

- `acquireProjectLock` takes `{ reentrant: false }`, and the public `indexProject`, `updateProjectIndex` and `startWatching` use it: a lock this process already holds — the barrier's or another run's — refuses the acquisition, and a refused acquisition starts nothing and releases nothing. The full-index body moved to a private `indexProjectLocked`, which `updateProjectIndex` calls for its fallback under the one lock it took; only the acquirer releases. The re-entrant acquire itself remains available to `acquireIdentityLock` callers that opt in; no public writer path uses it. `assertNoReclamationBarrier` stays as the early, cheap refusal; ownership is now decided by the lock alone.
- Regressions in `tests/unit/reclamation-lock-orderings.test.ts`, on real lock files and the real barrier with only the store and the file watcher stood in for. Order 1, reclamation first: with the barrier held, `indexProject` is refused, writes nothing and the barrier still holds all five locks until it releases them; with only the `index`/`watch` lock held in this process and no flag visible, `indexProject`, `updateProjectIndex` and `startWatching` are refused by the non-reentrant acquire, write nothing, and the held lock is untouched. Order 2, writer first: from inside the writer's lock the barrier answers null and clears its flag while the writer keeps its lock, no write is made, and every lock file is released when the writer exits; the update's fallback runs with exactly one `proper-lockfile` acquisition and holds the lock throughout. The obsolete inference tests are removed.

Verification: `npx tsc --noEmit`, `npm run lint`, `npm run build` clean; `npm test` 2392 passed, 78 skipped, 3 failed (the same GDScript tests needing `tree-sitter-gdscript`). I will trigger `@coderabbitai review` once CI is terminal and wait for its result before any further push.

SIRTHEO:
@coderabbitai review

SIRTHEO:
Pushed fbd8278. Confirmed: with `SOCRATICODE_PROJECT_ID` set in the environment, `projectIdFromPath` returns it for every project, so the pinned-identity assertion fails before any ordering is exercised. The same ambient value breaks the two other tests of this PR that pin identities through `.socraticode.json` (`graph-cache-identity.test.ts`, where every checkout would collapse to one identity, and the Qdrant integration test, where the three projects would share one), so all three now clear the variable in their setup and restore the environment afterwards (`vi.stubEnv(..., undefined)` / `vi.unstubAllEnvs()`).

Evidence: with `SOCRATICODE_PROJECT_ID=ambient_pin` the three files fail 9 of 13 tests without the guard and pass 13 of 13 with it; without the variable they pass either way. `npx tsc --noEmit`, `npm run lint`, `npm test` 2392 passed, 78 skipped, 3 failed (the same GDScript tests needing `tree-sitter-gdscript`). I will trigger `@coderabbitai review` once CI is terminal.

Outside this diff: the `SOCRATICODE_AUTO_RESUME=all` block of `tests/unit/startup.test.ts` (from June) derives two identities from the path hash the same way and has the same exposure; left untouched here as it is not part of this change.

SIRTHEO:
@coderabbitai review

PR #174, opened 2026-09-16
feat(graph): support configured Python import roots

### LINKED ISSUE #171: [Bug]: Python imports unresolved when the source root is neither the project root, src/, nor lib/ and there is no pyproject.toml (e.g. Airflow dags/) — graph reports 0 dependents with confidence
### Describe the bug

Python absolute imports resolve against only three candidate roots — the project root, `src/`,
and `lib/` — plus (since #107) roots declared by a `pyproject.toml`. A project whose Python
source root is **none of those and has no `pyproject.toml` at all** resolves zero imports, so
every Python file becomes an orphan: `codebase_graph_query` answers "No dependency information
found for this file", `codebase_impact` answers "No callers found — nothing else depends on
this", and `codebase_symbol` reports "Callers (0)".

The dangerous part is that these are **confident negative answers, not errors**. The graph
builds, reports READY, and `codebase_impact` on a module imported by 9 other modules says
nothing depends on it. Used as intended — "check the blast radius before refactoring" — it is
a green light to break the build.

This is the same shape as #107, but the trigger is the *absence* of a manifest rather than a
workspace layout, so #107's fix does not reach it.

**Apache Airflow is the common case.** Airflow puts the DAG bundle root on `sys.path`, so the
conventional layout is `dags/` as the import root with no packaging manifest anywhere:

```
repo/
  dags/            <- sys.path root (Airflow adds it; nothing declares it)
    etl/config.py
    etl/cache.py   ->  from etl.config import dw_engine
  requirements.txt <- no pyproject.toml
```

`etl.config` is probed as `etl/config.py`, `src/etl/config.py`, `lib/etl/config.py` — all
missing, since the file is `dags/etl/config.py`. The sibling-flat fallback (#46) probes
`dags/etl/etl/config.py`, also missing. Result: `null`.

### Steps to reproduce

Minimal repro — two files, identical content, only the nesting differs.

```bash
# A: package at project root
mkdir -p pyroot/etl
printf 'def dw_engine():\n    return 1\n' > pyroot/etl/config.py
printf 'from etl.config import dw_engine\n\ndef run():\n    return dw_engine()\n' > pyroot/etl/cache.py

# B: same package one level down (Airflow-style)
mkdir -p pynested/dags/etl
cp pyroot/etl/*.py pynested/dags/etl/
```

Index both, then:

```
codebase_graph_query { projectPath: "<pyroot>",   filePath: "etl/config.py" }
codebase_graph_query { projectPath: "<pynested>", filePath: "dags/etl/config.py" }
```

### Expected behavior

Both resolve `from etl.config import dw_engine`, since `dags/` is the import root exactly the
way a project-root or `src/` layout is — `etl/config.py` exists under it and `etl.config`
names it unambiguously.

### Actual behavior

```
A  etl/config.py        ->  Imported by (1): etl/cache.py          OK
B  dags/etl/config.py   ->  No dependency information found        WRONG
```

Verified directly against the **1.13.3** resolver (no `pyproject.toml`, so
`pythonRootsForFile` returns `[]`):

```js
const m = await import('./dist/services/graph-resolution.js')
m.resolveImport('etl.config', '/repo/dags/etl/cache.py', '/repo',
  new Set(['dags/etl/config.py','dags/etl/cache.py']), 'python',
  ...Array(6).fill(undefined), /* pythonImportRoots */ [])
// -> undefined   (UNRESOLVED)

// same call, package at the root:
m.resolveImport('etl.config', '/repo/etl/cache.py', '/repo',
  new Set(['etl/config.py','etl/cache.py']), 'python', ...)
// -> 'etl/config.py'
```

On a real 31-file Airflow repo the whole Python half of the graph is empty, while the
TypeScript half of the same repo is exact:

| Query | Reported | Actual (grep) |
|---|---|---|
| `codebase_impact dags/etl/config.py` | 0 files | 9 importers |
| `codebase_symbol copy_rows` | 0 callers | 4 call sites |
| `codebase_symbol dw_engine` | 0 callers | 6 call sites |
| `codebase_graph_query api/src/db/index.ts` | 8 importers | 8 — correct |

Re-indexing `dags/` as its own project root fixes it completely and is a usable workaround
(30 files / 75 edges; `copy_rows` then reports exactly the right 4 call sites) — but it costs a
second index of the same repo, its paths are relative to `dags/`, and it drops the sibling
`scripts/` that imports `etl.*` too.

Worth noting the parser is fine — `@ast-grep/lang-python` parses the files correctly and
`extractImports` returns `etl.config`. This is purely source-root discovery.

### Embedding provider

Ollama (local)

### Qdrant mode

Managed (Docker)

### MCP host

Claude Code

### Operating system

macOS 15 (Darwin 24.6.0, arm64)

### Node.js version

22.22.0

### Additional context

Possible directions, roughly in order of how much they guess:

1. **Infer roots from the tree.** If `<dir>/<first-segment-of-module>/` exists and contains
   `.py` files, treat `<dir>` as a path entry. Here `dags/etl/` exists, so `dags/` becomes a
   candidate. Self-resolution guards from #157 still apply.
2. **Let the project declare it** — `pythonRoots: ["dags"]` in `.socraticode.json`. Explicit,
   no inference, and useful for any framework with a convention-only source root.
3. **Recognise Airflow specifically** — an `airflow.cfg`/`dags_folder`, or a top-level `dags/`
   containing `.py`, is a strong signal. Narrow, but covers a large population of repos.

(1) or (2) would also cover the general "source root is neither `src/` nor `lib/` and there is
no manifest" case — `app/`, `lambdas/`, `functions/`, and similar convention-only roots.

Happy to send a PR for (2) if that is the direction you'd prefer.

### Checklist

- [x] I have searched existing issues and this hasn't been reported before (closest are #46 and
      #107; both are fixed and neither covers a root with no manifest)
- [x] I am using the latest version of SocratiCode — observed on the running 1.12.0, and the
      failing resolution re-verified directly against 1.13.3 as shown above

giancarloerra:
## Summary

- add optional project-root-relative `pythonRoots` configuration in `.socraticode.json`
- resolve Python imports through configured roots after the existing project-root, `src/`, `lib/`, and sibling probes, and before manifest-derived roots
- record effective roots in the graph settings fingerprint so relevant configuration changes rebuild only the graph on the next update
- document the setting with an Airflow `dags/` example

## Compatibility

- absent, empty, malformed, or unusable `pythonRoots` preserves existing resolution and the legacy graph settings fingerprint
- absolute paths, outside-root paths and symlink escapes, missing paths, files, and non-string entries are ignored
- existing project-root, `src/`, `lib/`, sibling, and `pyproject.toml` behavior remains covered
- no code/vector re-index, migration, reset, or manual intervention is required

## Validation

- `npm run lint`
- `npx tsc --noEmit`
- `npm run build`
- `npm test` (2,474 passed, 14 skipped)
- local CodeRabbit review (0 findings)

Fixes #171

## Summary by CodeRabbit

- **New Features**
  - Added support for configuring Python import roots through `.socraticode.json`.
  - Configured roots are validated, normalized, deduplicated, and searched before automatically detected project roots.
  - Python import resolution now supports configured project, `src`, and manifest-derived roots.

- **Bug Fixes**
  - Invalid, missing, out-of-project, and non-directory roots are safely ignored.
  - Graph data is rebuilt when effective Python roots change, while unchanged roots avoid unnecessary rebuilds.

- **Documentation**
  - Added configuration guidance and validation rules for `pythonRoots`.


giancarloerra:
@coderabbitai review

Issue #180, opened 2026-09-19
[Feature]: Allow the plugin-manifest launcher to be overridden, as socraticode.command already does for VS Code


gregoryfoster:
### Problem or motivation

**Please let the plugin-manifest launcher be overridden per host, the way `socraticode.command` already can be in the VS Code extension.** The default should stay exactly as it is.

The update-aware launcher is deliberate — `tests/unit/distribution-launchers.test.ts` pins `npx -y --prefer-online socraticode@latest` across every bundled definition and install link — and I am not asking you to change it. The ask is only that the plugin path gain the escape hatch the extension already has: `extension/package.json` documents `socraticode.command` as "Override only if you have a custom install", and `mcpProvider.ts` honours it. `.mcp.json` has no equivalent, so on Claude Code, Codex and Cursor the only way to hold the engine still is to disable the plugin — which also drops the `codebase-exploration` and `codebase-management` skills and the `codebase-explorer` agent, and renames every tool from `mcp__plugin_socraticode_socraticode__*` to `mcp__socraticode__*`, which our own tooling references.

Two costs follow. The second has a number; the first is the one we meet more often.

#### 1. Engine versions that move independently

On a plugin host there is no single "the engine version":

| | What moves it |
|---|---|
| Plugin package | `claude plugin update` — pinned to a release |
| **Session engine** | `@latest`, resolved when the session starts |
| **Out-of-process engine** | `@latest`, resolved whenever a script or hook starts one — a *different* moment |

The two engines agree only by coincidence of timing. Ours diverged in an ordinary way: the plugin cache held 1.13.2 while a launch pulled 1.14.0. That matters in two concrete places:

- **Graph artifacts carry a builder version.** "Is this graph older than the server now answering questions about it" is a real question, but a floating engine adds a spurious case: stale by accident of which minute each process launched, rather than by an actual upgrade.
- **A shared index store gets two writers at two versions.** We run several repos against one external Qdrant. "One engine version per store" is something we currently hope for and monitor; we cannot make it true.

We shipped a workaround — our setup skill prefers a pre-installed engine and reports a finding when its version and the session's `@latest` differ by a feature release. It detects the divergence. It cannot remove it, because we cannot pin the session.

#### 2. The launch installs, and that is the largest thing the integration does

[...]

Issue #149, opened 2026-09-05
Clarify MCP and plugin updates and document host-specific environment configuration


giancarloerra:
## Problem

SocratiCode has two independently updated components:

1. The MCP engine published as the `socraticode` npm package.
2. Host plugin or extension files containing skills, instructions, manifests, and UI integrations.

The README says that users can change the package argument to `socraticode@latest` for an immediate update, but the Quick Start, install links, examples, and bundled launcher configurations still use bare `socraticode`. A bare package name does not clearly guarantee that a newly published npm release will be checked for on the next MCP start.

The README also documents environment-variable syntax for direct MCP configurations, but native plugins do not all provide the same override mechanism. In particular, Codex, VS Code Agent Plugin, and the current SocratiCode Cursor plugin need an explicit workaround when users want to retain plugin skills while supplying custom MCP environment variables.

The result should explain one consistent update model and provide an accurate configuration path for every supported host.

### 1. Use an explicit latest-release MCP command

Use the following command in the Quick Start, host examples, install links, and bundled MCP launcher defaults:

```text
npx -y --prefer-online socraticode@latest
```

Equivalent structured arguments:

```json
"args": ["-y", "--prefer-online", "socraticode@latest"]
```

The documentation must explain that:

- `@latest` selects the version currently published under npm's `latest` distribution tag.
- `--prefer-online` forces npm to check whether newer package metadata is available instead of relying only on fresh cached metadata.
- The update check happens when the MCP server starts. Existing processes do not replace themselves while running.
- Users must restart or reconnect the MCP server and start a new host session after a release.
- Obtaining a newly published version requires access to the npm registry. Offline behaviour must be described according to the validated result rather than assumed.
- No separate Git branch is involved in resolving the npm release.

Update every shipped location that currently launches bare `socraticode`, including:

- README Quick Start and host examples.
- VS Code and Cursor installation links.
- `.mcp.json`.
- `mcp.json`.
- `gemini-extension.json`.
- VS Code extension defaults and descriptions.

Search for any additional generated or packaged launcher configuration so the documented command and shipped defaults cannot drift.

### 2. Explain engine updates separately from plugin updates

Add a short **Keeping SocratiCode up to date** section immediately after Quick Start.

It must state that:

- The npm MCP engine and host plugin files are separate update channels.
- The explicit npm command keeps the engine on the latest published npm release when the server starts and the registry is reachable.
[...]

PR #182, opened 2026-09-20
feat: allow the Claude Code plugin engine specification to be pinned

gregoryfoster:
## Summary

Claude Code's plugin manifest now points at its own MCP definition file, `.claude-plugin/mcp.json`, in which the package specification is `${SOCRATICODE_SPEC:-socraticode@latest}`. Unset — which is every existing installation — it resolves to `socraticode@latest` and the launcher is byte-for-byte what it is today. Set, every server started on that host resolves the same engine build instead of whatever `latest` is at each start.

This is #180, and it is narrower than that issue proposed. Two things about it are worth reading before the diff, because both changed the shape of the change.

**Only the specification is overridable, not the command.** The issue asked for two independent overrides. Expansion is applied per string, so a fixed-arity `args` array cannot express a whole-command override: replacing `npx` would leave `-y --prefer-online` in front of the new command, and `SOCRATICODE_COMMAND=node` would run `node -y --prefer-online <spec>`, which exits with `node: bad option: -y`. Per the caveat in the issue — *"If the supported plugin packaging formats do not permit the definitions to be separated cleanly, document that constraint"* — that half is documented here rather than forced. Pinning the specification removes drift between servers and stops the re-download on each release; it still contacts the registry for metadata, so it is not a start-up-cost fix. Skipping npm entirely would need a mechanism this format does not offer.

**A referenced file rather than an inline object.** General `${VAR:-default}` expansion is documented for `.mcp.json` files. For `plugin.json` the reference documents only `${CLAUDE_PLUGIN_ROOT}`, `${CLAUDE_PLUGIN_DATA}` and `${CLAUDE_PROJECT_DIR}`, and says nothing either way about arbitrary environment variables. Inline *does* expand them — I measured it on both 2.1.71 and 2.1.278 — but a shipped plugin resting on behaviour the reference does not describe seemed like the wrong trade, and a separate file also keeps Claude's launcher physically apart from the one Codex and Cursor share. If you would rather have it inline, it is a two-line change.

`.mcp.json` and `mcp.json` are untouched, so Codex and Cursor keep the literal launcher and no host-specific syntax reaches the file they share.

## Changes

- `.claude-plugin/plugin.json` — `mcpServers` now names `./.claude-plugin/mcp.json` instead of the shared `./.mcp.json`.
- `.claude-plugin/mcp.json` — new; the Claude Code launcher, with `${SOCRATICODE_SPEC:-socraticode@latest}` as the package specification.
- `tests/unit/distribution-launchers.test.ts` — the Claude default and pinned cases; an assertion that the manifest names a *file* rather than defining servers inline, carrying the reason; an assertion that Claude's definition is not the shared file; and the existing `${`-free check extended to both shared files. Every pre-existing literal assertion for the other distributions is unchanged.
- `README.md` — a *Pinning the engine version* subsection in the Claude Code plugin section. No other host's documentation is touched.

## Type of change

- [x] Documentation update

## Testing

- [x] Unit tests pass (`npm run test:unit`) — 97 files, 2284 tests
- [ ] Integration tests pass (`npm run test:integration`) — not applicable; no engine code changed

`npx biome check src/ tests/` is clean and warning-free, matching the baseline.

**The expansion behaviour was measured, not assumed.** A throwaway local plugin was installed whose manifest referenced a file containing both `${VAR:-default}` and a bare `${VAR}`, and the argv the spawned process actually received was recorded — not just the `claude mcp list` display. Run on Claude Code **2.1.278**, and separately reproduced on 2.1.71:

| Variable state | argv element received |
|---|---|
| unset, with a default | the default |
| set to a value | the value |
| set to the empty string | empty — **not** the default |
| unset, no default | the literal `${VAR}` |

Two things follow. The referenced-file form expands, which is what this PR depends on. And `${VAR:-default}` is spelled like shell `:-` but behaves like `-`: an empty value is a value, not an absence. The README says to remove the setting rather than blank it, for that reason.

The same probe with an inline definition and no referenced file present produced the identical three results on both versions, which is why the choice above is framed as a preference for the documented surface rather than a necessity.

The new assertions were each verified by reintroducing the fault they guard: an inline definition fails two, pointing Claude back at the shared file fails two, and expansion syntax in `.mcp.json` fails both the new guard and the pre-existing update-aware-launcher assertion.

## Checklist

- [ ] I have addressed all CodeRabbit review comments (or marked as resolved with explanation) — none yet

## Related issues

Relates to #180. Context from #149, whose per-host override recipes this completes: Codex, the VS Code Agent Plugin and Cursor can each disable a plugin's bundled MCP server and substitute their own, and Claude Code cannot — its plugin granularity is whole-plugin, so the only equivalent is disabling the plugin and losing its skills and agent with it.

## Summary by CodeRabbit

- **New Features**
  - Added Claude Code plugin support for configuring the Socraticode engine version through `SOCRATICODE_SPEC`.
  - The plugin uses the latest engine by default, with support for pinning a specific version to prevent automatic updates.

- **Documentation**
  - Added guidance explaining engine version pinning and clarifying configuration behavior across Claude Code, Codex, Cursor, and the VS Code extension.


giancarloerra:
@coderabbitai review

PR #183, opened 2026-09-20
fix(graph): capture PHP structural symbol edges so codebase_impact stops reporting zero dependents

### LINKED ISSUE #181: [Bug]: PHP symbol edges are calls only — extends/implements/new/type hints produce none, so codebase_impact under-reports against its own file graph
### Describe the bug

`extractFromPhp` (`src/services/graph-symbols.ts:2547`) extracts PHP **symbols** correctly - `class_declaration`, `interface_declaration`, `trait_declaration`, `function_definition` and `method_declaration` all become symbols. But its `rawCalls` loop reads only three node kinds, and all three are calls: `function_call_expression`, `member_call_expression`, `scoped_call_expression`.

So a PHP class whose tie to another file is **structural** contributes no symbol edge. `extends`, `implements`, a trait `use`, `new Foo()`, and parameter or return type hints all produce nothing. `codebase_impact` walks the symbol graph, so it silently omits those dependents - while `codebase_graph_query`, walking the file graph, lists them. The two tools disagree about the same repository.

`codebase_graph_status` documents symbol edges as *"calls, imports, re-exports and type or value references"*. For PHP only the first of the four is captured. `extractFromTsLike` is the only extractor emitting `type_reference` / `value_reference` edges (lines 1483, 1506, 1569) - what #135 added for TypeScript after #132. This is the PHP counterpart one level down: #132 and #140 were about symbols never being extracted; here the symbols exist and their **edges** do not.

The consequence is the one #132 named. `codebase_impact` phrases the empty result as *"No callers or references found - nothing else depends on this"*, and it is the tool run before a rename or a delete.

### Steps to reproduce

Four files and a manifest. No framework, no vendor directory.

`composer.json`

```json
{
  "name": "repro/php-extends-edge",
  "autoload": { "psr-4": { "App\\": "src/" } }
}
```

`src/Base/Base.php`

```php
<?php

namespace App\Base;

class Base
{
    public function hello(): string
    {
        return 'hi';
    }
}
```

`src/Child/Child.php` - subclass declaring no member of its own; its ONLY tie to `Base` is `extends`

```php
<?php

namespace App\Child;

use App\Base\Base;

class Child extends Base
{
}
```

`src/Caller/Caller.php` - the control: reaches `Base` through a method CALL

```php
<?php

namespace App\Caller;

use App\Base\Base;

class Caller
{
    public function go(Base $b): string
    {
        return $b->hello();
    }
}
```

`src/Consumer/Consumer.php` - reaches `Child` only through `new` and a return type

```php
<?php

namespace App\Consumer;

use App\Child\Child;

class Consumer
{
    public function make(): Child
    {
        return new Child();
    }
}
```

Then:

1. `codebase_index` the directory, then `codebase_graph_build`; wait for `READY`.
2. `codebase_graph_query filePath="src/Base/Base.php"`
3. `codebase_impact target="src/Base/Base.php"`
4. `codebase_graph_query filePath="src/Child/Child.php"`
5. `codebase_impact target="src/Child/Child.php"`

### Expected behavior

`codebase_impact` reports the same dependents its own file graph reports. Step 3 lists both `Caller.php` and `Child.php`; step 5 lists `Consumer.php`.

### Actual behavior

| probe | `codebase_graph_query` (file graph) | `codebase_impact` (symbol graph) |
|---|---|---|
| `src/Base/Base.php` | imported by **2** - `Caller.php`, `Child.php` | **1** - `Caller.php` only; `Child.php` dropped |
| `src/Child/Child.php` | imported by **1** - `Consumer.php` | **0** - *"nothing else depends on this"* |

Only the method call survives. `extends`, `new` and the type hints contribute nothing.

The graph reports `Symbols: 7`, `Call edges: 1`, **`Unresolved: 0.0%`** - so this is not unresolved-edge noise. `codebase_symbols` on `src/Child/Child.php` returns `class Child src/Child/Child.php:6`: the symbol is captured, its edges are not.

### Embedding provider

Ollama (Docker — default)

### Qdrant mode

Managed (Docker — default)

### MCP host

### Operating system

macOS 26.6.2

### Node.js version

v22.18.0

### Relevant log output

```shell
Code Graph Status — Files (nodes): 4   Dependencies (edges): 3   Built by: v1.14.0
Symbol graph: Files: 4  Symbols: 7  Call edges: 1  Unresolved: 0.0%

$ codebase_graph_query src/Base/Base.php
Imported by (2):
  ← src/Caller/Caller.php
  ← src/Child/Child.php

$ codebase_impact src/Base/Base.php
Depth: 3    Total impacted files: 1
Hop 1 (1 files):
  - src/Caller/Caller.php

$ codebase_graph_query src/Child/Child.php
Imports (1):      → src/Base/Base.php
Imported by (1):  ← src/Consumer/Consumer.php

$ codebase_impact src/Child/Child.php
Depth: 3    Total impacted files: 0
No callers or references found — nothing else depends on this.
```

### Additional context

At scale on a real WordPress/Bedrock monorepo (1476 files, 3087 file edges, graph built by v1.14.0 from a from-scratch index): `codebase_graph_query` on a widely-extended base class lists 24 importers - its `use` census exactly - where `codebase_impact` lists 21. Two of the three missing files are absent from the entire depth-3 radius, not merely demoted; the third is a direct importer reported at hop 2. All three reach the base structurally. One of them is a subclass declaring no member at all, and `codebase_impact` on *it* returns 0 while two files import it.

`extractFromGo`, `extractFromRust`, `extractFromRuby` and the others share the calls-only shape, so the gap is probably not PHP-specific - but this report is measured on PHP only.

### Checklist

- [X] I have searched existing issues and this hasn't been reported before - nearest are #132 (TS, non-function exports never extracted), #140 (Rust, same), #172 (unresolved-share accounting) and #120 (PHP import resolution). All closed, all about symbols or file-import edges; none about missing symbol **edges** for captured PHP symbols.
- [X] I am using the latest version of SocratiCode - plugin 1.14.0, server 1.14.0 (`npm view socraticode version` -> 1.14.0), graph `Built by: v1.14.0`, `builder.state: "current"`

gregoryfoster:
## Summary

Fixes #181. `extractFromPhp` emitted rawCalls from only three call-expression node kinds, so a
PHP class reached only through `extends`, `implements`, a trait `use`, `new`, or a type hint had
no symbol edge — and file-mode `codebase_impact`, which derives its reverse index from resolved
symbol edges, reported no dependents for a base class every subclass names.

Follows the scope set out in the issue: PHP only, existing edge kinds, resolution through the
existing import machinery, an end-to-end fixture matching the report, and coverage for aliases
and duplicate short names.

## Changes

- `extractFromPhp` now also emits `type_reference` edges for `extends`, `implements`, trait
  `use`, and parameter, return and property types, and a `call` edge for `new` — matching how
  `extractFromTsLike` records `new_expression`.
- `phpTypeRefName` reduces an FQCN, a namespace-relative path or a bare name to its terminal
  segment and maps a bare name through the file's `use` aliases, so
  `use App\Base\Base as Ancestor; class X extends Ancestor` is an edge to `Base` with
  `localAlias: "Ancestor"`.
- `use` aliases are scoped to their **namespace** and to their **position** — a file-wide table
  applies one block's alias inside another, and an unpositioned one applies an alias above the
  `use` that declares it. Both fabricate an edge to a class the reference never names. PHP
  agrees: `namespace A; class One extends Al {} use X\Base as Al;` fatals with
  `Class "A\Al" not found`, and the same file with `use` moved above resolves to `X\Base`.
- Pseudo-types (`self`, `parent`, `static`) and built-in type names are filtered. `named_type`
  is read through `safeFindAll` on the type node, which unwraps `?Foo`, `A|B`, `A&B` and
  `(A&B)|null` in one pass.
- New edges are deduplicated on `(callerId, calleeName, kind)` against the existing call edges,
  so a `new Foo()` beside a `Foo()` adds nothing and no existing edge is dropped.

**`sourceModule` is deliberately left undefined on every new edge.** Resolution's
module-targeted branch matches `sourceModule` as a path; a PHP FQCN is not one, so handing it a
backslashed name strands the edge as `unresolved` and additionally suppresses the same-file
branch, which is guarded on `!edge.sourceModule`. Left undefined, each edge takes the untargeted
scan — which searches only the caller's resolved file-import dependencies, for PHP exactly what
the `use`/require, PSR-4 and FQCN machinery produced. The search is confined to the file's own
import closure by construction rather than guessing a short name repository-wide. Within that
closure the match is still by short name, which has one consequence worth knowing before you
review — it is the first item under Known limitations.

Two additions beyond the issue's literal enumeration, both stated rather than slipped in:
constructor-**promoted** properties (a promoted property is a parameter in the signature, and
constructor injection is where most modern PHP names its collaborators) and **longhand** typed
properties (capturing one without the other reads as an oversight).

## Removed on request: the non-ASCII identifier commit

`ed29855` and its three dedicated tests have been rebased out, as asked. It widened two
ASCII-only name guards to match PHP's byte-based identifier rule, repairing a pre-existing bug
in `extractCalleeNamePhp` that has nothing to do with #181 — happy to open it as its own issue
whenever you want it looked at.

It was also the only commit that touched a line `main` already had. With it gone the whole
branch is **purely additive: `+1,318 / -0`** — `git diff main...HEAD` deletes nothing and
modifies nothing across all three files, so every pre-existing code path is byte-identical and
no existing PHP call edge *can* change.

Confirmed empirically as well as structurally: across the 1,059-file corpus below, all 54,745
of `main`'s `call` edges reappear under the same `(callerId, calleeName)` key — **0 lost, 0
renamed, 2,753 added**, every one of them a `new Foo()`.

## Review round: five correctness gaps

All five you raised reproduced exactly as described, and each was checked against the PHP 8.4
runtime before being treated as a bug rather than a preference. Batched into one commit, with
tests; of the 15 cases this round added, 11 fail against the previous head and 4 are guards that
must not - three positive controls, plus one protecting the case-fold below.

| | gap | what the runtime says |
|---|---|---|
| 1 | `\Foo` took the alias table | `use X\Other as Foo; extends \Foo` fatals with `Class "Foo" not found`, while `extends Foo` resolves to `X\Other` - the slash is the whole difference |
| 2 | grouped `use function`/`use const` read as type aliases | `use function App\Fn\{Base as Thing}; class X extends Thing {}` fatals with `Class "Main\Thing" not found` |
| 3 | alias lookup was case-sensitive | `use X\Base as ImportedBase; class C extends importedbase {}` resolves the parent to `X\Base` |
| 4 | `integer`/`double`/`boolean` filtered as built-ins | PHP warns `"integer" will be interpreted as a class name` and then resolves it as one |
| 5 | closures and arrow functions contributed no return type | - (a coverage gap, not a semantic one) |

Two of these needed a wider fix than the report implies, and both are worth naming:

**Gap 2 needs *both* checks, not one.** The grammar files the `function` qualifier on the
DECLARATION for `use function A\{B as C};` but on the CLAUSE for `use function A\B as C;` **and**
for the mixed group `use A\{B, function c};`. Checking only the enclosing declaration would have
fixed the reported case and let the mixed group's function member through as a type alias. Both
levels are checked, and a mixed-group test pins each half.

**Gap 5's collection has to sit outside the declaration loop.** That loop registers a symbol per
node via `safeFind(node, "name")`; an anonymous function has no name of its own, so it would have
been filed under the first `name` in its body. A test asserts an anonymous function registers no
symbol.

A full re-review of the whole PR after these five found one more thing worth your attention:
resolution can point one of these new edges at the wrong class. That is how PHP names are already
resolved on `main` — this PR extends it to type references rather than introducing it — and its fix
belongs in the resolver, as its own issue. It is the first item under Known limitations, with a
reproduction and a measurement.

## Performance

The PHP extractor gets slower, and the number belongs in the PR rather than in your profiler.
Measured over all 1,059 tracked PHP files of the tree in the issue, best of three, same process,
re-run against this rebased head:

| | ms | per file | `call` | `type_reference` |
|---|---|---|---|---|
| `main` | 1,378 | 1.30 | 54,745 | 0 |
| this branch | **2,527** | **2.39** | 57,498 | 2,526 |

**1.83x**, for +9.6% edges. Two recoveries are in: return types are collected in the existing
declaration loop rather than in two more full-tree passes, and the alias resolver is built
lazily so a file with no structural reference pays nothing. Both were checked with the edge
totals held constant, so neither optimisation changed behaviour — only the time taken.

The review round's closure and arrow-function support added two more full-tree walks, and those
two *are* batched into one `safeFindAllAny`: it returns nodes in document order rather than
grouped by kind, which is fine for two new kinds nothing downstream distinguishes. The named
declarations cannot use it, because their emission order is what the dedupe and the tests pin.
That leaves eight added full-tree passes as the remaining cost, and collapsing them would mean
changing an emission order this PR deliberately holds fixed - happy to take it as a follow-up.

## Known limitations

- **Resolution can point an edge at the wrong class.** PHP names are resolved by their terminal
  name alone, matched across the caller's dependency closure, and whatever qualified that name is
  discarded. This PR does not introduce that — `main` already resolves `Class::method()` this way
  — but it extends it to type references, so it belongs here. When the class a reference names is
  not indexed (a framework class, typically) and a different class with the same short name is in
  the closure, the edge resolves `unique` to the wrong one, whether the name was written fully
  qualified or imported:

  ```php
  namespace App\Http;

  use Illuminate\Http\Request;
  use App\Models\Request as RequestModel;   // indexed; Illuminate's is not

  class Controller
  {
      public function viaFqcn(\Illuminate\Http\Request $r): void {}  // -> App\Models\Request, `unique`
      public function viaUse(Request $r): void {}                     // -> App\Models\Request, `unique`
  }
  ```

  PHP's reflection gives `Illuminate\Http\Request` for both parameters. On `main` neither method
  has an edge at all. The existing `Class::method()` form there is broader still: the PHP extractor
  records no class qualifier, so `Request::capture()` resolves to a `capture()` declared on *any*
  class in the closure — an `Invoice::capture()` included.

  Measured on the corpus below, **none of the 2,229 edges this PR resolves points at the wrong
  class**: each reference was checked against the class PHP's own name resolution picks. It takes
  the collision above, and this tree has none.

  The fix is resolution-side — carry the qualifying name on the edge and drop candidates declared
  under any other. `calleeQualifier` already exists for this, but resolution honours it for `.rs`
  files only. Since that also fixes the `Class::method()` case on `main`, it is better as its own
  issue than as part of this PR; one will follow, cross-referencing this PR.
- An inline FQCN with no matching `use` or `require` produces no file-import dependency, so the
  scan cannot reach it and the edge stays `unresolved`. Documented at the emission site. The
  only alternative is the repository-wide short-name guessing the issue rules out. Measured on the
  corpus below, of the 437 fully-qualified structural references left `unresolved`, 255 name a PHP
  built-in (`\RuntimeException`, `\InvalidArgumentException`, `\DateTimeZone` …) and 156 a
  WordPress core class the tree declares only as a test stub (`\WP_Post`, `\WP_Query` …) — no
  edge is the right answer for either. The remaining **26** name first-party classes outside the
  caller's dependency closure (the plugin's own record classes, mostly), and those
  are genuine misses.
- An `enum`'s `implements` edge attributes to the module symbol, because `extractFromPhp`
  registers scopes for class, interface and trait only and creates no symbol for
  `enum_declaration`. Pre-existing and left alone; the edge itself resolves, only its
  symbol-level attribution is coarse. Happy to file it separately.
- A class named somewhere that is neither a call, a type, a parent nor a `new` produces no
  edge: `catch (MyEx $e)`, an attribute `#[Route]`, `$x instanceof T`, `T::CONST` and
  `T::$prop` were each checked against the extractor and none emits one. All are outside the
  issue's enumeration; listed in full so the omission is visibly deliberate rather than
  sampled.
- A type name that is not ASCII — `class X extends Café`, which PHP accepts, its identifier
  rule being stated over bytes — is dropped by `phpTypeRefName`'s guard. Documented at the
  site. `extractCalleeNamePhp` makes the same assumption, which is why the fix for both was
  the commit you asked be taken out of this PR.
- `Foo(); new Foo();` in one scope collapses to a single `call` edge, though PHP keeps functions
  and classes in separate symbol tables. Resolution returns every match in the dependency
  closure as a candidate either way, so the candidate set is unchanged; the TypeScript extractor
  dedupes identically, and diverging here would make PHP the odd one out.

## Type of change

- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] Unit tests pass (`npm run test:unit`) — 99 files, **2349 tests**, up from 2281 on `main`
- [ ] Integration tests pass — not run (Docker)

**Tier 1**, `tests/unit/graph-symbols-php-references.test.ts` — extractor-level, with
`ensureDynamicLanguages()` in `beforeAll` (without it the PHP parse throws, `safeFindAll`
swallows it, and every assertion passes vacuously). Covers each construct, type-wrapper
unwrapping, alias and FQCN naming, the pseudo-type and built-in filters, caller attribution,
aliases scoped by namespace (braced and unbraced) and by position, and that existing call edges
survive beside the new ones.

**Tier 3**, `tests/unit/code-graph-php-symbol-edges.test.ts` — a real `mkdtempSync` Composer
tree, real `buildCodeGraph` + `resolveCallSites`, modelled on
`code-graph-rust-crate-scope.test.ts`. It is the fixture from the issue: `Child.php` reaches
`Base.php` by `extends`, `Caller.php` by a parameter hint, `Consumer.php` reaches `Child.php` by
`new` and by its return type. Plus an aliased `extends`, and **two dependencies declaring the
same short name returning `multiple-candidates` with both ids rather than a wrong `unique`**.

**68 new tests, and what they are worth against the pre-fix source.** Checked out `main`'s
extractor and ran both files against it: **58 of the 68 fail**, which is the number that says
they test the fix. The other **10 pass**, necessarily — each asserts an *absence* (no edge for a
primitive type, no trait edge for a closure's `use ($x)`, no second edge for `new Foo()` beside
a `Foo()`, no symbol for an anonymous function, existing call edges still resolving, no
structural edge carrying a namespace as a `sourceModule`). Those ten are regression guards
against this branch, not evidence about it.

Separately, five tests are labelled positive controls and non-regression guards. They fail
against `main` like the other 44, and their job is to catch an over-correction *within* this
branch — the single-namespace, no-namespace and alias-before-reference cases must still
resolve, so no fix here could pass by dropping aliases wholesale. The review round added three
more of the same kind: an alias must still apply to a bare `Foo`, a mixed group's type member
must still resolve, and an anonymous function must still register no symbol.

## Verified end to end against the tool itself

The unit tests assert on resolved `calleeCandidates`, which is the mechanism. The symptom in
the issue is what `codebase_impact` answers, so that was checked directly: built this branch,
pointed a server at `dist/index.js`, indexed the fixture from the issue, and asked the tool.

```
$ codebase_impact src/Base/Base.php  --depth 1
Depth: 1    Total impacted files: 2
Hop 1 (2 files):
  - src/Caller/Caller.php
  - src/Child/Child.php          <- was absent

$ codebase_impact src/Child/Child.php  --depth 1
Depth: 1    Total impacted files: 1
Hop 1 (1 files):
  - src/Consumer/Consumer.php    <- was "nothing else depends on this"
```

Worth noting what the test suite does *not* cover: no test asserts on `getImpactRadius`
itself. The reverse file index it reads is built inline in the persistence path rather than
behind a seam a unit test can reach, so asserting the symptom would mean either duplicating
that index-building in a test or adding a Docker-dependent case. Both looked worse than
leaving the mechanism covered by unit tests and the symptom verified as above. Glad to add it
if you would rather have it, and would take a pointer on the seam you would prefer.

## Real-world validation

The 1,059 tracked PHP files of the WordPress/Bedrock plugin that surfaced the issue, copied to
a temp path and built in-process with `buildCodeGraph` + `resolveCallSites` — the same entry
points the Tier 3 test uses, so it needs no infrastructure and reproduces on demand. Only the
extractor differs between the two runs.

| | `main` | this branch |
|---|---|---|
| PHP symbol edges | 54,745 | 60,024 |
| — of kind `call` | 54,745 | 57,498 |
| — of kind `type_reference` | 0 | 2,526 |
| resolved `unique` | 6,830 | **8,943** |
| resolved `local` | 8,887 | 9,003 |
| `multiple-candidates` | 972 | **972** |
| `unresolved` | 38,056 | 41,106 |
| distinct files reachable as a callee target | 884 | **897** |
| file-graph nodes / edges | 1,150 / 2,854 | 1,150 / 2,854 |

**+2,229 newly resolved edges, and 13 more PHP files that impact analysis can see as a target.**

Three of these rows matter as much as the headline:

- **`multiple-candidates` does not move.** The fix adds no new ambiguity. On this corpus every
  edge it adds also resolves correctly or honestly does not: each of the 2,229 it resolves was
  checked against the class PHP's own name resolution picks, and all 2,229 match. That is a
  measurement of this tree, not a guarantee — the first Known limitation shows the construction
  that breaks it.
- **The file graph is untouched** — identical nodes and edges. This change is confined to the
  symbol graph, which is what the issue described.
- **`unresolved` rises by 3,050**, and that is expected rather than hidden: a WordPress plugin
  type-hints `WP_Error`, `WP_REST_Request`, `WP_Post`, `DateTime` and PHPUnit's `TestCase`
  constantly, and those have no first-party symbol to resolve to. It is the documented
  inline-FQCN limitation, and the same reason those files' call edges are unresolved today.

This corpus is the plugin alone, so it carries none of the theme-to-plugin cross-root edges the
full monorepo has; the figures are conservative against a whole-repo build.

## Checklist

- [ ] I have updated documentation (README.md / DEVELOPER.md) if needed — no public API change; happy to add a DEVELOPER.md note on request

## Related issues

Fixes #181

## Summary by CodeRabbit

* **New Features**
  * Improved PHP code analysis to resolve type references across namespaces, aliases, inheritance, interfaces, traits, object creation, and type declarations.
  * Added support for nullable, union, intersection, and complex type expressions.
  * Preserved and deduplicated existing call relationships while identifying object-creation relationships.

* **Tests**
  * Added coverage for PHP symbol and type-reference resolution, aliases, namespaces, candidate matching, and edge deduplication.


giancarloerra:
Please remove the isolated non-ASCII identifier change in `ed29855` and its dedicated tests from this PR before review. It fixes a distinct pre-existing call-extraction bug and changes behavior beyond #181, so it should be evaluated separately.

Keep this PR limited to the PHP structural edges, alias/type resolution, and regression coverage required for #181. After rebasing, let CI reach a terminal state and leave the new head stable for review.

giancarloerra:
The requested scope split is complete: the non-ASCII commit and its dedicated tests are gone, and #184 now tracks that bug separately. CI is green on the new head.

Five correctness gaps remain before technical review:

1. A leading-backslash single-segment FQCN is treated as a bare alias. With `use X\Other as Foo; class C extends \Foo {}`, `phpTypeRefName` strips the leading slash, sees one segment, and emits an edge to `Other`. PHP names the global `Foo` there. Preserve whether the raw spelling was fully qualified and bypass alias lookup for `\Foo`; cover both `extends \Foo` and `new \Foo()`.

2. Grouped function and constant imports are treated as type aliases. `use function App\Fn\{Base as Thing}; class X extends Thing {}` currently emits a type edge to `Base`, because the `function` or `const` qualifier is on the enclosing use declaration rather than the clause. Inspect the enclosing declaration's import kind and reject grouped function/constant clauses; add cases for both forms.

3. PHP class aliases are case-insensitive, but the new alias table is keyed case-sensitively. `use X\Base as ImportedBase; class C extends importedbase {}` currently emits `importedbase` instead of mapping to `Base`. Normalize alias keys and lookups while preserving the imported spelling on the edge, and cover aliased plus unaliased imports with different casing.

4. `integer`, `double`, and `boolean` are not PHP type keywords. PHP interprets those spellings as class names, with a warning suggesting `int`, `float`, or `bool`. Removing them from `PHP_NON_SYMBOL_TYPES` is required so real class references are not dropped; add regression cases.

5. Return types on closures and arrow functions are omitted. Parameters are found through the root scan, but return types are collected only from `method_declaration` and `function_definition`. Include the corresponding anonymous-function and arrow-function return-type nodes and prove a type used only there produces an edge.

Please batch these corrections, update the focused tests, and leave the resulting head unchanged after CI reaches a terminal state.

giancarloerra:
@coderabbitai review

giancarloerra:
Both review findings are valid and reproduce on the current head:

1. A typed variadic parameter such as `Handler ...$handlers` emits no `type_reference` edge. The parameter scan includes `simple_parameter` and promoted properties but omits the grammar's `variadic_parameter` node. Add that node kind to the same type-field path and add a focused regression test proving the variadic type is retained and attributed to the enclosing function or method.

2. The shared-line braced-namespace test passes malformed PHP. Its JavaScript source literal currently uses `"X\\\\Base"`, producing two consecutive namespace separators in the PHP fixture; PHP rejects it with a parse error, while the extractor still finds the terminal alias segment and lets the assertion pass. Change the source literal to `"X\\Base"`, so the generated PHP contains one separator, and keep the same namespace/offset assertion against valid PHP.

Please batch both corrections, run the focused tests and complete unit suite, then leave the new head unchanged after CI reaches a terminal state. The PR is not ready until that exact head receives a clean incremental review and its actionable threads are resolved.

gregoryfoster:
Both corrections are batched on the new head, `7776dbf`:

- `4aadb27` adds `variadic_parameter` to the parameter type scan. The regression test asserts that a method's variadic type and a function's (by-reference, nullable) variadic type are each attributed to their enclosing method or function, and also covers closure and arrow-function variadics. It fails without the fix.
- `1a5be0f` changes the literal to `"X\\Base"`. The generated PHP now passes `php -l`, and the namespace/offset assertion is unchanged.
- `7776dbf` is style only: three comment dashes that the em-dash conversion in `dceb7e1` missed.

Focused tests (`graph-symbols-php-references`, `code-graph-php-symbol-edges`): 69 passed. Full unit suite: 2,350 passed on Node 22 and 26. `npm run lint` and `tsc --noEmit` are clean.

I'll leave the head unchanged from here.

giancarloerra:
@coderabbitai review

PR #186, opened 2026-09-22
fix(graph): read PHP names by the language's identifier rule, non-ASCII included

### LINKED ISSUE #184: [Bug]: non-ASCII PHP identifiers are truncated into a different name — `$o->crée()` records a confidently resolved call to `e()`
### Describe the bug

`extractCalleeNamePhp` (`src/services/graph-symbols.ts`) finds the callee name by anchoring a
scan at the end of the receiver with an ASCII-only pattern:

```ts
const m = receiver.match(/([A-Za-z_]\w*)$/);
```

PHP states its identifier rule over **bytes** — `[a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*` —
and every byte of a UTF-8 multibyte sequence is at least `0x80`, so any non-ASCII character is
legal in a PHP identifier. The guard therefore disagrees with the language in two different
ways, and the first is worse than a miss:

1. **A name with an ASCII tail is silently truncated into a different name.** `$f->crée()`
   matches only the trailing `e` and records a call to `e`. That name is not missing, it is
   *fabricated* — and it resolves like any other name, so it can attach to a real, unrelated
   symbol with `confidence: "unique"`.
2. **A name with no ASCII tail produces no edge at all.** `$f->данные()` and `$f->文書()`
   return `null`, so the call is dropped without a word.

Confirmed against the runtime rather than read off the grammar — PHP 8.4.23:

```
$ php -r 'function crée(){ return "yes"; } function данные(){ return "yes"; }
          echo "crée() -> ", crée(), "; данные() -> ", данные(), "\n";'
crée() -> yes; данные() -> yes
```

### Steps to reproduce

A four-file Composer project. The point is that the fabricated name `e` **exists** as a real
function somewhere else in the project, which is what turns a truncation into a wrong answer.

`composer.json`

```json
{ "autoload": { "psr-4": { "App\\": "src/" } } }
```

`src/Helpers/escape.php` — `render()` never calls `e()`; the file is required for other reasons

```php
<?php

namespace App\Helpers;

function e(string $s): string
{
    return htmlspecialchars($s);
}
```

`src/Model/Formatter.php`

```php
<?php

namespace App\Model;

class Formatter
{
    public function crée(): string
    {
        return 'ok';
    }
}
```

`src/Model/Document.php`

```php
<?php

namespace App\Model;

require_once __DIR__ . '/../Helpers/escape.php';

class Document
{
    public function render(Formatter $f): string
    {
        return $f->crée();
    }
}
```

1. `codebase_index` the project, then `codebase_graph_build`.
2. `codebase_symbol` for `e`, then `codebase_symbol` for `crée`.

(`codebase_symbol` is the lens that isolates this. `codebase_impact` also reports
`escape.php → Document.php`, but the `require_once` puts a file-graph edge there too, so that
answer cannot tell you which of the two produced it. A symbol-level caller list can only come
from a symbol edge.)

### Expected behavior

`$f->crée()` names the method `crée`. The edge should either carry that name, or — if the
extractor cannot represent it — not exist at all. A call to `crée()` must never be recorded as
a call to `e()`: an omission costs a true edge, but a truncation asserts a false one, and
`e()` should not be credited with a caller that never names it.

### Actual behavior

`e()` is credited with a caller it does not have, and `crée()` is credited with none:

```
$ codebase_symbol e
Symbol: e (function)
Defined: src/Helpers/escape.php:6-9  [php]

Callers (1):
  ← src/Model/Document.php:12 (call)          <- line 12 IS `return $f->crée();`

$ codebase_symbol crée
Symbol: crée (method)
Defined: src/Model/Formatter.php:7-10  [php]

Callers (0):
  (none — possibly an entry point or unused)
```

Underneath, the single edge out of `Document.php` through `buildCodeGraph` + `resolveCallSites`
is:

```
calleeName="e"  kind=call  confidence=unique  candidates=["src/Helpers/escape.php::e#6"]
```

`confidence: "unique"` — the graph is not unsure. It states positively that
`Document::render()` calls `App\Helpers\e()`, which it does not, and everything downstream of
the symbol graph inherits that edge.

The full set of behaviours, against `main` at v1.14.0:

| receiver | recorded callee | |
|---|---|---|
| `$o->render()` | `render` | correct |
| `$o->crée()` | **`e`** | fabricated |
| `$o->données()` | **`es`** | fabricated, and truncated differently |
| `$o->данные()` | *no edge* | dropped |
| `$o->文書()` | *no edge* | dropped |
| `Café::make()` | `make` | correct — the method name is ASCII |

### Embedding provider

Ollama (Docker — default)

### Qdrant mode

Managed (Docker — default)

### MCP host

### Operating system

macOS 26.6.2

### Node.js version

v22.18.0

### Relevant log output

```shell
(nothing is logged — both failure modes are silent)
```

### Additional context

Scope, and why this is its own issue: this came out of the work on #181, where it rode along as
one isolated commit. You asked for it to be taken out of #183 and evaluated separately, which
is what this is. #183 no longer contains it.

The fix is a one-line change in each of two guards, replacing the ASCII-only patterns with one
documented `PHP_IDENTIFIER` and its end-anchored twin. Worth stating up front, since it is a
behaviour change and not only a repair:

- Measured over a real 1,059-file PHP tree, **no existing call edge is lost or renamed** — that
  corpus is ASCII-only, which is the common case.
- A codebase that *does* use non-ASCII identifiers will see real movement: edges appearing where
  there were none, and names corrected. That is the fix working, but it is visible.

One detail worth deciding deliberately: the character range should run to the top of the code
unit space rather than transliterating the spec's `\x80-\xff` literally. That byte range
describes UTF-8 bytes; read as decoded characters it would still reject `class Документ` and
`class 文書`, which the packaged grammar parses as `name` exactly as it parses `Café`.

If #183 lands first, note that the `phpTypeRefName` guard it adds makes the same ASCII
assumption (documented at the site there), so one change would cover both.

I have this implemented and tested against the cases above. Happy to open a PR if you want it —
or to leave it here if you would rather take it yourself.

### Checklist

gregoryfoster:
## Summary

PHP states its identifier rule over bytes (`[a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*`), and every byte of a UTF-8 multibyte sequence is at least `0x80`, so any non-ASCII character is legal in a name. Both PHP name guards were ASCII-only:

- `extractCalleeNamePhp` anchors its match at the end of the receiver, so `$o->crée()` matched the ASCII tail alone and recorded a call to `e`, which resolves `unique` to any real `e()` in scope. `$o->данные()` and `$o->文書()` recorded nothing.
- `phpTypeRefName` rejected the whole name, so `extends Café`, `new 文書()` and a `Документ $d` parameter produced no edge.

As scoped in #184, both now read one PHP identifier rule.

## Changes

- **One rule, two anchorings.** `PHP_IDENTIFIER_START` / `PHP_IDENTIFIER_PART` build `PHP_IDENTIFIER_SOURCE`, and from it:
  - `PHP_IDENTIFIER`, the whole-name test used by `phpTypeRefName`;
  - `PHP_IDENTIFIER_TAIL`, the end-of-receiver match used by `extractCalleeNamePhp`.

  The two guards in this file cannot drift apart. (`graph-resolution.ts` states the same rule again for the file-import graph, which reads PHP namespace and class declarations out of source text rather than a parse.)
- **The range is `\^@-\uffff`, with no `u` flag.** Every UTF-16 code unit of a non-ASCII character is at least `0x80`, so a character outside the BMP is admitted by its two surrogate halves: `$o->𠮷Id()` gives `𠮷Id`. With `u`, it would be cut to `Id`; a test pins this. The range deliberately does not transliterate the spec's `\x80-\xff`, which read as decoded characters would still reject `Документ` and `文書`.
- **The tail match is pinned to the start of a run by a lookbehind.** Anchored only at the end, it retries from every character of a run of name characters and rescans to the run's end each time, which is quadratic in the run's length. Once non-ASCII is admitted, a CJK paragraph in an earlier argument of a fluent chain is a single run:

  | pattern | one 52,000-character statement |
  |---|---|
  | end-anchored | 1.26 s |
  | pinned | 0.2 ms |

  The old ASCII pattern had the same cost on a long ASCII word (0.96 s). Both forms return the same name.
- The "Deliberately ASCII-only" comment #183 left at the `phpTypeRefName` guard is replaced.

## Not in this PR: a third ASCII-only site

The file-import graph makes the same assumption, outside the two guards #184 scopes. `phpUseSpecifiers` (`graph-imports.ts`) reads a `use` clause with `[\w\\]+`, so a `use` of a non-ASCII name reaches no file. Where the name has an ASCII prefix, it is truncated into a different one:

```php
// composer.json: { "autoload": { "psr-4": { "App\\": "src/" } } }
// src/Caf.php declares App\Caf
use App\Café\Menu;   // read as App\Caf: a file edge to src/Caf.php
```

The new real-build test reaches its non-ASCII class files through `require_once` for that reason, and says so. Happy to take this as a follow-up.

## Known limitations

- **The grammar cuts a name short at a character outside the BMP,** on both sides of a call. Its `name` token stops at U+FFFF, and the rest parses as an `ERROR` fragment. Inside the BMP, where every accented, Cyrillic and CJK name lives, both sides read the name whole.
  - **A type reference is not emitted at all** when the parse returns part of one: `extends 野𠮷`, `extends 𠮷Doc` and `extends App\野𠮷` produce nothing, rather than `野`, `Doc` or `App`. `main` emits `App` for the qualified form today.
  - **The same below the BMP**, at the characters the grammar's `name` token treats as whitespace although PHP admits them inside an identifier - U+00A0, U+200B and U+FEFF. `extends Caf<U+00A0>e` is valid PHP naming one 6-byte class; it parses as `Caf`, which `main` emits, and now produces nothing.
  - **Declarations** are still filed under the remainder: `function 𠮷Id()` becomes `Id`, `class 𠮷Doc {}` becomes `Doc`.
  - **A callee** is read from the call node's own text, which the cut does not reach, so `$o->𠮷Id()` is named `𠮷Id` and answers no declaration. It is unresolved where `main` resolved it - but `main` matched only by truncating the call to `Id`, a name any unrelated `Id()` in the closure could answer just as well.
  - Two neighbouring shapes are left as they are, both already on `main`: an alias imported from a split `use` (`use App\Base𠮷 as Al;` registers `Al` as `App`), and a split leaving an ASCII tail that parses as its own call (`野𠮷Id();` records `Id`).

  Recorded in the `PHP_IDENTIFIER` doc comment and in `extractFromPhp`'s guard, and pinned by tests so a grammar change cannot flip any of it silently.

## Type of change

- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] Unit tests pass (`npm run test:unit`): 100 files, **2,378 tests** (25 new), on Node 22 and 26
- [ ] Integration tests pass: not run (Docker)
- [x] TypeScript compiles cleanly (`npx tsc --noEmit`); `npm run lint` is clean

**Extractor tests.**

- `graph-symbols-php-calls.test.ts` runs a table of receivers and asserts the exact callee for each: accented, Cyrillic and CJK methods, the ASCII-tailed forms, a name outside the BMP, a bare function, and static calls on non-ASCII classes.
- `graph-symbols-php-references.test.ts` covers:
  - non-ASCII class names in every structural position, including by terminal segment of a qualified name;
  - a non-ASCII alias through the ASCII-only case fold: `ÄRGER` names the import, and `ärger` does not.

**Real build.** `code-graph-php-non-ascii-identifiers.test.ts` runs the issue's Composer layout through real `buildCodeGraph` + `resolveCallSites`.

- Three non-ASCII methods end in an ASCII tail. A required `decoys.php` declares a function under each tail (`e`, `Json`, `Id`), inside the caller's dependency closure.
- **Against `main`, all three calls resolve `unique` to those decoys.**
- Now each call resolves to the method it names, and nothing resolves to a decoy.
- Cyrillic and CJK class names resolve as type references and through `new`.

**Against `main`'s extractor, 22 of the 25 new tests fail.** The three that pass are expected to:

- an ASCII method;
- an ASCII static method on an accented class (`Café::make()` gives `make` on `main` too);
- the guard that the decoys sit inside the dependency closure, so the decoy assertion cannot pass vacuously.

Every PHP fixture in the new tests passes `php -l` (PHP 8.4).

**ASCII behaviour is unchanged.** Over a 1,296-file ASCII-only PHP tree:

- all 62,821 extracted edges (60,218 `call`, 2,603 `type_reference`) are identical before and after;
- extraction takes 2,726-2,743 ms, against 2,859-2,947 ms on `main` (best of three).

## Checklist

- [ ] I have updated documentation (README.md / DEVELOPER.md) if needed: no public API change

## Related issues

Fixes #184

## Summary by CodeRabbit

* **Bug Fixes**
  * PHP code graphs now recognize method, function, and type names containing accented, Cyrillic, CJK, and other non-ASCII characters, preserving full names in call and type references.
  * Names that the parser splits are no longer recorded as misleading partial references, preventing links to unrelated symbols.


giancarloerra:
The known-limitations section identifies a new false-edge case: `extends 野𠮷` reaches `phpTypeRefName` as `野`, which this PR now accepts, whereas `main` emitted no reference. If a reachable class `野` exists, that partial name can resolve as the wrong parent. Please suppress a type-reference edge when the parser has split a PHP identifier at an adjacent non-BMP character or parse-error fragment, while retaining the accented, Cyrillic, and CJK cases this PR fixes. Add a regression with a reachable `野` decoy that proves the partial name cannot resolve to it. Full support for identifiers beyond the grammar's BMP limit can remain separate; avoiding a new confidently wrong edge is needed here.

gregoryfoster:
Fixed in `b8cf848`.

`pushTypeRef` now reads the source around a reference before emitting it: a character that would carry the name on, touching either end of the node, means the parse returned part of one. `extends 野𠮷`, `extends 𠮷Doc` and `f(𠮷Doc $d)` now emit nothing.

The qualified form is worth flagging, because the parse gives up at the cut and hands back only what precedes it: `extends App\野𠮷` arrives as the name `App`, which `main` emits today. The same guard closes that.

The two ends ask different questions, which is what keeps valid PHP intact. On the right an identifier character or a `\` carries the name on. On the left only two identifier characters meeting do, so `new\App\Made()` - valid PHP, since `\` ends the keyword there - keeps its edge instead of losing it to the `w` of `new`. (`extends\App\Foo` is not the same case: PHP 8 lexes that as one namespaced name and rejects it, so it is no test fixture.) Accented, Cyrillic and CJK references are untouched: they parse whole.

One case below the BMP comes with it. The grammar's `name` token treats U+00A0, U+200B and U+FEFF as whitespace, though PHP admits them inside an identifier, so `class A extends Caf<U+00A0>e {}` - valid PHP, and one 6-byte class name, with the character written out here because it renders invisibly - parses as `Caf`, and `main` emits that. It now emits nothing. Same trade as above: the parse had already cut the name, so the choice is between no edge and an edge to whatever the surviving part names.

Regression coverage, all of it failing without the guard:

- a real build with a reachable `野` decoy: `Split.php` extends `野𠮷`, the decoy is required into its dependency closure, and nothing resolves to it;
- extractor cases for the bare, leading, qualified and ASCII-surviving splits;
- positive controls that whole names, including qualified and `\`-rooted ones, still resolve.

Unit suite 2,387 on Node 22 and 26; lint and `tsc --noEmit` clean. Over a 1,296-file ASCII-only PHP tree, all 62,821 extracted edges are identical to `main`'s.

Two neighbouring shapes are left alone, both already on `main` and neither introduced here: an alias imported from a split `use` (`use App\Base𠮷 as Al;` registers `Al` as `App`, so a later `extends Al` carries `App`), and a split that leaves an ASCII tail parsing as its own call (`野𠮷Id();` records `Id`). Happy to take either as a follow-up.

giancarloerra:
@coderabbitai review

PR #187, opened 2026-09-23
fix(graph): resolve PHP static calls and class references by the class they name

### LINKED ISSUE #185: [Bug]: PHP Class::method() calls are resolved by the method name alone — a call to another class's same-named method is recorded as the caller calling itself
### Describe the bug

A PHP `Class::method()` call is resolved by the method name alone. `extractCalleeNamePhp` names `Cls::make(…)` as `make` and `Acme\Support\Cls::of(…)` as `of` — as its own doc comment sets out — and leaves `calleeQualifier` unset, so resolution never learns which class the call names. It takes any method of that name: first in the caller's own file, then across the caller's dependency closure.

That produces two kinds of wrong edge, and the first is the common one:

1. **Delegation is recorded as recursion.** The same-file branch runs first, so a method that calls a same-named method on *another* class is recorded as calling itself — and the method it really calls is left with no callers at all.
2. **A call lands on a different class.** Where the class written is not indexed (a framework class, typically), the closure scan settles on whichever indexed class declares a method of that name.

Neither is an ambiguity the graph admits to: both fixtures below come back `local` or `unique`, as do 108 of the 113 wrong edges measured further down, and `codebase_graph_status` counts them as resolved.

### Steps to reproduce

**1. Delegation recorded as recursion** — three classes in one namespace:

`composer.json`

```json
{ "autoload": { "psr-4": { "App\\": "src/" } } }
```

`src/Schema/Schema.php`

```php
<?php

namespace App\Schema;

class Schema
{
    public static function all(): array
    {
        return array_merge(
            UserSchema::all(),
            OrderSchema::all(),
        );
    }
}
```

`src/Schema/UserSchema.php` and `src/Schema/OrderSchema.php`, each:

```php
<?php

namespace App\Schema;

class UserSchema    // OrderSchema in the other file
{
    public static function all(): array
    {
        return [];
    }
}
```

`codebase_index` the project, then `codebase_symbol` for `all`, with `file` set to `src/Schema/Schema.php` and then to `src/Schema/UserSchema.php`.

**2. A call lands on a different class** — a framework class that is not indexed, and an unrelated indexed one that happens to declare the same method:

`src/Models/Invoice.php`

```php
<?php

namespace App\Models;

class Invoice
{
    public static function capture(): self
    {
        return new self();
    }
}
```

`src/Http/Controller.php`

```php
<?php

namespace App\Http;

use Illuminate\Http\Request;
use App\Models\Invoice;

class Controller
{
    public function handle(): void
    {
        Request::capture();
    }

    public function bill(Invoice $invoice): void
    {
    }
}
```

`codebase_index` the project, then `codebase_symbol` for `capture`.

### Expected behavior

`Schema::all()` calls `UserSchema::all()` and `OrderSchema::all()`, and they each have one caller. `Invoice::capture()` has no callers: the one `capture()` call in the project is to Illuminate's, which is not indexed, so that call is `unresolved`.

Checked against the runtime, PHP 8.4.23, with a stub `Illuminate\Http\Request`:

```
Schema::all() invokes:
  UserSchema::all
  OrderSchema::all

Controller::handle() invokes:
  Illuminate\Http\Request::capture
```

### Actual behavior

v1.14.0 as published to npm, run by the marketplace plugin (`npx socraticode@latest`, the registry's current `latest`):

```
$ codebase_symbol all  (file: src/Schema/Schema.php)
Symbol: all (method)
Defined: src/Schema/Schema.php:7–13  [php]

Callers (2):
  ← src/Schema/Schema.php:10 (call)       <- line 10 is UserSchema::all()
  ← src/Schema/Schema.php:11 (call)       <- line 11 is OrderSchema::all()

Callees (3):
  → array_merge [unresolved, call]
  → all [local, 1 candidate(s), call]
  → all [local, 1 candidate(s), call]

$ codebase_symbol all  (file: src/Schema/UserSchema.php)
Symbol: all (method)
Defined: src/Schema/UserSchema.php:7–10  [php]

Callers (0):
  (none — possibly an entry point or unused)
```

```
$ codebase_symbol capture
Symbol: capture (method)
Defined: src/Models/Invoice.php:7–10  [php]

Callers (1):
  ← src/Http/Controller.php:12 (call)     <- line 12 is Request::capture()
```

`Schema::all()` is reported as recursive, `UserSchema::all()` as possibly unused, and `Invoice::capture()` as called by a line that names `Request`.

**How often, on real code.** Measured over a 1,059-file WordPress plugin with `buildCodeGraph` + `resolveCallSites` on `main`, checking each `Class::method()` call's resolved method against the class PHP's own name resolution picks (through the file's `use` imports and current namespace; `self`, `static`, `parent` and `$var::` left out):

| | |
|---|---|
| `Class::method()` call edges | 6,439 |
| resolved | 4,629 |
| **resolved to a method of a different class** | **113** |
| — `local`, every one of them onto the caller's own class | 99 |
| — `unique` | 9 |
| — `multiple-candidates`, the right class not among them | 5 |
| of the 113, written fully qualified | 13 |

About 2.4% of resolved scoped calls, and 88% of those are the recursion case. The shape behind nearly all of them is an ordinary one — a registry or aggregator that calls the same-named method on each class it collects — and it is exactly where `codebase_impact` and `codebase_symbol` are most likely to be asked "is this still used?".

Writing the class out in full does not help. A call spelled `\Vendor\Repository\ThingRepository::find_by_id(…)` (names changed — the tree is private), whose target exists in the tree, resolved to the same-named method on seven *other* repositories and not to `ThingRepository`'s own — the fully-qualified name was discarded along with everything else to the left of `::`.

### Embedding provider

Ollama (Docker — default)

### Qdrant mode

Managed (Docker — default)

### MCP host

### Operating system

macOS 26.6.2

### Node.js version

v22.18.0

### Relevant log output

```shell
(nothing is logged — the edges resolve normally; codebase_graph_status reports "Unresolved: 0.0%" for fixture 2)
```

### Additional context

**This is the PHP counterpart of #143.** That issue raised the same question for Rust — `resolveCallSites` matches a callee by name only, and the qualifier is the information that would narrow it — and #147 answered it: *"A qualified call is resolved by its qualifier or not at all."* PHP never got the resolution half. The extractor already sees the qualifier; it just does not record it.

The branch in `graph-symbol-resolution.ts` that honours `calleeQualifier` is gated on `.rs`, deliberately, and its comment says why: everything below it reads `::`, `crate`, `self` and `super` with Rust's meaning, and the field is carried for every language. So a PHP fix needs its own resolution semantics rather than reusing Rust's. The shape I would suggest:

- **Extraction.** For `Cls::m()`, set `calleeQualifier` to the class, resolved to a fully-qualified name through the file's `use` imports and current namespace.
- **Resolution, for `.php`.** Keep only candidates declared on that class, and apply the rule to the **same-file branch too** — that is where 99 of the 113 come from. It needs each candidate's enclosing class and namespace, which PHP symbols do not carry today (`qualifiedName` is the bare name).
- **Out of scope.** `self::`, `static::` and `parent::` name the enclosing class or its parent and could be resolved the same way, but they are worth their own decision. `$obj->m()` is a type-inference question, not a qualifier one.

**How this relates to #183.** #183 adds type-reference edges for #181 and resolves them the same way — by terminal name within the caller's closure — so there a type edge can point at a same-named class in another namespace. On the same corpus that happened 0 times in 2,229 resolved edges, and #183 documents it as a known limitation deferred to this issue, because the fix is this one: carry the qualifying name — the class for a call, the namespace for a type — and filter by it. It is also why the fix is not in #183: it is a resolution change, and it repairs `main`'s own call edges, which #183 does not touch.

I have the measurement harness from above and would be glad to implement whichever shape you prefer, with the same edge-by-edge before/after.

### Checklist

gregoryfoster:
## Summary

A statically named PHP call, `Cls::m()`, was emitted as its method name alone, and a class reference as its short name, so resolution could only match by name: the caller's own file first, then its dependencies. That is how `UserSchema::all()` inside `Schema::all()` became a `local` self-edge, and how `Request::capture()` reached a same-named `Invoice::capture()` with `unique` confidence. The same short-name matching was #183's documented wrong-class limitation for structural references.

To the scope set out in #185, this PR:

- resolves the class a call or reference names under PHP's namespace and `use` rules, and carries it on the edge;
- gives every PHP class and method an owner;
- requires the two to match, and leaves the edge `unresolved` when they don't.

## Changes

- **Qualifying static calls.** A static call whose class is written as a name carries that class as `calleeQualifier`, resolved the way PHP resolves it. The name can be bare, imported (plain, aliased or grouped), qualified, namespace-relative or fully qualified (`phpClassRef` in `graph-symbols.ts`). The rules:
  - A fully qualified name is taken as written.
  - `namespace\X` is read under the current namespace.
  - A qualified name goes through an import of its first segment, and otherwise sits under the current namespace.
  - A bare name goes through its import, and otherwise sits under the current namespace. A class name never falls back to the global namespace.
- **Qualifying class references.** #183's structural references (`extends`, `implements`, a trait `use`, type hints, `new`) are qualified by the same rule, and carry the class's namespace. A namespace-relative `extends` or `new` (`namespace\Base`) is now read; it was skipped before.
- **Owners.**
  - Each PHP class, interface and trait carries the new optional `SymbolNode.phpOwner`: its namespace, as a prefix with PHP's trailing `\` (`\App\Models\`, and `\` for the global namespace).
  - Each method carries its class, interface, trait or enum (`\App\Models\Invoice`), read from the declaration's own `name` field.
  - Functions, and methods of anonymous classes, have no owner.
  - The two forms cannot collide, so a class qualifier reaches only methods and a namespace qualifier only classes.
- **Names.** The `use` table keeps each import's full path, with a grouped `use`'s prefix folded in, and each namespace scope keeps its name.
- **Cut names** (#184's guard). A static call's class now passes the same `isSplitName` check class references do, and is left unqualified when the parse cut it short. Because the qualifier now reads every segment, a fragment cut from in front of a `\`-rooted name counts as a cut too: `new 𠮷\App\Made()` would otherwise name `\App\Made`. It is recognised by a non-ASCII character in the run before the `\`, so `new\App\Made()` keeps its edge.
- **Dedupe.** The key includes the qualifier, so `new A\Foo()` and `new B\Foo()` in one method stay two edges.
- **Resolver** (`graph-symbol-resolution.ts`). A PHP branch sits beside the Rust one, gated on the caller's language.
  - A qualified edge is answered from a per-file index keyed by owner and name. Class and namespace names fold ASCII-case-insensitively, as PHP matches them; method names are matched exactly, as every other call is.
  - The caller's own file is searched first (`local`), then its dependencies (`unique` / `multiple-candidates`).
  - When neither holds a match, the edge is `unresolved` with no candidates. There is never a fallback to the method name.
- **Unchanged.**
  - `$obj->m()`, `$class::m()`, `self::`, `static::`, `parent::`, and bare or qualified function calls carry no qualifier and resolve as before.
  - Symbol `id`s and `qualifiedName`s are untouched.
- **Shared fold.** `phpFoldCase` moves to a small module of its own (`graph-php-case.ts`), so the resolver can share it without loading the extractor and its native parser binding.
- **Performance** (separate commit). Namespace declarations are read from the root's top-level statements, where PHP requires them, instead of by searching the tree.

Two of #183's tests change with this:

- Two classes sharing a short name in one signature are now two edges, each `unique`, where they were one edge with two candidates (`multiple-candidates`).
- `new Foo()` beside a bare `Foo()` call is its own qualified edge, instead of being merged into the function call.

A separate commit fixes one of #183's own fixtures: the "emits nothing for `self`, `parent` and `static`" test used `parent` in a class with no parent, which PHP rejects (`php -l`). `C` now extends `B`, and the test asserts that `extends` is the only reference.

## Upgrading: no reindex, migration or reset

- `phpOwner`, and PHP's use of `calleeQualifier`, are optional fields. A graph persisted before them stays readable, and its edges answer as they did.
- New edges never resolve against old symbols:
  - Any changed or removed file forces a full symbol rebuild (`symbol-graph-incremental.ts`).
  - A graph built by another version is rebuilt once, on the first update after an upgrade (`builtByVersion` in `graph-inputs.ts`).

## Measured on a real tree

These figures cover the 6,477 static calls in a 1,059-file WordPress plugin, comparing `main` with this branch on the same tree:

| | `main` | this branch |
|---|---|---|
| to the class written | 4,538 | **4,588** |
| to an unrelated class | **112** | **0** |
| to a parent class | 1 | 0 |
| unresolved | 1,826 | 1,889 |

- **The 113 wrong-class edges the issue counted are gone.**
  - 50 now resolve to the class they name. A wrong same-named method, often the caller's own, had been shadowing it.
  - 63 are left `unresolved`: the class named isn't among the caller's dependencies, or, once, the method is inherited.
- **No resolution to the class written was lost.**
- **Structural references on the same tree are unchanged:** 5,322 edges, 2,257 `unique`, and 0 to the wrong class, both before and after. No caller there reaches two classes that share a short name.
- **Extraction time** is within run-to-run noise of `main`: 2,675–2,712 ms against 2,646–2,707 ms, best of three. It was about 4% slower before the performance commit.
- The harnesses read each call's intended class independently of this code, with a separate namespace and `use` reader, so they check this change rather than restate it.

## Known limitations

- **Inherited static methods are not followed.** `Child::create()` resolves only to a `create` declared on `Child` itself. One inherited through `extends` or a trait is left `unresolved`, never answered wrongly.
  - On the tree above, 93 of the 6,477 static calls name an inherited method, and `main` resolved 1 of them.
  - Following the already-resolved `extends` and trait `use` edges would be a natural follow-up. Happy to take it.
- **A same-namespace class referenced without a `use` is out of reach.** The file graph draws no dependency for it, so the issue's self-edge reproduction now ends `unresolved` rather than resolved. The self-edge is gone, and importing the sibling schemas resolves each call (tested). The file graph is unchanged here.
- **Method names are matched exactly**, as every other call is, so `Invoice::Capture()` is left `unresolved` even though PHP would call `capture()`.

## Type of change

- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] Unit tests pass (`npm run test:unit`): 103 files, **2,450 tests** (63 new), on Node 22 and 26
- [ ] Integration tests pass: not run (Docker)
- [x] TypeScript compiles cleanly (`npx tsc --noEmit`); `npm run lint` is clean

**Extractor** (`graph-symbols-php-qualifiers.test.ts`, 32 tests):

- the qualifier each spelling produces: bare, imported, aliased, grouped, qualified through an import or through the namespace, namespace-relative, fully qualified, global, and case-folded;
- braced namespace blocks, and a `use` declared below the call;
- the forms left unqualified;
- owners for classes, interfaces, traits, enum methods, global and braced namespaces, anonymous classes and attributed classes;
- ids and qualified names left as they were;
- a class name the parse cut short, on a static call and on `new`, beside `new\App\Made()`, which keeps its edge.

**Resolver** (`graph-symbol-resolution-php-qualified.test.ts`, 14 tests):

- cross-file and same-file candidates, and two same-named methods;
- no self-edge from a method the caller also declares;
- unresolved for a class the project doesn't have, or a method the class doesn't declare;
- case folding for class names but not method names;
- class references, and proof that the two owner forms cannot meet;
- symbols without an owner, and a search space no wider than before;
- unqualified calls unchanged.

**Real build** (`code-graph-php-scoped-qualifiers.test.ts`, 17 tests):

- both reproductions from the issue, as filed;
- the self-edge case again with the schemas imported, where each resolves to its own class;
- a controller that declares a `capture()` of its own and calls four others: through an import, an alias and a fully qualified name, and on a class the project doesn't have;
- a namespace-relative class and parent in the same file;
- `new` on a fully qualified class;
- type hints naming Illuminate's `Request` and the project's own.

**Against `main`'s source, 46 of the 63 new tests fail.** The other 17 pass because they must: they guard behaviour this PR leaves alone (the unqualified forms, ids, the search space, a cut static call), plus one check that the wrong answer is reachable to be refused.

Every PHP fixture these tests add passes `php -l` (PHP 8.4).

## Checklist

- [ ] I have updated documentation (README.md / DEVELOPER.md) if needed: no public API change

## Related issues

Fixes #185


giancarloerra:
Two correctness gaps remain before technical review:

- The original #185 Schema.php fixture expects UserSchema::all() and OrderSchema::all() to resolve to their respective methods. The new real-build test instead requires both to stay unresolved unless redundant use imports are added. These sibling classes are present under the configured PSR-4 namespace. Use the exact qualified class to find its indexed declaration even when the file graph has no use edge, without falling back to method-name-only matching, and assert the expected edges in the original fixture.
- The PR reports one previously correct inherited-method edge becoming unresolved. PHP resolves Child::create() to a method declared on its parent. After checking the named class, follow only verified inheritance to preserve that existing edge; leave unrelated methods unresolved. Add a real-build regression for this previously correct case.

Please retain the unresolved external Request::capture() regression while addressing both points.

gregoryfoster:
Thanks, both addressed:

- **Same-namespace siblings** (`258fe80`): a qualified edge now looks for its exact qualified class across the project, after the caller's own file and its dependencies. The key is the whole class name, so it never falls back to the method name. The original `Schema.php` fixture now asserts `UserSchema::all()` and `OrderSchema::all()` resolve to their own methods, and the imported variant is gone.
- **Inherited static methods** (`7749032`): classes and traits record their `extends` and trait `use` (`phpExtends`, `phpTraits`). When the named class doesn't declare the method, resolution checks its traits, then its parent, on up the chain. It follows a class only when exactly one declaration of it is found under its exact name, so a parent outside the project ends the search and unrelated methods stay unresolved. An abstract trait method doesn't shadow the implementation the class inherits. The real-build regression is the previously correct case: the caller imports both `User` and `Model`, and `User::create()` resolves to `Model::create()`.
- `Request::capture()` still stays unresolved.
- Also: `Cls::$m()` is no longer qualified (`0b97f46`), and a sort that the top-level namespace read made dead is gone (`3eb0707`).

On the same plugin tree, the 6,477 static calls went from 4,538 resolving to the class written, 1 to the inherited method PHP runs, 112 to an unrelated class and 1,826 unresolved on `main`, to 6,339 / 93 / 0 / 45 on this branch. No previously correct edge was lost. The PR description has the details.

giancarloerra:
@coderabbitai review

